Issue #265, opened 2026-05-09
Mobile preview build: map tile resolution lower than dev build


cner-smith:
Preview EAS build shows lower resolution map tiles compared to dev build. Likely cause: different Mapbox token or token configuration between dev (.env) and preview (EAS secrets).\n\nCheck EXPO_PUBLIC_MAPBOX_TOKEN in EAS env vars matches the token used in dev.

cner-smith:
## Audit notes (2026-05-13) — dev/preview parity deep dive

Most likely actual cause: **\`RNMapboxMapsDownloadToken\` is missing from the \`@rnmapbox/maps\` plugin config.**

Current \`apps/mobile/app.json:42-49\`:
\`\`\`json
["@rnmapbox/maps", { "RNMapboxMapsImpl": "mapbox" }]
\`\`\`

\`@rnmapbox/maps\` requires a **secret SDK download token** (\`sk.xxx\` with \`DOWNLOADS:READ\` scope) to fetch the Mapbox Native SDK from Mapbox's private Maven during the native Android build. The plugin reads it from the \`RNMapboxMapsDownloadToken\` field in the config.

**Without this token**, the build falls back to a publicly-available Mapbox SDK distribution that is older / OSS-variant / lower-quality assets. Visible tile quality differences between builds — exactly what this ticket reports.

The dev client APK was probably built when the token was previously configured (or with a cached SDK), so it has the modern SDK. New preview builds on EAS fetch fresh and get the public fallback.

## Other findings from the dev/preview/prod audit (filed separately)

- **EAS production env vars are empty.** Real launch blocker for production builds. → see separate issue.
- **Supabase URL skew between local \`.env.preview\` and EAS preview.** Mismatch in which Supabase project preview points at vs what \`.env.preview\` would imply. → see separate issue.
- **Mapbox access tokens (\`pk.\`) all look type-consistent.** All three local files have \`pk.\`-prefixed tokens, 92 chars. EAS preview also has \`EXPO_PUBLIC_MAPBOX_TOKEN\` set. Can't tell from prefix/length whether identical; user should verify by comparing token IDs in Mapbox dashboard.
- **\`@rnmapbox/maps\` version is \`10.1.33\`** in both — no SDK drift from package.json side.
- **\`styleURL\` uses the same \`Mapbox.StyleURL.Satellite\`** in both. Not a style-URL difference.
[...]

Issue #264, opened 2026-05-09
Mobile: map crashes randomly during live round — possible memory leak


cner-smith:
App crashes randomly when using the map during live rounds. Suspected memory leak from map instances not being properly cleaned up between hole transitions or navigation.\n\nNeeds investigation:\n- Multiple MapView instances created without destruction\n- WebGL context management\n- Memory pressure from high-res tile caching\n\nReproduction: use live round map for several holes, crashes occur randomly.

cner-smith:
## Audit notes (2026-05-13) — performance-detective agent

The 18-instances-per-round theory is real. Architectural root cause:

**Mechanism.** Each hole is its own Expo Router screen. `navigateHole` and `finishHole` in \`apps/mobile/app/(app)/round/[id]/hole/_hooks/useShotActions.ts:402,407\` call \`router.replace(/round/{id}/hole/{n+1})\`. Expo Router renders the new screen as a fresh component tree — \`<HoleScreen>\` → \`<HoleMap>\` → \`<Mapbox.MapView>\` mounts new every transition. Each MapView brings:

- A native EGL context (~30–60 MB GPU memory on Android)
- Its own satellite tile cache
- 5 \`Mapbox.ShapeSource\`s + \`LineLayer\`s + N \`PointAnnotation\` views
- A \`LocationPuck\` with internal native subscription

\`@rnmapbox/maps\` v10.1.x on Android has a known issue where the \`MapView\` native peer is **not torn down promptly** when its React host unmounts — the GL context lingers until GC, and GC of native textures only runs under memory pressure. By the time pressure hits, you're already past the OOM kill threshold.

Over 18 holes you allocate 18 native GL surfaces in succession; native peers stay in the Mapbox SDK's internal MapView registry until Android's LowMemoryKiller fires.

**Compounding factor (filed separately): BreadcrumbLayers PointAnnotation buildup.** Each \`<PointAnnotation>\` is implemented as a native ViewAnnotation on Android. A par-5 with 5 strokes + 2 putts has 14 PointAnnotation views; Mapbox is unhappy above ~10 per map. Combined with the stranded MapView native peers from this issue, hundreds of view-annotation handles leak per round.

## Proposed fix

Collapse \`/hole/[number]\` into a single resident screen that swaps hole data internally. Two options:

1. Move active hole into a query param (\`?h=N\`) on a stable route, OR
[...]

PR #271, opened 2026-05-13
feat(mobile): center-on-me button + course-gated GPS auto-center

cner-smith:
## Summary
- Circular `crosshairs-gps` Pressable bottom-right on `HoleMap`, **PLACE_BALL phase only**. Tap animates the camera to the smoothed GPS position (`gpsPosition` from `useHoleState`). Disabled at 50% opacity until GPS converges.
- One-shot auto-center in `useHoleCamera` after the initial hole frame, **gated by distance ≤ 1000 m (1094 yd) from the course centroid**. Testing from home won't trigger because the proximity check fails.
- Auto-center ref resets on `center` prop change (hole change). Survives the future single-screen refactor of #264 because the reset signal is the same.

## Why
From Saturday's live test: without a way to recenter, once the camera drifted (or started on the tee while the player was at their ball), the player had to manually pinch-and-pan to find themselves. Combined with the LocationPuck shipped in #268, the player can now both **see** where they are and **return to it** with one tap.

Auto-center gating addresses your home-office testing case directly: 1000 m radius around the course centroid means launching the app at your house never yanks the camera to a course you happen to have an active round on.

## Files touched
- `apps/mobile/components/round/HoleMap.tsx` — new `gpsPosition`/`courseCenter` props, `recenterOnGps` callback, conditional Pressable.
- `apps/mobile/components/round/hooks/useHoleCamera.ts` — new opts, `AUTO_CENTER_GATE_YARDS = 1094` constant, hole-change reset effect, auto-center effect.
- `apps/mobile/app/(app)/round/[id]/hole/[number].tsx` — threads `finalState.gpsPosition` and `data.courseCenter` to `<HoleMap>`.

## Test plan
- [ ] Launch dev build at home, start a live round on a real course — camera should NOT auto-snap to your home (gate fails)
- [ ] At the course, start a live round — once GPS converges, camera auto-centers once on you
- [ ] Tap the crosshairs button anywhere → camera flies to current GPS
- [ ] Advance to next hole → camera re-frames the new tee, then auto-recenters once GPS converges (because the ref reset on hole change)
- [ ] Button is hidden in SET_AIM, PIN, and TEE phases
- [ ] Button is greyed/disabled before GPS has acquired
- [ ] Ball placement, Kalman behavior, `manuallyPlacedRef` flow all unchanged

## Stacked context
Third in the Saturday-bugs series:
- #268 LocationPuck — merged
- #269 Hole route noise cleanup — merged
- #270 SET_AIM/PIN camera snap-back — merged
- **this PR** — recenter UX
- Chunk D: dev↔preview parity audit (#265) — investigation, separate work
- Chunk E: single-MapView-per-round (#264 architectural fix) — separate larger PR

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/hole/[number].tsx
+++ b/apps/mobile/app/(app)/round/[id]/hole/[number].tsx
@@ -294,6 +294,8 @@ export default function HoleScreen() {
           aim={finalState.aim}
           ball={finalState.ball}
           previousShots={data.previousShots}
+          gpsPosition={finalState.gpsPosition}
+          courseCenter={data.courseCenter}
           missingHoleLayout={data.tee == null && data.storedPin == null && data.roundPin == null}
           phase={
             pinPlacementOpen
--- a/apps/mobile/components/round/HoleMap.tsx
+++ b/apps/mobile/components/round/HoleMap.tsx
@@ -1,6 +1,7 @@
 import { useCallback, useMemo, useRef, useState } from 'react'
-import { Text, View } from 'react-native'
+import { Pressable, Text, View } from 'react-native'
 import Mapbox from '@rnmapbox/maps'
+import { MaterialCommunityIcons } from '@expo/vector-icons'
 import { Gesture, GestureDetector } from 'react-native-gesture-handler'
 import { runOnJS } from 'react-native-reanimated'
 import { distanceYards, ensureMapboxInitialized } from '../../lib/maps'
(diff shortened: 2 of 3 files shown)


cner-smith:
Superseded by #273 — bundled with #272 and the actual fix for the Android `getCurrentPositionAsync` hang. The original code here works, but only on top of the #273 GPS subscription rewrite.

Issue #305, opened 2026-05-13
iOS: stub NSCameraUsageDescription + NSPhotoLibraryAddUsageDescription in app.json pre-emptively


cner-smith:
**Severity:** low (avoids future build-edit-resubmit cycles when those features ship)
**Audit source:** mobile-app-developer agent, 2026-05-13

## Background

We don't currently use the camera or write to photo library. If we ever add (e.g. scorecard photo upload, profile picture, save scorecard to camera roll), the absence of these strings will fail the build review.

Pre-emptively adding stub strings means the first feature that touches those APIs doesn't need an app.json change + rebuild + resubmit cycle.

## Suggested copy (stub, refine when feature lands)

```json
\"ios\": {
  \"infoPlist\": {
    \"NSCameraUsageDescription\": \"OGA uses the camera to capture photos for your golf rounds.\",
    \"NSPhotoLibraryAddUsageDescription\": \"OGA needs permission to save your scorecard images to your photo library.\"
  }
}
```

If the feature never ships, the strings sit unused and have zero runtime cost (iOS only displays them when the corresponding permission is requested).

## Files

- \`apps/mobile/app.json\` (ios.infoPlist block)

## Note

Don't pre-emptively add \`NSUserTrackingUsageDescription\` (ATT) — adding it can imply we DO track and trigger ATT prompt expectations during review. Only add when ATT actually applies.

cner-smith:
## Backfill — ticket profile 2026-05-22

**Three-system shape**

| Layer | Owns |
|---|---|
| `@oga/core` | none |
| `apps/web` | none |
| `apps/mobile` | `app.config.ts` — `ios.infoPlist` block |
| `supabase/` | none |

**Acceptance criteria**

- [ ] `NSCameraUsageDescription` declared with a sensible string ("Take photos of scorecards or course features to add to your rounds")
- [ ] `NSPhotoLibraryAddUsageDescription` declared ("Save shot-pattern share cards to Photos")
- [ ] Neither permission requested at runtime today (declarative only — prevents future runtime crash if a feature lands without a config update)
- [ ] `npx expo prebuild --platform ios` regenerates Info.plist correctly
- [ ] No runtime permission prompt appears on current app flows

**Out of scope**
- Camera-based shot capture / OCR scorecard features (separate future tickets)
- Live photo / video recording (no current product use)
- Photo library *read* access (`NSPhotoLibraryUsageDescription`) — different permission, not preemptively needed

**Dependencies**
[...]

Issue #304, opened 2026-05-13
iOS: set usesNonExemptEncryption=false to skip per-build export compliance prompt


cner-smith:
**Severity:** low (saves a few minutes per submission — small ergonomic win)
**Audit source:** certification-planner agent, 2026-05-13

## Background

Apple requires every app to declare whether it uses non-exempt encryption. Without a declaration in Info.plist, every build prompts a manual questionnaire in App Store Connect during submission.

OGA uses Supabase HTTPS only — standard, exempt encryption. We can declare \`ITSAppUsesNonExemptEncryption: false\` once and never see the questionnaire again.

## Proposed fix

Add to \`apps/mobile/app.json\`:

```json
\"ios\": {
  \"config\": {
    \"usesNonExemptEncryption\": false
  }
}
```

Expo's prebuild step translates this to \`ITSAppUsesNonExemptEncryption: false\` in the generated Info.plist.

## Files

- \`apps/mobile/app.json\` (add ios.config.usesNonExemptEncryption)

cner-smith:
## Backfill — ticket profile 2026-05-22

**Three-system shape**

| Layer | Owns |
|---|---|
| `@oga/core` | none |
| `apps/web` | none |
| `apps/mobile` | `app.config.ts` — `ios.config.usesNonExemptEncryption: false` |
| `supabase/` | none |

**Acceptance criteria**

- [ ] `app.config.ts:ios.config.usesNonExemptEncryption` set to `false`
- [ ] No per-build export-compliance prompt during `eas submit`
- [ ] Generated Info.plist contains `ITSAppUsesNonExemptEncryption = NO`
- [ ] Verify accurate: app uses only standard HTTPS / TLS (exempt under US BIS encryption rules)

**Out of scope**
- Custom encryption (none in OGA)
- Cryptographic key management features (none)

**Dependencies**
- Depends on: none
- Blocks: #313 (TestFlight uploads), App Review submission (each prompt requires manual answer otherwise)
- Related: #299 (Privacy Manifest — same `app.config.ts:ios` block)

**Effort:** XS (~10 min)

**Verification**
- Generated Info.plist contains the key
- `eas submit` no longer prompts for the encryption question
- Verify against US BIS rules: HTTPS / TLS is "ancillary" use of cryptography, exempt from export classification

**Risk + rollback**
- Risk: misclassification — only if OGA ever adds custom crypto (e.g. E2E messaging, certificate pinning beyond what TLS provides)
[...]

Issue #302, opened 2026-05-13
iOS: StatusBar style='light' hardcoded — unreadable on iOS light mode


cner-smith:
**Severity:** medium (status bar text invisible/unreadable for light-mode iOS users)
**Audit source:** mobile-app-developer agent, 2026-05-13

## Mechanism

\`apps/mobile/app/_layout.tsx:193\` hardcodes \`<StatusBar style=\"light\" />\`. On Android the visual impact is muted, but iOS strictly renders the status bar text in the requested color. Light-mode iOS users get white text on whatever's behind the status bar.

## Proposed fix

Switch to \`style=\"auto\"\` so the system inverts based on user's color scheme:

\`\`\`tsx
<StatusBar style=\"auto\" />
\`\`\`

If specific screens need a forced dark/light status bar (e.g. the hole header is dark, so light text is correct there), use the per-screen \`<StatusBar>\` override from \`expo-status-bar\` inside that screen.

## Files

- \`apps/mobile/app/_layout.tsx:193\`

cner-smith:
## Validation 2026-05-22

**Verdict: real bug — confirmed.** Verified against `apps/mobile/app/_layout.tsx` on `dev`.

### Code state

`apps/mobile/app/_layout.tsx:193`:
```tsx
<StatusBar style="light" />
```

Hardcoded at the root layout. No per-screen overrides found in the mobile tree (grep for `<StatusBar` in `apps/mobile/` returns just this one render site). So light status-bar text is forced for the entire app on both platforms.

### Visual impact

- Android: low — most screens have either a dark header band or the splash/loading overlay (background `#1C211C` per `app/_layout.tsx:228`) behind the status bar. The system renders the icons over a translucent strip that mostly works visually.
- iOS light mode: high — iOS strictly renders the system status text in the requested color. Light text on a light app background (the Home tab uses `#FBF8F1` per the tab-bar style) makes the time/battery icons functionally invisible.
- iOS dark mode: fine — light text on dark backgrounds is correct.

So the ticket's framing ("medium" severity, "iOS light mode") is accurate. Android impact is incidental.

### Reasoning for fix path

`style="auto"` is the right default. It picks dark text in light mode and light text in dark mode, which matches the app's current `userInterfaceStyle: 'automatic'` (set at `app.config.ts:13`). The "right" answer is "match the current system theme", and `auto` is exactly that.

[...]

Issue #300, opened 2026-05-13
iOS: tab bar 'height: 64' ignores home indicator inset on iPhone X+


cner-smith:
**Severity:** high (tab bar overlaps home indicator on every modern iPhone)
**Audit source:** mobile-app-developer agent, 2026-05-13

## Mechanism

\`apps/mobile/app/(app)/_layout.tsx:142-149\` hardcodes \`tabBarStyle: { height: 64, paddingBottom: 10 }\`. iPhone X+ home indicator needs ~34 pt extra bottom clearance.

Expo Router's tab navigator inherits from react-navigation/bottom-tabs, which **automatically adds the safe-area inset when \`height\` is not overridden**. The fixed height is what breaks it.

## Proposed fix

Two options:

1. **Remove the fixed height entirely** — let react-navigation compute it (includes home indicator inset automatically). Simplest.
   ```ts
   tabBarStyle: {
     // remove height
     paddingBottom: 10,  // your existing bottom padding
     // ...
   }
   ```

2. **Compute height + paddingBottom from safe-area insets:**
   ```ts
   const insets = useSafeAreaInsets()
   tabBarStyle: {
     height: 54 + insets.bottom,
     paddingBottom: insets.bottom + 10,
   }
   ```

Option 1 is cleaner unless we have specific design tokens to preserve.

## Files

- \`apps/mobile/app/(app)/_layout.tsx:142-149\`

cner-smith:
## Validation 2026-05-22

**Verdict: real bug — confirmed.** Verified against `apps/mobile/app/(app)/_layout.tsx` on `dev`.

### Code state

`apps/mobile/app/(app)/_layout.tsx:142-151`:
```ts
tabBarStyle: {
  backgroundColor: '#FBF8F1',
  borderTopWidth: 1,
  borderTopColor: '#D9D2BF',
  paddingTop: 8,
  paddingBottom: 10,
  height: 64,
  elevation: 0,
  shadowOpacity: 0,
},
```

Hardcoded `height: 64` + `paddingBottom: 10`. The ticket's diagnosis is correct: react-navigation's bottom-tabs automatically pads for the home indicator inset when `height` is unset; the fixed value overrides that, so the bottom 34pt of the tab bar sits under the iOS home indicator on iPhone X+ (every modern model).

Android impact: none — Android's gesture nav area is below this region and isn't computed into `useSafeAreaInsets().bottom` the same way. So the ticket's "high severity, iOS only" framing holds.

[...]

PR #389, opened 2026-05-22
chore(mobile/ios): app config micro-fixes (#300 #302 #305)

cner-smith:
## Summary

Three small iOS config fixes bundled as one PR. All zero-risk on Android.

- **#305** — stub `NSCameraUsageDescription` + `NSPhotoLibraryAddUsageDescription` in `app.config.ts` so App Review can't reject if `expo-image-picker` / `expo-camera` lands without a config bump.
- **#302** — `<StatusBar style="light" />` → `"auto"` so the system color scheme drives the status bar (matches `userInterfaceStyle: 'automatic'` already in `app.config.ts`).
- **#300** — remove fixed `height: 64` from tab bar; react-navigation/bottom-tabs auto-adds the safe-area inset for iPhone X+ home indicator. Android keeps the same visible footprint via `paddingTop`/`paddingBottom`.

#304 was already addressed by PR #383 (`ITSAppUsesNonExemptEncryption: false`). Closing separately.

## Test plan

- [ ] CI green (typecheck, test, mobile-typecheck) — verified locally
- [ ] iOS device: status bar legible in both light and dark mode
- [ ] iOS device (iPhone X+): tab bar clears home indicator
- [ ] Android device: tab bar height visually unchanged

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -39,6 +39,14 @@ const config: ExpoConfig = {
       // false skips Apple's export-compliance prompt on every
       // TestFlight upload.
       ITSAppUsesNonExemptEncryption: false,
+      // Stubbed pre-emptively (#305). Not used at runtime today — Expo
+      // adds these automatically once expo-image-picker / expo-camera is
+      // imported, but declaring them upfront prevents an App Review
+      // rejection if either is added without a config bump.
+      NSCameraUsageDescription:
+        'OGA may use the camera to attach photos to round notes.',
+      NSPhotoLibraryAddUsageDescription:
+        'OGA may save round summary cards to your photo library.',
     },
   },
   android: {
--- a/apps/mobile/app/(app)/_layout.tsx
+++ b/apps/mobile/app/(app)/_layout.tsx
@@ -145,7 +145,9 @@ export default function AppLayout() {
           borderTopColor: '#D9D2BF',
           paddingTop: 8,
           paddingBottom: 10,
-          height: 64,
+          // Height omitted (#300) so react-navigation/bottom-tabs adds the
+          // safe-area inset automatically — required for iPhone X+ home
+          // indicator clearance.
           elevation: 0,
           shadowOpacity: 0,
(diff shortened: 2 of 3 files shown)


cner-smith:
## Review — chore(mobile/ios): app config micro-fixes (#300 #302 #305)

**Concerns**
- `apps/mobile/app.config.ts:45-50` — The stub adds `NSPhotoLibraryAddUsageDescription` (saving to the library) but not `NSPhotoLibraryUsageDescription` (reading/picking from the library). The PR description names `expo-image-picker` as the motivating package, but `expo-image-picker` in its typical pick-from-library flow requires `NSPhotoLibraryUsageDescription`, not the Add variant — Apple will still reject without it. May be intentional if only save-to-library is ever planned (round summary card export), but if picking existing photos is on the roadmap the stub covers the wrong key.

**Nits**
- This PR closes three separately-tracked issues (#300, #302, #305). CONTRIBUTING.md says "One concern per PR." The grouping is defensible — all three are zero-scope iOS config fixes in the same file or adjacent layout files — but it's worth being explicit that this is an intentional "micro-fixes bundle" exception rather than a habit.

---
Looked good: targets `dev`, no package boundary violations, no security issues, no dependency range changes, `nativewind` untouched, `StatusBar style="auto"` aligns correctly with the existing `userInterfaceStyle: 'automatic'` in app.config.ts.

---

cner-smith:
Re: NSPhotoLibraryUsageDescription concern

Intentional — Add-only stub covers the only roadmap use case I can name today: round summary share-card export (#374, 1200×630 PNG). No product surface needs library *picking* (importing existing photos as round notes etc.), and pre-emptively stubbing a key we may never need bloats the prompt the user sees on first photo-save.

If pick-from-library lands later, the consuming PR can add `NSPhotoLibraryUsageDescription` alongside the `expo-image-picker` import in one commit.

Re: One-concern-per-PR nit

Acknowledged. Intentional bundle exception — same-file zero-scope iOS config — not a habit.

PR #390, opened 2026-05-22
refactor(core): move formatBandLabel to @oga/core (#207)

cner-smith:
## Summary

Pure display formatter for `ApproachBandStat` ranges. Lives in `packages/core/src/stats.ts` (not `units.ts`) because it consumes `ApproachBandStat` — putting it in `units.ts` would invert the existing `stats.ts → units.ts` dep direction.

- Add `formatBandLabel` + 4 tests to `@oga/core`
- Strip from `apps/web/src/pages/stats/format.ts` (drops 3 unused imports)
- Update `StatTiles.tsx` (the only consumer) to import from `@oga/core`

Closes #207.

## Test plan

- [x] `pnpm typecheck` — 3 packages clean
- [x] `pnpm test --filter @oga/core` — 302 tests pass (4 new)
- [ ] Web stats page renders bucket labels unchanged

PR #391, opened 2026-05-22
refactor(core): move SGBreakdown averaging to @oga/core (#195)

### LINKED ISSUE #195: Move SGBreakdown averaging math to @oga/core
SGBreakdown.tsx:31-44 contains pure averaging math (SG_KEYS.reduce → avg per category, maxAbs) inside a UI component. Per CLAUDE.md all pure math belongs in @oga/core/stats.ts next to existing sgAverages. Extract to a named helper function in core and import from there.

cner-smith:
## Summary

Pure SG averaging math + `maxAbs` floor extracted from `apps/mobile/components/home/SGBreakdown.tsx` to `@oga/core/stats.ts`. The math lives next to existing `sgAverages`; the UI keeps its own label dictionary so the function returns DB-column keys (`sg_off_tee` etc.) instead of UI strings.

Behavior preserved exactly:
- Empty/all-null categories collapse to `0` (renders as a neutral bar at center).
- `maxAbs` floored at `0.5` so low-magnitude bars stay visible.
- Averages rounded to 2 decimal places.

## Test plan

- [x] `pnpm typecheck` — 3 packages clean
- [x] `pnpm test --filter @oga/core` — passes (7 new tests added for sgBreakdown)
- [ ] Mobile home screen renders the SG breakdown unchanged

Closes #195.

### DIFF
--- a/apps/mobile/components/home/SGBreakdown.tsx
+++ b/apps/mobile/components/home/SGBreakdown.tsx
@@ -1,6 +1,6 @@
 import { useMemo } from 'react'
 import { Text, View } from 'react-native'
-import { formatSG } from '@oga/core'
+import { formatSG, sgBreakdown, type SGBreakdownKey, type SGRoundLike } from '@oga/core'
 
 const KICKER: import('react-native').TextStyle = {
   color: '#8A8B7E',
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — refactor(core): move SGBreakdown averaging to @oga/core (#195)

**Nits**
- `packages/core/src/__tests__/sgBreakdown.test.ts:2` — the test imports from `'../stats'` (the internal module path) rather than `'@oga/core'` (the public barrel). `SGBreakdown.tsx` consumes these symbols via the barrel, so a missing re-export in `index.ts` would break the mobile app while leaving these tests green. If the rest of `@oga/core`'s tests already follow the internal-path convention this is consistent and fine to ignore; unclear from the diff alone.
- `packages/core/src/stats.ts:155` (the comment block above `sgBreakdown`) — "renders as a neutral bar at center" and "instead of disappearing into the axis" are UI-layer rationale inside a core function's comment. The non-obvious design choices — collapse to 0 for nulls, floor `maxAbs` at 0.5 — do warrant a comment per CONTRIBUTING.md's standard, but the rendering framing embeds caller knowledge here. Trimming to the invariant ("null-only category treated as 0; `maxAbs` minimum of 0.5 prevents a zero-range scale") would keep the why without the UI coupling.

---
Looked good: math moved in the correct direction (mobile → `@oga/core`), UI component correctly retains the label dictionary, label mapping and key ordering preserved exactly, seven tests cover the cases that matter (empty input, all-null, partial nulls, rounding, `maxAbs` floor and ceiling), no mixed concerns, targets `dev`.

---

cner-smith:
Re: barrel-vs-internal test import (nit 1)

Existing convention in `packages/core/src/__tests__/` is internal-path imports — `formatBandLabel.test.ts` (just merged in #390) uses the same `'../stats'` pattern, as do the other tests in that directory. Switching this one PR to barrel imports introduces inconsistency without addressing the underlying risk you flagged. If we want the barrel convention, it's worth a separate PR that flips all tests at once.

Keeping as-is for consistency. Open to switching the convention later if you want to file it.

Re: UI-layer rationale in comment (nit 2)

Fixed in the follow-up commit — comment now reads "null-only category treated as 0; maxAbs minimum of 0.5 prevents a zero-range scale when every category has very low magnitude."

Issue #301, opened 2026-05-13
iOS: rewrite location usage strings + remove unused Always permission


cner-smith:
**Severity:** high (App Review rejection vector — vague usage strings, unjustified Always permission)
**Audit source:** certification-planner agent, 2026-05-13

## Mechanism

Current \`apps/mobile/app.json\`:

\`\`\`json
\"infoPlist\": {
  \"NSLocationWhenInUseUsageDescription\": \"OGA uses your location during a round to track shots and yardages.\",
  \"NSLocationAlwaysAndWhenInUseUsageDescription\": \"OGA uses your location during a round to track shots and yardages.\"
}
\`\`\`

Two issues:
1. \`NSLocationAlwaysAndWhenInUseUsageDescription\` is **declared but we never request 'Always'** — \`useHoleState.ts:108\` only calls \`requestForegroundPermissionsAsync\`. Apple guideline: do not declare a permission you don't request. Either ship the Always-capable code (with strong justification) or remove the string.
2. The usage strings are generic and don't specify which feature uses location.

### NSLocationWhenInUseUsageDescription (keep, refine)

> \"OGA uses your location during active rounds to track shot locations on the course map. Location is only accessed while you're playing a round.\"

Alternative shorter version:

> \"Used during active rounds to track shot locations on the course map.\"

### NSLocationAlwaysAndWhenInUseUsageDescription (remove)

Delete entirely. Re-add only if/when we ship background location tracking with a clear justification (e.g. auto-shot-detection while phone in pocket).

## Updated app.json snippet

```json
\"ios\": {
  \"infoPlist\": {
    \"NSLocationWhenInUseUsageDescription\": \"OGA uses your location during active rounds to track shot locations on the course map.\"
  }
}
```

## Files

- \`apps/mobile/app.json\` (ios.infoPlist block)

cner-smith:
## Validation 2026-05-22

**Verdict: real App Review risk — confirmed, but partially acknowledged in code.**

### Code state

`apps/mobile/app.config.ts:29-42`:
```ts
infoPlist: {
  NSLocationWhenInUseUsageDescription:
    'OGA uses your location during a round to track shots and yardages.',
  // App-store-review note (#301): the app only calls
  // requestForegroundPermissionsAsync(), so the Always key is inert
  // at runtime but reviewers will flag it. Removed before App Store
  // submission, kept for dev/ad-hoc builds for now.
  NSLocationAlwaysAndWhenInUseUsageDescription:
[...]

PR #393, opened 2026-05-23
fix(mobile/ios): address #389 cross-review findings (StatusBar, plist dedupe, tab bar)

cner-smith:
## Summary

Three findings from a cross-review pass on PR #389 (the iOS micro-fixes bundle), bundled here as one same-area follow-up.

### 1. StatusBar dark-on-dark regression

\`<StatusBar style="auto"/>\` renders **black icons on a near-black background** (\`#1C211C\`) during three flows:
- Splash overlay in \`app/_layout.tsx\`
- Loading view in \`(app)/_layout.tsx\` (auth + profile fetch)
- Error view in \`(app)/_layout.tsx\` (profile fetch failure)

Fix: keep the global \`auto\` for the (light-background) app, scope \`light\` to the dark views.
- \`app/_layout.tsx\`: reactive root \`style={overlayMounted ? 'light' : 'auto'}\` so the effect re-fires when the overlay dismisses.
- \`(app)/_layout.tsx\`: \`setStatusBarStyle\` in a \`useEffect\` keyed on \`[authLoading, profileState]\`. Avoids expo-status-bar's last-mounted-wins-no-restore quirk.

### 2. Info.plist duplicate location key

\`NSLocationAlwaysAndWhenInUseUsageDescription\` was declared both in \`infoPlist\` and injected by the \`expo-location\` plugin via \`locationAlwaysAndWhenInUsePermission\`. Duplicate plist keys are undefined behavior (last-write-wins on most parsers, but not guaranteed). Removed the inline duplicate; plugin owns it. #301's larger cleanup (drop Always entirely) is unchanged.

### 3. Tab bar height shrunk on non-inset devices

Removing \`height\` entirely (#300, in #389) dropped Android + iPhone SE / 8 to 49 px (react-navigation/bottom-tabs default), a 15px shrink. Pivoted from "remove height entirely" to **explicit \`height: 54 + insets.bottom\`** via \`useSafeAreaInsets()\`. iPhone X+ still gets home-indicator clearance via \`insets.bottom\`; non-inset devices get the original visible footprint.

## Test plan

- [x] \`npm run typecheck\` mobile clean
- [ ] iOS device: confirm status bar icons stay legible during splash + loading
- [ ] iOS device (iPhone X+): tab bar still clears home indicator
- [ ] Android device: tab bar visible footprint unchanged
- [ ] iPhone SE / 8 (no notch): tab bar at original 64-ish height

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -29,12 +29,11 @@ const config: ExpoConfig = {
     infoPlist: {
       NSLocationWhenInUseUsageDescription:
         'OGA uses your location during a round to track shots and yardages.',
-      // App-store-review note (#301): the app only calls
-      // requestForegroundPermissionsAsync(), so the Always key is inert
-      // at runtime but reviewers will flag it. Removed before App Store
-      // submission, kept for dev/ad-hoc builds for now.
-      NSLocationAlwaysAndWhenInUseUsageDescription:
-        'OGA uses your location during a round to track shots and yardages.',
+      // NSLocationAlwaysAndWhenInUseUsageDescription is injected by the
+      // expo-location plugin below (locationAlwaysAndWhenInUsePermission).
+      // Don't duplicate the key here — Info.plist with duplicate keys is
+      // undefined behavior. #301 tracks removing the Always permission
+      // entirely since OGA only uses requestForegroundPermissionsAsync().
       // No proprietary encryption beyond standard TLS/Keychain.
       // false skips Apple's export-compliance prompt on every
       // TestFlight upload.
--- a/apps/mobile/app/(app)/_layout.tsx
+++ b/apps/mobile/app/(app)/_layout.tsx
@@ -2,6 +2,8 @@ import { useEffect, useState } from 'react'
 import { ActivityIndicator, Pressable, Text, View } from 'react-native'
 import { Tabs, Redirect } from 'expo-router'
 import { MaterialCommunityIcons } from '@expo/vector-icons'
+import { setStatusBarStyle } from 'expo-status-bar'
+import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import { supabase } from '../../lib/supabase'
 import { useAuth } from '../../hooks/useAuth'
 import { ErrorBoundary } from '../../components/errors/ErrorBoundary'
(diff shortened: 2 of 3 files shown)


cner-smith:
## Review — fix(mobile/ios): address #389 cross-review findings (StatusBar, plist dedupe, tab bar)

**Blockers**
- `apps/mobile/app/(app)/_layout.tsx:28–33` — The imperative `setStatusBarStyle` approach has a gap that undermines the whole point of this fix. The overlay dismissal is gated on `!authLoading` (root layout effect, line ~149), so auth is resolved before the fade starts. But the Supabase profile fetch starts concurrently; `profileState` can stay `'loading'` for up to `PROFILE_FETCH_TIMEOUT_MS` (10 s) after auth resolves. When `setOverlayMounted(false)` fires, the root re-renders and its declarative `<StatusBar style="auto">` re-pushes `auto` to the native API. The child's `useEffect` deps (`authLoading`, `profileState`) haven't changed, so it doesn't re-fire. Result: dark icons on the near-black `#1C211C` loading view for however long the profile fetch takes — exactly the regression this PR is fixing. A declarative `<StatusBar style={dark ? 'light' : 'auto'}>` inside this layout (or scoped to the loading/error branches) would join expo-status-bar's stack and survive the parent re-render without needing a dep change. The imperative approach would be safe if nothing in the parent could re-render after the child calls it, but that isn't guaranteed here.

**Concerns**
- `apps/mobile/app/(app)/_layout.tsx:161–162` — Can't fully verify from the diff whether react-navigation/bottom-tabs applies its own `useSafeAreaInsets`-based `paddingBottom` on top of an explicitly-set `tabBarStyle.paddingBottom`. If it does, `insets.bottom` is counted twice on iPhone X+, shifting tab icons visually higher than intended. The test-plan checkbox "iOS device (iPhone X+): tab bar still clears home indicator" will catch this; flagging so the review doesn't sign off before that box is checked.

---
Looked good: `app.config.ts` correctly removes the duplicate `NSLocationAlwaysAndWhenInUseUsageDescription` key and explains the expo-location plugin owns it; target branch is `dev`; no new dependencies introduced; no `any`, no unexplained eslint-disables, scope held to one area.

---

cner-smith:
Re: StatusBar imperative race (blocker)

Confirmed and fixed in the follow-up commit. Pivoted to declarative-scoped per your recommendation — `<StatusBar style="light" animated />` placed inside the loading view and error view JSX, dropping the imperative `setStatusBarStyle` effect entirely. Child StatusBars now join expo-status-bar's mount stack and survive the parent's `auto` re-render while the dark view is rendered.

Re: tab bar `paddingBottom` doubling (concern)

Acknowledged — not patching code without device verification. Test plan item 3 (iPhone X+) is the gate. If padding ends up doubled, the follow-up is `paddingBottom: 10` (no `insets.bottom` add) since the explicit `height: 54 + insets.bottom` already accounts for the safe area.

Issue #340, opened 2026-05-14
Mobile: PuttingSheet break should support uphill/downhill + L→R / R→L simultaneously (split into 2 axes)


cner-smith:
**Severity:** medium (data fidelity — every putt on a sloped green captures only half the break information)

## Symptom

The Break selector in PuttingSheet is a single-select chip group. A putt that's both **uphill** AND **left-to-right breaking** (extremely common — most sloped greens have both vertical and horizontal components) can only record one. Player has to pick which axis matters more, and the other dimension is lost.

## Why this is structurally wrong

\`break_direction\` packs two independent axes into one enum:
- **Vertical axis:** \`uphill\` / \`downhill\` / (implicit flat)
- **Horizontal axis:** \`left_to_right\` / \`right_to_left\` / \`straight\` / (legacy \`left\` / \`right\` from pre-split rows)

The schema (\`supabase/migrations/0001_initial_schema.sql:123\`, expanded in \`0009_break_direction_values.sql\`) lists 7 allowed values for one column — but those 7 are really 2 dimensions × {direction-or-null} that got serialized into a flat enum.

This is exactly the same shape as the **lie_slope** problem fixed by migration 0003, which split a single \`lie_slope\` column into \`lie_slope_forward\` (uphill/downhill axis) and \`lie_slope_side\` (ball above/below feet) and kept the original column for back-compat reads.

## Current state

**DB** (\`supabase/migrations/0001_initial_schema.sql:123\` + \`0009_break_direction_values.sql\`):
- Single \`shots.break_direction text\` column
- CHECK constraint: \`null\` OR one of \`left, right, straight, left_to_right, right_to_left, uphill, downhill\`

**Types** (\`packages/core/src/types.ts:17-27\`):
\`\`\`ts
export type BreakDirection =
  | 'left'  // legacy
  | 'right' // legacy
  | 'straight'
  | 'left_to_right'
  | 'right_to_left'
  | 'uphill'
  | 'downhill'
\`\`\`

**UI** (\`apps/mobile/components/round/PuttingSheet.tsx:320-331\`):
\`\`\`tsx
<Section title="Break">
  <View style={{ flexDirection: 'row', flexWrap: 'wrap', gap: 6 }}>
    {BREAK_OPTIONS.map((b) => (
      <Chip
        active={value.breakDirection === b.value}
        onPress={() => set('breakDirection', b.value)}
      />
    ))}
  </View>
</Section>
\`\`\`

[...]

PR #400, opened 2026-05-23
feat(core): break_direction axes split — data + types (#340 PR 1/2)

cner-smith:
## Summary

PR 1 of 2 for #340. Pure additive: migration + types + helpers + tests. Touches **zero consumers** — every existing read keeps going through the legacy `break_direction` column unchanged.

- **Migration `0028`** — adds `shots.break_direction_vertical` and `shots.break_direction_horizontal` with CHECK constraints + backfill from the legacy column. Legacy single-letter `left` / `right` map onto the break-from→to vocabulary the same way `mapBreakDirection()` in `@oga/core` already handles them.
- **Types** — `BreakDirection` marked `@deprecated`; new `BreakDirectionVertical` and `BreakDirectionHorizontal` types added.
- **Helpers** — `combinedBreakDirection` / `decombinedBreakDirection` in `@oga/core/types.ts`, mirroring the existing `combinedPuttResult` / `decombinedPuttResult` pattern. Precedence: horizontal wins when both axes are set (more frequently diagnostic for putt-line correction).
- **Tests** — 19 new tests in `breakDirection.test.ts` cover precedence, legacy letter mapping, and roundtrip behaviour for the five canonical values.

Mirrors migrations 0003 (lie_slope split) and 0006 (putt_result split). Same shape, same back-compat strategy.

## Why split into 2 PRs

This half is data-only and low-risk: nothing reads the new columns yet, so it can land cleanly even if PR 2 (mobile UI + write-path) needs iteration.

## What PR 2 will cover

- Mobile `PuttingSheet.tsx` UI redesign — two `Section`s replacing the single Break selector
- Mobile write-path — populate both new columns + derived legacy on save
- `shotRowToDraft` read-path update with legacy fallback
- Regenerate `packages/supabase/src/types.ts`
- Audit `sg.ts` / `stats.ts` / `shotInference.ts` consumers
- File companion issue for the identical web bug at `PuttFormFields.tsx:138-153`

## Test plan

- [x] `pnpm --filter=@oga/core test` — 352 tests pass (19 new)
- [x] `pnpm typecheck` — workspace green
- [ ] CI mobile typecheck — covered by `preview.yml`
- [ ] Manual `npx supabase db push` against the local stack once on a Supabase-ready dev machine

### DIFF
--- a/packages/core/src/__tests__/breakDirection.test.ts
+++ b/packages/core/src/__tests__/breakDirection.test.ts
@@ -1,0 +1,125 @@
+import { describe, it, expect } from 'vitest'
+import {
+  combinedBreakDirection,
+  decombinedBreakDirection,
+  type BreakDirection,
+} from '../types'
+
+describe('combinedBreakDirection', () => {
+  it('returns horizontal value when only horizontal is set', () => {
+    expect(combinedBreakDirection({ horizontal: 'left_to_right' })).toBe('left_to_right')
+    expect(combinedBreakDirection({ horizontal: 'right_to_left' })).toBe('right_to_left')
+    expect(combinedBreakDirection({ horizontal: 'straight' })).toBe('straight')
+  })
+
+  it('returns vertical value when only vertical is set', () => {
+    expect(combinedBreakDirection({ vertical: 'uphill' })).toBe('uphill')
+    expect(combinedBreakDirection({ vertical: 'downhill' })).toBe('downhill')
+  })
+
+  it('prefers horizontal when both axes are set', () => {
+    expect(
+      combinedBreakDirection({ vertical: 'uphill', horizontal: 'left_to_right' }),
+    ).toBe('left_to_right')
+    expect(
+      combinedBreakDirection({ vertical: 'downhill', horizontal: 'right_to_left' }),
+    ).toBe('right_to_left')
+  })
+
+  it('returns null when both axes are null or absent', () => {
+    expect(combinedBreakDirection({})).toBeNull()
+    expect(combinedBreakDirection({ vertical: null, horizontal: null })).toBeNull()
+  })
+
+  it('drops `flat` vertical when horizontal is null — no legacy equivalent', () => {
+    expect(combinedBreakDirection({ vertical: 'flat' })).toBeNull()
+    expect(
+      combinedBreakDirection({ vertical: 'flat', horizontal: null }),
+    ).toBeNull()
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — feat(core): break_direction axes split — data + types (#340 PR 1/2)

**Concerns**
- `packages/core/src/types.ts:202-255` — both helpers (`combinedBreakDirection`, `decombinedBreakDirection`) land with zero callers. CONTRIBUTING.md: "default to no, ship the minimum that satisfies the requirement, then stop." These could move to PR 2 alongside the write-path and `shotRowToDraft` read-path that will consume them, keeping this PR strictly migration + types. The two-PR split is well-reasoned and the helpers double as executable documentation of the backfill semantics, so this may be intentional — raising it rather than blocking on it.

**Nits**
- `packages/core/src/types.ts:19-22` — the `@deprecated` JSDoc's trailing sentence ("Migration 0028 added the split columns; this single-axis enum is kept only for back-compat reads") narrates what rather than why. CONTRIBUTING.md: "Don't narrate what the code does." The `@deprecated` tag plus the link to the two new types is sufficient for IDE tooling; the prose could stop after "not one."

---
Looked good: targets `dev`, migration `0028` correctly follows `0027`, column additions to an existing table don't require new RLS policies, backfill SQL is semantically consistent with `decombinedBreakDirection`'s switch cases, 19 tests cover all canonical values + legacy mappings + roundtrips + edge cases (`flat` vertical, null/undefined inputs), no `any` and no `eslint-disable`.

---

cner-smith:
Thanks for the review.

**Nit fixed.** Trimmed the `BreakDirection` JSDoc to stop after "not one." — pushed in `b62f329`. The `@deprecated` tag and the `{@link}` targets are enough for IDE tooling; the trailing prose was redundant.

**Helpers stay.** Real reason they ship in PR 1 rather than PR 2: they're the executable validation of the migration's backfill semantics. The `CASE` statement in `0028_break_direction_axes.sql` and the `switch` in `decombinedBreakDirection` encode the same mapping (legacy `left` → `right_to_left`, etc.); the 19 tests assert the JS side, and the SQL side is consistent with it by construction. If the helpers move to PR 2, the migration's backfill lands without any automated test exercising the same value table — vitest is the only test framework in this repo, no in-SQL test runner. Worth the ~24 hr of zero-caller exposure to keep the data + verification slice atomic.

PR #401, opened 2026-05-23
feat(mobile): PuttingSheet break — split into slope + line axes (#340 PR 2/2)

### LINKED ISSUE #340: Mobile: PuttingSheet break should support uphill/downhill + L→R / R→L simultaneously (split into 2 axes)

cner-smith:
## Summary

PR 2 of 2 for #340. Wires PR #400's data layer (migration 0028 + the `combined` / `decombined` helpers) into the mobile shot logger so a putt on a sloped green can capture **both** axes simultaneously — slope (uphill / flat / downhill) AND line (L→R / straight / R→L).

Before: single-select chip group; player picks one axis, loses the other.
After: two independent chip rows, both nullable, same toggle-to-clear pattern as the Distance / Direction selectors right above them.

## What this PR changes

| Layer | File | Change |
|---|---|---|
| Types | `packages/supabase/src/types.ts` | Hand-added `break_direction_vertical` + `break_direction_horizontal` to `shots` Row / Insert / Update (matches what `supabase gen types` would emit after 0028). |
| Domain | `packages/core/src/round.ts` | `DraftShot` gains two-axis fields. `shotRowToDraft` reads the new columns with `decombinedBreakDirection` fallback. Legacy `breakDirection` kept `@deprecated` for web back-compat. |
| Tests | `packages/core/src/__tests__/shotRowToDraft.test.ts` | +5 tests covering new columns, legacy decomposition, single-letter mapping, new-over-legacy precedence, empty state. |
| Tests | `packages/core/src/__tests__/sg.test.ts` | Row-builder defaults for the two new required columns. |
| Mobile UI | `apps/mobile/components/round/PuttingSheet.tsx` | Single Break Section → two Sections (`Break (slope)` and `Break (line)`). |
| Mobile | `apps/mobile/components/round/ShotLogger.tsx` | `ShotLoggerValue` + pass-through. |
| Mobile | `apps/mobile/components/round/hole/useShotActions.ts` | Writes all three columns on save: `break_direction` (combined, back-compat), `break_direction_vertical`, `break_direction_horizontal`. |

## What's intentionally out of scope

- **GreenDiagram rendering both axes.** Still receives the legacy single-axis `BreakDirection` (via `combinedBreakDirection`). When both axes are set, the diagram shows the horizontal axis only (preferring it per the helper's precedence rule). v1.1 follow-up.
- **Web parity.** `apps/web/src/components/rounds/shots/PuttFormFields.tsx:138-153` has the identical single-select bug. Filed as a companion issue rather than bundled here — keeps this PR mobile-only and easy to review.
- **Backfill of old rows.** Migration 0028's `UPDATE` handles existing rows. New rows write all three columns. No second backfill pass needed.

## Verification

- [x] `pnpm --filter=@oga/core test` → **357 / 357** (5 new in `shotRowToDraft.test.ts`)
- [x] `pnpm typecheck` → workspace green
- [x] `cd apps/mobile && npm run typecheck` → green
- [ ] CI mobile typecheck via `preview.yml`
- [ ] Manual: log an uphill + R→L putt on a sloped green, confirm both axes persist on round detail

Closes #340 once dev → main releases.

### DIFF
--- a/apps/mobile/components/round/PuttingSheet.tsx
+++ b/apps/mobile/components/round/PuttingSheet.tsx
@@ -10,7 +10,9 @@ import {
 } from 'react-native'
 import {
   FEET_TO_CM,
-  type BreakDirection,
+  combinedBreakDirection,
+  type BreakDirectionHorizontal,
+  type BreakDirectionVertical,
   type GreenSpeed,
   type LieType,
   type PuttDirectionResult,
@@ -24,7 +26,8 @@ export interface PuttingValue {
   puttMade?: boolean
   puttDistanceResult?: PuttDistanceResult
   puttDirectionResult?: PuttDirectionResult
-  breakDirection?: BreakDirection
+  breakDirectionVertical?: BreakDirectionVertical
+  breakDirectionHorizontal?: BreakDirectionHorizontal
   puttSlopePct?: number // 0-4 intensity bucket
   greenSpeed?: GreenSpeed
   aimOffsetInches?: number
@@ -64,12 +67,15 @@ const DIRECTION_OPTIONS: { value: PuttDirectionResult; label: string }[] = [
   { value: 'right', label: 'Missed right' },
 ]
 
-const BREAK_OPTIONS: { value: BreakDirection; label: string }[] = [
+const BREAK_SLOPE_OPTIONS: { value: BreakDirectionVertical; label: string }[] = [
+  { value: 'uphill', label: 'Uphill' },
+  { value: 'flat', label: 'Flat' },
+  { value: 'downhill', label: 'Downhill' },
+]
+const BREAK_LINE_OPTIONS: { value: BreakDirectionHorizontal; label: string }[] = [
   { value: 'left_to_right', label: 'L → R' },
   { value: 'straight', label: 'Straight' },
   { value: 'right_to_left', label: 'R → L' },
-  { value: 'uphill', label: 'Uphill' },
-  { value: 'downhill', label: 'Downhill' },
 ]
 
 const SLOPE_INTENSITY_LABELS = ['Flat', 'Slight', 'Moderate', 'Strong', 'Severe']
@@ -104,7 +110,8 @@ export function PuttingSheet({
     puttMade: initial?.puttMade,
     puttDistanceResult: initial?.puttDistanceResult,
     puttDirectionResult: initial?.puttDirectionResult,
-    breakDirection: initial?.breakDirection ?? 'straight',
+    breakDirectionVertical: initial?.breakDirectionVertical,
+    breakDirectionHorizontal: initial?.breakDirectionHorizontal,
     puttSlopePct: initial?.puttSlopePct ?? 0,
     greenSpeed: initial?.greenSpeed,
     aimOffsetInches: initial?.aimOffsetInches ?? 0,
@@ -151,6 +158,20 @@ export function PuttingSheet({
       ...prev,
       puttMade: false,
       puttDirectionResult: prev.puttDirectionResult === v ? undefined : v,
+    }))
+  }
+
+  function setBreakSlope(v: BreakDirectionVertical) {
+    setValue((prev) => ({
+      ...prev,
+      breakDirectionVertical: prev.breakDirectionVertical === v ? undefined : v,
+    }))
+  }
+
+  function setBreakLine(v: BreakDirectionHorizontal) {
+    setValue((prev) => ({
+      ...prev,
+      breakDirectionHorizontal: prev.breakDirectionHorizontal === v ? undefined : v,
     }))
   }
 
@@ -248,7 +269,12 @@ export function PuttingSheet({
         <GreenDiagram
           distanceFt={distance}
           aimOffsetInches={value.aimOffsetInches ?? 0}
-          breakDirection={value.breakDirection ?? 'straight'}
+          breakDirection={
+            combinedBreakDirection({
+              vertical: value.breakDirectionVertical,
+              horizontal: value.breakDirectionHorizontal,
+            }) ?? 'straight'
+          }
           onAimChange={(n) => set('aimOffsetInches', n)}
         />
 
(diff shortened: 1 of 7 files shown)


cner-smith:
## Review — feat(mobile): PuttingSheet break — split into slope + line axes (#340 PR 2/2)

**Concerns**
- `packages/core/src/round.ts:186-193`: `BreakDirectionVertical` includes `'flat'`, which has no equivalent in the legacy `BreakDirection` union and therefore no path through `decombinedBreakDirection` back to it. I cannot read migration 0028 from this diff (it lives in PR #400's branch) to confirm the `break_direction_vertical` column's CHECK constraint includes `'flat'`. The comment at the top of `shotRowToDraft` notes that Postgres CHECKs keep columns inside their union vocabularies — if 0028's constraint doesn't list `'flat'`, tapping "Flat" in the slope section and saving will produce a DB error at runtime. Worth a quick cross-check against the migration before this lands.

**Nits**
- `packages/core/src/round.ts:186-193`: `decombinedBreakDirection(s.break_direction as BreakDirection | null)` is called twice — once to pull `.vertical`, once to pull `.horizontal`. The helper is pure and a single switch, so there's no real cost, but `const { vertical, horizontal } = decombinedBreakDirection(...)` extracted before the two fields would read more clearly and remove the repeated cast.

---
Looked good: five new tests cover the full read-path matrix (direct columns, legacy fallback, single-letter mapping, new-over-legacy precedence, empty state); `combinedBreakDirection` / `decombinedBreakDirection` live in `@oga/core` not in mobile lib; legacy `break_direction` is still written on save for web back-compat; edit-save round-trip preserves all legacy single-axis values correctly; PR targets `dev`, scope held to one concern.

---

cner-smith:
Thanks for the review.

**Nit fixed.** Hoisted `decombinedBreakDirection(s.break_direction)` into a `breakAxes` prep var alongside the existing `legacy` (lie slope) and `puttResult` prep vars — pushed in `325d2e0`. The legacy column gets cast once now and the helper runs once instead of twice.

**`'flat'` concern — safe to land.** I just confirmed both halves:

1. **New column** — Migration 0028 (now on `dev`):
   ```sql
   add column break_direction_vertical text check (
     break_direction_vertical is null
     or break_direction_vertical in ('uphill', 'downhill', 'flat')
   )
   ```
   `'flat'` is in the CHECK, so writing `'flat'` to `break_direction_vertical` is valid.
2. **Legacy column** — `combinedBreakDirection({vertical: 'flat', horizontal: null})` returns `null`. The `'flat'` value drops at the boundary; it never reaches the legacy `break_direction` column. The early `if (args.horizontal === ...)` branches handle the both-set case (horizontal wins). The PR #400 test `combinedBreakDirection › drops 'flat' vertical when horizontal is null — no legacy equivalent` covers this.

So tapping "Flat" + saving writes:
- `break_direction_vertical = 'flat'` ✓ (CHECK allows it)
- `break_direction = null` ✓ (legacy CHECK doesn't include `'flat'`, but we never try)

Issue #243, opened 2026-05-07
Shot patterns: curved ball flight visualization with correct shot shape rendering


cner-smith:
Replace or augment the current scatter plot on the shot patterns page with a top-down curved ball flight visualization.

## Design reference
Designed and approved in Claude conversation, May 2026.

## How it works
Each shot is rendered as a quadratic bezier curve:
- Start point: shot start_lat/lng (normalized to tee at bottom center)
- Control point: aim_lat/lng (determines curve direction)
- End point: end_lat/lng (landing position)

This correctly renders shot shapes:
- Fade: control point LEFT of target line, lands RIGHT
- Draw: control point RIGHT of target line, lands LEFT
- Push: straight line going right, minimal curve
- Pull: straight line going left, minimal curve
- Hook/Slice: exaggerated versions of draw/fade

## Visual elements
- Individual shot curves: amber (on target) or red/coral (miss)
- Average shot: thick green bezier curve
- 68% dispersion ellipse: solid border
- 95% dispersion ellipse: dashed border
- Landing dots at end of each curve
- Target line reference (dashed, vertical)
- 'starts L / lands R' annotations for fade (or equivalent for other shapes)

## Coordinate system
All GPS coordinates normalized so:
- Tee always at bottom center
- Target (pin) always at top center
- Lateral offset calculated relative to aim line
- Shot shape inferred from relationship between start, aim, and end points

## Stats row below chart
- Avg carry distance
- Shot shape label (Fade / Draw / Push / Straight etc)
- Avg lateral bias (e.g. +8 yd R)
- Lateral spread (±18 yd)

## Data sources (all already stored)
- shots.start_lat/lng
- shots.end_lat/lng
- shots.aim_lat/lng (control point for bezier)
- shots.club (for filtering by club)
- shots.shot_result (for coloring on target vs miss)

## Note on side profile view
A side profile arc was considered but dropped — OGA does not track launch angle or apex height so any arc would be a generic estimate not the player's actual data. Top-down only is the correct approach with current data.

cner-smith:
<img width="1440" height="1412" alt="Image" src="https://github.com/user-attachments/assets/0a02807e-6bab-4fd7-8a9f-e053bd90a83d" />

cner-smith:
## Implementation notes (drop-in patterns)

### Math
Quadratic Bezier with the aim point as the control point is the right model — shot shape is preserved by the geometric relationship of the three points:
- Curve length ≈ shot distance (start to end)
[...]

PR #404, opened 2026-05-23
feat(patterns): standalone ball-flight chart on Shot Patterns (#243)

cner-smith:
## What

Adds a **standalone driving-range ball-flight chart** below the dispersion plot on Shot Patterns. The dispersion plot itself is **unchanged**. Implements #243.

The two charts now show complementary things:
- **Dispersion plot (above):** aim-relative miss (left/right, short/long around target) + the 68/95 rings.
- **Ball-flight chart (below):** absolute **distance and shape** — each shot plotted at its true carry up a yardage scale, curving left/right for draw/fade.

## How

**@oga/core** — `computeDispersion` projects the start position into the aim-relative plane via optional `startLateralOffsetYards` / `startDistanceOffsetYards` (set only when start coords exist), so each shot can be placed at its real distance. Additive + backward-compatible.

**apps/web** — co-located `BallFlightChart`: adaptive 25/50-yd gridlines (range backdrop), each shot's flight at its true carry, draw/fade curve, average flight line. Isotropic scale so distance and lateral spread stay in true proportion. Falls back to a note when average carry is too small to plot. Reuses `pointColor`.

## Build-on

Rebased onto `dev` after #406 (2D ellipse fix) merged.

## Test

- `@oga/core`: 2 vitest cases for the start projection (present + absent)
- Typecheck green
- Verified on seeded e2e data — see latest screenshots in the comments

## Known data caveat (not a chart bug)

The seed assigns "driver" to every long shot across all hole lengths (150→500 yd), so its range view looks like a tall streak. Mid-irons (only picked for their own distance band) render as tight clusters. Real data → compact clouds. Separate seed-realism follow-up.

### DIFF
--- a/apps/web/src/pages/patterns/ShotPatternsPage.tsx
+++ b/apps/web/src/pages/patterns/ShotPatternsPage.tsx
@@ -658,7 +658,7 @@ function BallFlightChart({
     )
   }
 
-  const x = (lat: number) => cx + lat * scale
+  const x = (lateralYards: number) => cx + lateralYards * scale
   const y = (dist: number) => teeY - dist * scale
   // Each shot's true forward distance from the tee.
   const carryOf = (p: DispersionPoint) =>
--- a/apps/web/src/pages/patterns/ShotPatternsPage.tsx
+++ b/apps/web/src/pages/patterns/ShotPatternsPage.tsx
@@ -17,6 +17,10 @@ import { useUnits } from '../../hooks/useUnits'
 const SVG_SIZE = 420
 const SVG_VIEW_WIDTH = `min(${SVG_SIZE}px, 90vw)`
 
+const FLIGHT_W = 460
+const FLIGHT_H = 520
+const FLIGHT_VIEW_WIDTH = `min(${FLIGHT_W}px, 92vw)`
+
 export function ShotPatternsPage() {
   const { unit, toDisplay } = useUnits()
   const [club, setClub] = useState<Club>('7i')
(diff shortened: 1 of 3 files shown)


cner-smith:
### Ball-flight tracers — rendered on seeded e2e data

**9i (22 shots)** — clean fan, average tracer in green, 68/95 ellipses around the landing cluster:

9i ball flight

**driver (128 shots)** — high-N density; tracers at 0.3 opacity so the average curve + dots stay readable:

driver ball flight

Tracers begin at the synthetic bottom tee-line (lateral true-to-scale, distance pinned), curve through the aim point, and land in the dispersion cluster. Legacy rows without a recorded start fall back to plain dots.

cner-smith:
## Review — feat(patterns): ball-flight bezier tracers on Shot Patterns (#243)

**Concerns**

- `packages/core/src/shot-patterns.ts:23,75,82` / `packages/core/src/shot-patterns.test.ts:65,78` — `startDistanceOffsetYards` is computed, stored on `DispersionPoint`, and tested, but has zero consumers anywhere in the codebase. The web rendering pins the start y-coordinate to `teeY` (a fixed constant at the SVG bottom edge) and never calls `py(p.startDistanceOffsetYards)`. CONTRIBUTING.md is explicit: "No config vars, options, or parameters added 'for hypothetical future use' — add them when there's a real caller." The PR description confirms the tee-at-bottom full-distance layout is deferred, which makes this field speculative scope. Drop the field and its test case; re-add when a rendering path actually consumes it.

- `apps/web/src/pages/patterns/ShotPatternsPage.tsx:511-520` — The average tracer renders whenever `stats` is truthy, but `avgStartLat` silently falls back to `0` when no shots have a recorded start position. The result is a phantom curve drawn from center-bottom (`px(0), teeY`) to the average landing with no corresponding per-shot tracers alongside it — a misleading artifact for any club where all historical rows predate start-coord recording. The per-shot loop correctly gates on `p.startLateralOffsetYards == null`; the average tracer needs the same guard. One approach: surface `startLats.length` out of the `useMemo` return and wrap the average `<path>` in `{stats && startLatsCount > 0 && (…)}`.

---
Looked good: core logic stays in `@oga/core` with tests covering both branches (start present / absent), package boundaries respected, targets `dev`, single concern held throughout.

---

cner-smith:
### Reworked per review — now a standalone chart, decoupled from the dispersion plot

The dispersion plot is back to its original form (untouched). The ball flight is its own chart below it, following the mockup.

**driver (128 shots)** — tee at bottom, curves through the aim line to the landing cluster at the pin, avg carry labelled:

driver standalone flight

**9i — graceful degenerate case.** Seeded approach shots chain start→previous-landing (≈ at the pin), so carry ≈ 0 and there's no flight to plot; the chart shows a note rather than an empty box. Real approach data (start ~150 yd out) plots normally:

9i degenerate note

Supersedes the earlier overlay screenshots above.

cner-smith:
### Latest: driving-range view (genuine distances + yardage scale)

Reworked again — the flight chart no longer duplicates the dispersion plot. It now shows **absolute distance + shape** against a yardage scale, with each shot at its **true carry** (no avg-carry normalization), so the real spread shows and the chart grows to fit.

**6i (18 shots)** — gridlines at 50/100/150, shots fanning to a cluster at the ~150 yd carry:

6i range view

**driver (128 shots)** — tall because the seed assigns driver to 150→500 yd shots across all holes; the gridlines now make that *readable* rather than a mystery streak. Real driver data would be a compact cluster like the 6i:

driver range view

Supersedes earlier screenshots.

cner-smith:
## Review — feat(patterns): standalone ball-flight chart on Shot Patterns (#243)

**Concerns**
- `packages/core/src/shot-patterns.ts:19-22` and `:75-76` — `startLateralOffsetYards` is added to `DispersionPoint` and computed in `computeDispersion`, but nothing in this diff reads it. `BallFlightChart` uses only `startDistanceOffsetYards` (for the carry fallback); the lateral start field is never referenced in any chart or display logic. CONTRIBUTING.md is explicit: "No config vars, options, or parameters added 'for hypothetical future use' — add them when there's a real caller." Drop the field and the lateral assertion in the test (`expect(p!.startLateralOffsetYards!).toBeLessThan(0)`) until something actually consumes it.

**Nits**
- `apps/web/src/pages/patterns/ShotPatternsPage.tsx:641` — `const x = (lat: number) => cx + lat * scale` — this parameter is lateral yards, but `lat` reads as latitude in a codebase saturated with `lat`/`lng` coordinate naming. `lateralYards` or just `yards` avoids the false signal.

---
Looked good: targets `dev`, `BallFlightChart` stays co-located in the same file (the 190-line SVG block would genuinely be unreadable inline), `startDistanceOffsetYards` — the only new core field the chart actually uses — has proper test coverage for both the present and absent cases.

---

Issue #405, opened 2026-05-23
Mobile signup: Cloudflare Turnstile stuck at 'verifying' on iOS only


cner-smith:
## Symptom

On the mobile **signup** screen, the Cloudflare Turnstile widget sticks at "Verifying…" and never completes, so `captchaToken` is never set and **Create account stays disabled** — sign-up is impossible.

**Confirmed iOS-only.** Works correctly on web and on Android. Reproduced on a physical iOS device (first real-hardware iOS build, 2026-05-23).

## Mechanism (receipts)

- `apps/mobile/app/(auth)/signup.tsx` renders a `react-native-webview` (line ~107) pointed at `https://oga.golf/captcha.html?siteKey=…`
- `apps/web/public/captcha.html` loads `https://challenges.cloudflare.com/turnstile/v0/api.js` and calls `turnstile.render()`; on success it `postMessage`s `{type:'success', token}` back, which the RN `onMessage` turns into `captchaToken`
- "Verifying…" forever = the challenge never reaches `callback` (and not `error-callback`, or we'd get an error message)

## Root cause (high confidence)

Because web + Android both work, the shared pieces are exonerated: the site-key domain allowlist and `captcha.html` itself are fine. That isolates the failure to **iOS WKWebView**.

Leading cause: **WKWebView third-party storage / cookie restrictions.** The Turnstile challenge runs in an iframe from `challenges.cloudflare.com` — third-party relative to `oga.golf`. iOS WKWebView blocks third-party storage far more aggressively than Android, which strands a managed challenge at "verifying." The `<WebView>` currently sets no iOS storage props (`sharedCookiesEnabled`, etc.).

## Likely fix directions (not yet implemented)

- Add iOS WebView props (`sharedCookiesEnabled`, evaluate `thirdPartyCookiesEnabled` semantics) and confirm storage availability inside the challenge frame
- If WKWebView storage partitioning can't be satisfied, consider an alternative captcha delivery for native (e.g., native Turnstile integration, or a server-side challenge exchange)

## Repro

1. iOS device, fresh signup screen
2. Enter username / email / password
3. Turnstile widget shows "Verifying…" and never resolves → Create account stays disabled

## Scope

iOS-only; gates the iOS launch (v0.6 / v1.1 track), not the web/Android v1.0 path.

cner-smith:
**Root cause identified — see #407.** Likely **not** third-party storage.

[...]

Issue #60, opened 2026-05-01
Improve offline sync robustness (backoff, conflict resolution, lock TTL)


cner-smith:
Current sync in apps/mobile/lib/sync.ts has known issues (#26 N+1, #29 stuck lock). As OGA grows, the sync layer needs to be more robust for real golf conditions (spotty GPS, tunnel under bridge, cart barn).

Improvements:
- Exponential backoff on failed sync attempts
- Lock TTL — if inFlight has been true for >30s, assume crash and reset
- Conflict resolution strategy — server wins vs last-write-wins per field
- Sync status indicator in UI (synced / syncing / pending X shots)
- Chunked batch inserts (already in #26) with progress tracking

Future: if group rounds are added, Supabase Realtime for live score sharing between playing partners.

cner-smith:
Fixed in PR #139 (merged to dev).

PR #407, opened 2026-05-24
fix(mobile): allow about: scheme in auth WebView so iOS Turnstile completes (#405)

cner-smith:
## What
Adds the `about:` scheme to the auth WebView `originWhitelist` on the mobile signup + login screens — fixes the iOS-only Cloudflare Turnstile hang.

## Root cause
Turnstile renders its managed challenge into an `about:blank` / `about:srcdoc` iframe. On iOS, `react-native-webview`'s `originWhitelist` filters **all** sub-frame loads (Android only gates top-level navigation). The prior `['https://*','http://*']` silently cancelled the challenge frame, so the widget hung at "Verifying…" forever with no error callback. Web + Android were unaffected.

This corrects the original theory in #405 (third-party storage / `sharedCookiesEnabled`): that prop only bridges the app cookie jar, not iframe storage, and a storage failure would *error*, not hang.

## Fix
`originWhitelist={['https://*', 'http://*', 'about:']}` on both screens. Additive — cannot affect the working web/Android paths.

## Receipts
- Cloudflare mobile WebView docs — must allow `about:blank` / `about:srcdoc`: https://developers.cloudflare.com/turnstile/get-started/mobile-implementation/
- react-turnstile #60 — identical iOS-only "spins forever", fixed by whitelisting `about:`
- react-native-webview #2567 — on v13.10.5 (~our 13.8.6), `['*']` did not cover `about:`; explicit `'about:'` required

## Testing
No local repro (no iOS device at fix time). Validate on a **captcha-on** iOS build: restore the preview `EXPO_PUBLIC_TURNSTILE_SITE_KEY` env var, then `eas build -p ios --profile preview` — expect the widget to resolve and the submit button to enable.

Refs #405

### DIFF
--- a/apps/mobile/app/(auth)/login.tsx
+++ b/apps/mobile/app/(auth)/login.tsx
@@ -97,7 +97,7 @@ export default function Login() {
         {captchaEnabled && (
           <WebView
             source={{ uri: `https://oga.golf/captcha.html?siteKey=${encodeURIComponent(TURNSTILE_SITE_KEY ?? '')}` }}
-            originWhitelist={['https://*', 'http://*']}
+            originWhitelist={['https://*', 'http://*', 'about:']}
             onMessage={(event) => {
               try {
                 const msg = JSON.parse(event.nativeEvent.data)
--- a/apps/mobile/app/(auth)/signup.tsx
+++ b/apps/mobile/app/(auth)/signup.tsx
@@ -106,7 +106,7 @@ export default function Signup() {
         {captchaEnabled && (
           <WebView
             source={{ uri: `https://oga.golf/captcha.html?siteKey=${encodeURIComponent(TURNSTILE_SITE_KEY ?? '')}` }}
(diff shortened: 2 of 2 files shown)


cner-smith:
## Review — fix(mobile): allow about: scheme in auth WebView so iOS Turnstile completes (#405)

**Concerns**
- `apps/mobile/app/(auth)/login.tsx:100`, `apps/mobile/app/(auth)/signup.tsx:109` — The fix is well-researched and the cited references (Cloudflare mobile WebView docs, react-turnstile #60, react-native-webview #2567) support the diagnosis, but it has not been run on an actual iOS device (the PR description says so explicitly). The EAS build instructions in the PR description are the right path — someone should validate on a captcha-on iOS build before this merges.

---
Looked good: Minimal additive change; single concern held; targets `dev`; no dependency bumps; `about:` is the correct minimum ask (inert scheme, no external navigation possible); both auth screens updated consistently.

---

PR #422, opened 2026-05-25
Release: dev → main — Learn catalog complete + accumulated dev work

cner-smith:
Cuts the accumulated `dev` work to production (`oga.golf`). Ends roadmap Phase 1.

## Headline
- **Learn catalog is complete** — every article built (ships as **draft** by design; "needs review" footers stay visible on prod, cleared up at the v1.0 release).
- Ball-flight chart (#243), dispersion-cone 2-D fix, mobile PuttingSheet 2-axis split, `@oga/core` polish sweep, and the post-review tracer-direction + cluster-centered-view fixes (PR #423).

## Closes (verified open + completed)
Closes #128, closes #129, closes #130, closes #131, closes #135, closes #142, closes #206, closes #340, closes #394, closes #395, closes #397, closes #398

- **#128–#142** — six Learn-expansion articles (built, merged as drafts).
- **#206** — `categorizeShot` → `@oga/core` (PR #408).
- **#340** — mobile PuttingSheet split into distance + direction axes (PRs #400/#401).
- **#394 #395 #397 #398** — `@oga/core` polish sweep (PR #403).
- Already closed, no keyword needed: #195 #205 #207 #243 #396.

## Intentionally NOT closed (v1.1 track / unvalidated)
- **#300 #301 #302 #305** — iOS config issues (v0.6 milestone, paused). Config work landed (PRs #389/#393/#399), but iOS validation belongs to the v1.1 track — let them close when iOS resumes.
- **#405** — iOS Turnstile. The #407 fix is unvalidated (preview Turnstile key still disabled); don't auto-close until verified on device.

## Draft articles
Will be user-visible on prod with "Draft, needs review" footers — confirmed acceptable for this release.

PR #423, opened 2026-05-25
fix(patterns): tracer curve direction + cluster-centered view; pin barrel exports

cner-smith:
Addresses the two nits from the PR #422 review.

## Tracer curve direction (nit 2)
`BallFlightChart` set every shot's quadratic-bezier control point x to the chart center (`cx`), so an off-center shot bulged toward the middle before bending out — a persistent left miss tracered **right-first**. Control x now tracks the landing lateral (`x(p.lateralOffsetYards)` / `x(stats.avgLateralOffset)`), so arcs curve toward where the ball actually finished.

## Cluster-centered view (your refinement)
The horizontal view no longer forces the aim line to dead center. It re-centers on the average lateral offset, so a consistent left miss **nudges the whole view left** and the aim line (lateral 0) sits off-center. The scale half-width still includes the tee origin and the 95% cone, so nothing leaves the frame. The tee marker, aim line, and label all move to the re-centered origin (`teeX = x(0)`).

## Barrel test (nit 1)
Pinned `getShotMarkerCategory` and `formatBandLabel` — the two new **function** exports missing from the list. (The reviewer also flagged `ShotMarkerCategory`, but it's a **type**; tsc already catches its barrel drift via consumers, and it can't go in a runtime `typeof` assertion. `sgBreakdown`/`barScale` were already pinned, so the gap was 2, not 5.)

## Verify
- `pnpm typecheck` clean; `pnpm --filter @oga/core test` green (391 tests).
- Tracer direction + view-shift are visual — worth an eyeball on the Vercel preview with a left-biased shot set.

Once merged to `dev`, release PR #422 (dev→main) picks these up automatically.

### DIFF
--- a/apps/web/src/pages/patterns/ShotPatternsPage.tsx
+++ b/apps/web/src/pages/patterns/ShotPatternsPage.tsx
@@ -606,7 +606,7 @@ function BallFlightChart({
 }) {
   const { toDisplay } = useUnits()
 
-  const { avgCarry, scale, maxDist } = useMemo(() => {
+  const { avgCarry, scale, maxDist, viewCenterLat } = useMemo(() => {
     const carries = points
       .map((p) => p.startDistanceOffsetYards)
       .filter((v): v is number => v != null)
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — fix(patterns): tracer curve direction + cluster-centered view; pin barrel exports

**Concerns**
- `CONTRIBUTING.md` / one-concern-per-PR — the barrel test addition (`packages/core/src/__tests__/barrel.test.ts:12-13`) and the chart rendering changes (`apps/web/src/pages/patterns/ShotPatternsPage.tsx`) are separate concerns in separate packages with separate failure modes and revert paths. CONTRIBUTING.md: "One concern per PR." Sharing a prior-review thread doesn't make them the same concern. Both are two-line changes; splitting them is low friction.

**Nits**
- `apps/web/src/pages/patterns/ShotPatternsPage.tsx:671` — the `teeX` comment opens with "Tee origin = the aim line (lateral 0)" which narrates what `teeX = x(0)` already says. CONTRIBUTING.md: "Don't narrate what the code does." The second sentence ("Once the view re-centers…this is no longer the chart center") is doing real work — worth keeping if the comment survives, but the first sentence can go.
- `apps/web/src/pages/patterns/ShotPatternsPage.tsx:717` — `{/* aim line (straight-ahead from the tee) — lateral 0, not the chart center */}`: the "not the chart center" clause records a diff artifact rather than informing a future reader. Trim to `{/* aim line — lateral 0 */}`.

---
Looked good: bezier control-point x now tracks landing lateral so Bx(t) moves monotonically from teeX to x(lateralOffset) — no reverse bulge for any shot direction; `Math.abs(viewCenterLat)` in `lateralExtent` is the right constraint to keep teeX in frame (the math holds); empty-points spread into `Math.max` is safe because the literal `8` floor is always present; targets `dev`.

---

Issue #317, opened 2026-05-13
v1.0 blocker: email round summary feature — highest-leverage retention lever


cner-smith:
**Severity:** v1.0 blocker per launch strategy
**Source:** launch-strategist agent, 2026-05-13

## Why this is the #1 pre-launch feature

From the launch strategy audit:

> The single highest-leverage thing you can do beyond the launch plan: **ship the email round-summary feature before launch.** It's the retention lever that converts curious-installer into weekly-user, and weekly-users are the only ones who eventually donate.

OGA has no paywall — retention = utility. The morning-after email is the strongest mechanism for utility-driven retention because it:
- Reminds the user OGA exists at peak relevance (right after they played)
- Surfaces insight (strokes gained breakdown) when motivation to improve is highest
- Hosts the donation CTA in a non-intrusive context (footer of an actually-useful email)
- Drives the user back to the app for the practice plan / next-round prep

This is the lever Arccos and Shot Scope under-invest in because it doesn't drive subscriptions — for OGA's model it's the conversion engine for ongoing engagement and eventually donations.

### Trigger

Round completion (\`completeRound\` in \`apps/mobile/lib/completeRound.ts\` and the web equivalent) → enqueue email send for ~6-12 hours after round end (next morning local time).

Approach: Supabase Edge Function on a cron schedule that queries for rounds completed in the last 24h that haven't had a summary email sent, then sends via an email API (Resend / Postmark / SendGrid).

### Email content

| Section | Detail |
|---|---|
| **Subject** | \"Your round at {course}: {score} ({+/-X} to par)\" |
| **Hero stat** | Score + to-par, holes played, total strokes |
| **Strokes gained breakdown** | Off-tee / Approach / Around-green / Putting — values + visual bar |
| **Best hole + worst hole** | Hole #, par, score, brief context |
| **One thing to practice** | Generated from the SG category with the worst differential ((e.g. \"Approach SG was -2.4 — your 150-yard shots cost the most strokes\")). Eventually: link into practice plan. |
| **Recent trend** | Mini sparkline / text: \"3rd round in the 80s in a row\" or \"SG improving 4 of last 5 rounds\" |
| **Footer** | Donation ask (Ko-fi, GH Sponsors), unsubscribe, link to oga.golf full report |

### Donation framing in footer

Per the launch strategy donation playbook (#316 has the full version):

> *\"OGA stays free because golfers chip in. If today's report helped, [chip in to keep it free for everyone →]\"*

[...]

PR #426, opened 2026-05-25
feat: email scaffold — prove the pipe (#317)

cner-smith:
## Summary
Scaffolds the email round-summary pipeline (#317, roadmap Phase 2) — "prove the pipe", no AI. A completed round → a minimal templated email (static lede + 4 SG bars) → inbox; opt-in respected; idempotent.

## What shipped
- **Migration 0029**: `rounds.completed_at`, `rounds.summary_email_sent_at`, `profiles.email_round_summaries_enabled` (default ON) + partial index; types regenerated; the column added to the profiles read allowlist.
- **`completed_at`** stamped at round completion (web + mobile).
- **`@oga/core` `buildRoundSummary`** — pure, import-free content builder (7 vitest tests, incl. a `formatToPar` parity test).
- **Inaugural Edge Function** `supabase/functions/round-summary-email` — manual-invoke (prove the pipe) + cron-scan modes, recipient via `auth.users` (service role), par from `course_tees.par`, Resend send, `List-Unsubscribe` header, idempotent `summary_email_sent_at` stamp. Pure builder vendored into `_shared/` via `pnpm email:vendor-core` (gitignored).
- **Opt-in toggle** in web Settings + mobile Profile (default ON).

## Out of scope (Phase 3/4)
Weakness picker, best/worst hole, donation-frequency cap, six-section editorial, AI commentary, per-user timezone window, one-click web unsubscribe endpoint.

## Reviewer notes
- Migration numbering continues from existing `0028`; the `0025` gap is pre-existing, not introduced here.
- "RLS-safe trigger" = service-role server-side only + owner-only email + opt-in write under the existing `profiles` UPDATE policy. No literal `CREATE TRIGGER`; the send is a (deferred) pg_cron → Edge Function HTTP post.
- Par is read from the round's `course_tees.par` (72 fallback) — matches the dashboard for standard courses; exact padded-holes parity deferred.
- ⚠️ Cron is NOT a migration and NOT scheduled in this PR. Scheduling it on prod would email every existing user (opt-in defaults ON); enable as a deliberate launch step after backfilling `summary_email_sent_at` on historical rounds. Ops steps in `docs/internal/email-setup.md` (gitignored).

## Test plan
- [x] `pnpm test` (398) + `pnpm typecheck` + mobile typecheck green
- [ ] Apply 0029 to dev+prod; deploy function; manual-invoke on own round → email lands, score + SG bars match dashboard
- [ ] Cron-mode empty-body invoke sends pending; re-run returns `sent:0` (idempotency)

Refs #317.


cner-smith:
## Review — feat: email scaffold — prove the pipe (#317)

**Concerns**

- `supabase/functions/round-summary-email/index.ts:109-113` — The `.update({ summary_email_sent_at })` result is never checked. If the Supabase write fails, `sent++` still runs and `summary_email_sent_at` stays null, so the next cron tick re-sends the same email. The PR description promises idempotency; this path quietly breaks it. Suggested fix: destructure `{ error: stampErr }` from the update, log and `continue` on failure, and only increment `sent` after a confirmed stamp.

- `supabase/functions/round-summary-email/index.ts:41` — No caller authorization is present. By default Supabase Edge Functions accept any valid user JWT, so any authenticated app user can POST to this endpoint. In cron mode (empty body) that triggers a send for every pending round across all users; in manual mode it sends to the owner of any round ID the caller supplies. A lightweight guard — checking that `Authorization` contains the service-role key, or verifying a shared secret in an `x-cron-secret` header — would close this before the function is deployed to production.

- `supabase/functions/round-summary-email/index.ts:85` — In manual mode the query has no `total_score IS NOT NULL` guard (unlike the cron path at line 58), so `round.total_score ?? 0` can reach `buildRoundSummary` with `totalScore: 0`, producing "Your round at X: 0 (−72 to par)". For a prove-the-pipe tool this is a testing hazard rather than a production bug, but it will produce a confusing first email if the test round was created without completing a score. Applying the same `.not('total_score', 'is', null)` filter to the manual branch, or guarding inside `buildRoundSummary`, would fix it.

---
Looked good: `buildRoundSummary` correctly placed in `@oga/core` with thorough test coverage (including `formatToPar` parity), migration is purely additive with no RLS surface change on new tables, no secrets in diff, domain pillars untouched, PR targets `dev`, scope held to the email scaffold concern.

---

Issue #201, opened 2026-05-06
Scorecard share: use native share sheet instead of auto-download


cner-smith:
Currently scorecard sharing auto-downloads a PNG.

Improve to use native share sheet:

**Web:** Use Web Share API (navigator.share() with files) on mobile browsers. Falls back to download on desktop where share API isn't supported.

**Mobile:** expo-sharing is already installed — use Sharing.shareAsync() to bring up native share sheet.

No server needed — image shared directly from memory/temp storage.

Ref: https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API

cner-smith:
## Implementation notes

### Mobile (expo-sharing) — the easy path
Already in deps. Standard pattern:

```ts
import * as Sharing from 'expo-sharing'
import { captureRef } from 'react-native-view-shot'
import * as FileSystem from 'expo-file-system'

async function shareScorecard(ref: React.RefObject<View>) {
  const uri = await captureRef(ref, { format: 'png', quality: 1.0 })
  if (!(await Sharing.isAvailableAsync())) {
    // Edge case: very old Android, Sharing not available
    // Fall back to saving to local files
    return
  }
  await Sharing.shareAsync(uri, {
    mimeType: 'image/png',
    dialogTitle: 'Share scorecard',
    UTI: 'public.png', // iOS
  })
}
```

The current `react-native-view-shot` + auto-download path already captures the PNG to a temp file — `expo-sharing` just opens the native share sheet on the same URI. Net change: ~5 lines.

### Web (Web Share API) — gotchas matter
```ts
async function shareScorecard(blob: Blob) {
  const file = new File([blob], 'oga-scorecard.png', { type: 'image/png' })
  // Feature detection MUST include canShare with the actual file
  if (typeof navigator.share === 'function' && navigator.canShare?.({ files: [file] })) {
    try {
      await navigator.share({ files: [file], title: 'OGA Scorecard' })
      return
    } catch (err) {
      // AbortError = user dismissed. Don't show error toast.
      if ((err as Error).name !== 'AbortError') throw err
    }
  }
  // Fallback: download path (current behavior)
  triggerDownload(blob)
}
```

### Critical Web Share API gotchas
- **User gesture required:** must be triggered DIRECTLY from a button onClick handler. Promise chains can break the user-gesture-token. The capture step (`captureRef`/`html2canvas`) takes time; do it BEFORE the user clicks if possible, or use a tiny placeholder image while capturing.
[...]

PR #427, opened 2026-05-26
feat(web): scorecard share via Web Share API, download fallback (#201)

cner-smith:
## What

Switches the web scorecard share from auto-download to the **native share sheet** (Web Share API), with a download fallback. Implements #201.

Scope is **web-only** — mobile already opened the native sheet via `expo-sharing` (`apps/mobile/app/(app)/round/[id]/index.tsx`), so that half of #201 was already done.

## How

One file changed: `apps/web/src/pages/rounds/hooks/useRoundActions.ts`.

- **Pre-capture into a ref.** A `useEffect` rasterises the off-screen scorecard card to a `Blob` (via `html-to-image` `toBlob`) and stores it in a ref, recapturing whenever the round data or the light/dark tone toggle changes. This lets `handleShare` call `navigator.share()` with **no awaited work between the click and the share call** — the Web Share API needs a live user-activation token, and an `await` in the handler expires it on iOS Safari.
- **Share, then fall back.** When `navigator.canShare?.({ files: [file] })` is true → `navigator.share({ files, title })`. `AbortError` (user dismissed the sheet) is silent. On unsupported browsers (e.g. desktop Firefox) or any non-dismissal failure → the existing download path, rebuilt from the same `Blob` via `URL.createObjectURL` (revoked after a tick). The `${courseSlug}-${date}-scorecard.png` filename is preserved.
- If the pre-captured blob isn't ready yet, a fresh `await toBlob` capture covers it (download fallback path; gesture risk only applies there).

## Verification

- ✅ `pnpm typecheck` — `navigator.canShare` / `share({ files })` compile with no assertions.
- ✅ `pnpm --filter web lint`
- ✅ `pnpm --filter web build`

## Testing notes

⚠️ **No automated coverage.** Web Share is gesture-gated and the e2e env is desktop Chromium, so this can't be driven headless. **iOS Safari is the key manual-QA target** (it's where mobile-web native share matters most and where the gesture-token handling is exercised). Desktop Chrome/Edge/Safari and Android Chrome should show the native sheet; desktop Firefox should fall back to download.

Implements #201.

### DIFF
--- a/apps/web/src/pages/rounds/hooks/useRoundActions.ts
+++ b/apps/web/src/pages/rounds/hooks/useRoundActions.ts
@@ -435,16 +435,12 @@ export function useRoundActions(input: UseRoundActionsInput): UseRoundActionsRes
     }
   }, [round.data, courseId, user, profile.data?.handicap_index, completeMutation, setCompleteError])
 
-  // Pre-rasterise the off-screen ShareableScorecardCard so the share
-  // handler can hand `navigator.share()` a ready Blob with NO awaited
-  // work in between the click and the share() call. The Web Share API
-  // needs a live user-activation token; an `await` inside the click
-  // handler expires that token on iOS Safari, so the capture has to
-  // already be done by click time. The card lives absolutely-positioned
-  // far off the left edge (the user never sees it) purely to give the
-  // rasteriser a real DOM node. 2x pixel ratio = crisp on retina without
-  // ballooning the file for group-chat shares. Recaptures whenever the
-  // round data or the light/dark tone changes so the Blob is never stale.
+  // Pre-rasterise the off-screen ShareableScorecardCard to a Blob so the
+  // share handler can hand `navigator.share()` a ready file with NO
+  // awaited work between the click and the share() call: the Web Share
+  // API needs a live user-activation token, and an `await` inside the
+  // click handler expires that token on iOS Safari. 2x pixel ratio keeps
+  // it crisp on retina without ballooning the file for group-chat shares.
   const shareBlobRef = useRef<Blob | null>(null)
   useEffect(() => {
     const node = shareCardRef.current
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — feat(web): scorecard share via Web Share API, download fallback (#201)

**Concerns**
- `apps/web/src/pages/rounds/hooks/useRoundActions.ts` (`if (!blob) return`): when both the pre-captured blob and the fresh-capture fallback fail — e.g. the scorecard card isn't in the DOM at click time — the handler sets `sharing = true`, hits this return, and the `finally` block resets the spinner. The user taps Share and nothing happens with no message. The original guard (`if (!shareCardRef.current || sharing) return`) had the same silent-return shape but fired before any state change. Consider calling `setCompleteError(toUserMessage(...))` or a generic "couldn't prepare scorecard image" message here to match how the rest of the handler surfaces failures.

**Nits**
- `apps/web/src/pages/rounds/hooks/useRoundActions.ts` (`useEffect` dep array — `[shareCardRef, round.data, shareTone]`): `shareCardRef` is a ref object; its identity is stable for the lifetime of the component, so this entry never triggers a re-run. `react-hooks/exhaustive-deps` flags ref objects in dep arrays for exactly this reason. Remove it.
- `apps/web/src/pages/rounds/hooks/useRoundActions.ts` (block comment above the `useEffect`): the tail — "Recaptures whenever the round data or the light/dark tone changes so the Blob is never stale" — narrates what the code already shows. Per CONTRIBUTING.md: "Don't narrate what the code does." The iOS Safari / gesture-token explanation in the first half is genuinely non-obvious and worth keeping; trim the rest.

---
Looked good: no `await` before `navigator.share` (gesture token preserved on iOS Safari), `AbortError` silenced and non-dismissal errors fall through to download correctly, blob URL revoked after a tick, scope held to one file and one concern, targets `dev`.

---

Issue #374, opened 2026-05-21
Shot Patterns: export 1200×630 share card (PNG)


cner-smith:
Add a small "↓ Export · 1200×630" button to the Shot Patterns page that renders a share-card image: dispersion plot + verdict headline + sample size + club label, in OGA paper-editorial palette. Sized to render cleanly when shared on Twitter / Instagram.

Marketing value: every shared image is a free recruiting poster.

## Why this is a feature, not a quick refinement

Needs an explicit rendering pipeline decision. Two paths:

- **(a) html2canvas** — new dep, has CORS issues with Mapbox satellite tiles. Shot Patterns doesn't render Mapbox (only the dispersion plot), so this may be fine. Easier implementation.
- **(b) Bespoke SVG renderer** — dispersion plot is already SVG; build a separate share-card SVG that re-uses the same primitives + adds text, then convert to PNG via the canvas API + an offscreen canvas. No new dependency, pixel-perfect typography control.

**Recommendation:** (b). Re-uses existing SVG primitives, no new dep, better typography control.

## Mobile

React Native uses `react-native-view-shot` (already in the Expo ecosystem) to render a hidden component view to a file. Different impl, same UX.

## Acceptance

- Web: clicking the export button downloads a 1200×630 PNG with the correct OGA branding
- Mobile: tapping the equivalent action triggers the native share sheet with the same PNG
- Exported image renders chart + verdict + sample size correctly with no missing fonts (Fraunces embedded if needed)
- No PII in the exported image

cner-smith:
## Implementation analysis 2026-05-22

### Current state on `dev`

- Shot Patterns dispersion plot is already SVG (no canvas, no Mapbox). Verified in `apps/web/src/pages/patterns/ShotPatternsPage.tsx`. **No CORS issues** because the page doesn't render satellite tiles.
- Scorecard share (#201) already uses `html-to-image`'s `toPng()` for raster export — there is already a rasterization dependency in `apps/web/package.json`. Worth knowing before deciding to skip it for #374.
- Mobile: `react-native-view-shot` is **not** yet a dep; the only image-share path that exists today is on `apps/mobile/app/(app)/round/[id]/index.tsx` using `expo-sharing`, and that path appears to share a pre-captured file, not a runtime-captured view.
- `apps/mobile/components/round/PuttingSheet.tsx` etc. all use `react-native-svg`, so SVG primitives transfer cleanly mobile-side.

### Two implementation paths

[...]

PR #428, opened 2026-05-26
feat(web): Shot Patterns 1200×630 share card with native share (#374)

cner-smith:
## What

Adds an **"↓ Export · 1200×630"** button to the Shot Patterns page that renders a social share card (dispersion plot + verdict headline + sample size + 68% spread + dominant miss + aim-correction, in the OGA paper-editorial palette) and opens the **native share sheet**, falling back to a download. Implements #374 (web; mobile parity is a fast follow).

The button appears only once a club has a computed pattern (≥5 shots with aim + landing).

## Rendering path — Path A (`html-to-image`), reusing #201

The ticket's Decision-log picked Path B (bespoke SVG→canvas), but that predates #201. Since #201 shipped on `html-to-image` and the acceptance criteria require **"same rasterizer as #201 — no two paths,"** this uses the same `toBlob` + Web Share API + download path. One pipeline, not two (the issue calls two rasterizers "the worst outcome").

## Structure

- New `components/DispersionPlot.tsx` — extracted `DispersionPlot` + `pointColor` so the page **and** the share card share one plot (no duplicated SVG). This also keeps `ShotPatternsPage.tsx` under the 1000-line ceiling (was heading to ~1030 inline; now 741). Added an optional `size` prop so the card renders the plot at a **fixed, viewport-independent** size (the page's responsive `min(420px, 90vw)` would otherwise make the capture depend on window width).
- New `components/ShotPatternsShareCard.tsx` — the 1200×630 layout (split: plot left, verdict right).
- `ShotPatternsPage.tsx` — Export button + off-screen render target + capture handler.

## Notes / decisions

- **Capture-on-click, not pre-captured** (unlike #201). Pre-rasterising a heavy 1200×630 card on every club-chip tap would be wasteful, and the club filter changes often. Trade-off: the `await` before `share()` can expire the iOS-Safari gesture token — in which case it degrades to the **download** path (same 1200×630 PNG), which the AC explicitly allows ("downloads **or** shares").
- **No PII** — the card renders club + dispersion stats only; no username, email, or avatar.
- `AbortError` (user dismisses the sheet) is silent; a render failure surfaces a small inline message next to the button.

## Verification

- ✅ `pnpm typecheck`, ✅ `pnpm --filter web lint`, ✅ `pnpm --filter web build`.

## Testing notes

⚠️ **The card layout has not been visually verified yet** — it's a fixed-dimension marketing asset, so it wants a real look on the Vercel preview (pick a club with a pattern, hit Export). Native share is gesture-gated and untestable headless; **iOS Safari** is the key manual check (else it downloads). Happy to capture a screenshot via the `media`-branch tooling if you want one before merging.

Implements #374.

Out of scope (per ticket): mobile parity (fast follow), 1080×1080 Instagram variant.

### DIFF
--- a/apps/web/src/pages/patterns/ShotPatternsPage.tsx
+++ b/apps/web/src/pages/patterns/ShotPatternsPage.tsx
@@ -72,8 +72,10 @@ export function ShotPatternsPage() {
     setExporting(true)
     setExportError(null)
     try {
-      // Mount the card on first export, then wait a paint for it to lay out
-      // before rasterising.
+      // The card is normally pre-mounted on the Export button's pointer-enter
+      // / focus (below), so the ref is ready here and the iOS share gesture
+      // token stays intact. Fallback: if a fast tap beat the mount, mount now
+      // and wait a paint — this rare path may forfeit the token (→ download).
       if (!shareCardRef.current) {
         setShowShareCard(true)
         await new Promise<void>((resolve) =>
@@ -200,6 +202,8 @@ export function ShotPatternsPage() {
               <button
                 type="button"
                 onClick={handleExport}
+                onPointerEnter={() => setShowShareCard(true)}
+                onFocus={() => setShowShareCard(true)}
                 disabled={exporting}
                 className="text-caddie-accent hover:bg-caddie-accent/10 disabled:opacity-40"
                 style={{
--- a/apps/web/src/pages/patterns/ShotPatternsPage.tsx
+++ b/apps/web/src/pages/patterns/ShotPatternsPage.tsx
@@ -55,6 +55,9 @@ export function ShotPatternsPage() {
   const shareCardRef = useRef<HTMLDivElement | null>(null)
   const [exporting, setExporting] = useState(false)
   const [exportError, setExportError] = useState<string | null>(null)
+  // Off-screen share card mounts lazily on first export so it isn't laid out
+  // on every club-chip re-render for users who never share.
+  const [showShareCard, setShowShareCard] = useState(false)
 
   // Rasterise the off-screen 1200×630 share card and open the native
   // share sheet (Web Share API), falling back to a download where file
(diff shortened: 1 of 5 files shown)


cner-smith:
## Review — feat(web): Shot Patterns 1200×630 share card with native share (#374)

**Concerns**
- `apps/web/src/pages/patterns/components/ShotPatternsShareCard.tsx:38` — `verdictHeadline` is a pure format helper: it takes `DispersionStats` and returns a string. CONTRIBUTING.md: "format helpers… live in `packages/core/src/`." Mobile's share card will need to duplicate this when it ships. Lift it to `@oga/core` (with a test, per the new-export rule) so both surfaces share one copy.

- `apps/web/src/pages/patterns/ShotPatternsPage.tsx:83` — The download fallback calls `link.click()` on a detached `<a>` element. Firefox and pre-15 Safari don't reliably honour downloads from detached anchors. The conventional fix is `document.body.appendChild(link); link.click(); document.body.removeChild(link)` before the `setTimeout` revoke — the PR description calls this the critical fallback path, so it's worth getting right before merging.

**Nits**
- `apps/web/src/pages/patterns/ShotPatternsPage.tsx:359` — The 1200×630 off-screen card lives permanently in the layout tree whenever `stats !== null`. The browser runs layout on it on every club-chip re-render even when the user never exports. Gating render behind a `showShareCard` boolean (flipped on first export attempt) costs nothing to the interaction model and avoids continuous hidden layout work.

---
Looked good: PR targets `dev`, no new `@oga/core` exports without tests, no secrets or RLS scope, `DispersionPlot` extraction is justified (avoids duplicating ~130 lines of SVG between page and card, and the `size` prop is the whole point), `Stat` component has three callers, one coherent feature concern throughout.

---

cner-smith:
Addressed review:

- **`verdictHeadline` → `@oga/core`** — lifted to `packages/core/src/shot-patterns.ts` as `dispersionVerdict` (alongside `getAimCorrection`) with 4 tests covering each branch. The web card now imports it; the mobile PR (#429) will consume the same copy (reconciles the duplicate — #429 merges after this).
- **Detached-anchor download** — the fallback now does `document.body.appendChild(link) → click() → removeChild(link)` before revoke, so Firefox/older Safari honour it.
- **Off-screen card always in the layout tree** — gated behind a `showShareCard` flag that flips on first export (then waits a paint before rasterising). Users who never export never mount the 1200×630 card. First export on iOS may degrade to download if the extra paint expires the gesture token — consistent with the existing `await toBlob` note; subsequent exports keep the card mounted and skip the wait.

Verified: `@oga/core` 17/17 (shot-patterns), `pnpm typecheck` clean, web lint + build clean.

**Merge order:** this PR provides `dispersionVerdict`, so merge it before #429.

cner-smith:
Follow-up on the off-screen-card gating: switched from "mount on first export click + wait a paint" to **pre-mount on the Export button's `onPointerEnter`/`onFocus`**. This keeps the perf win (non-exporters never mount the 1200×630 card) while moving the mount + layout **off** the gesture→share critical path — so on the first export the iOS share gesture token stays intact (only the existing `toBlob` await remains, same as before this PR). A lazy-mount fallback in `handleExport` still covers the rare fast-tap-before-mount case.

Rationale: estimated the click-then-mount approach had a ≥5% chance of forfeiting the first-tap share sheet on iOS (heavy card raster + mount on the critical path); pre-mounting on intent removes that at no cost.

PR #429, opened 2026-05-26
feat(mobile): Shot Patterns share card + native share (#374)

cner-smith:
## What

Mobile parity for the web Shot Patterns share card (#374, web shipped in #428). Adds an **"↓ Export · 1200×630"** button on the mobile Shot Patterns screen (shown once a club has a pattern) that captures an off-screen 1200×630 card and opens the **native share sheet**.

## How

One file: `apps/mobile/app/(app)/patterns.tsx`.

- `captureRef` (react-native-view-shot) → `Sharing.shareAsync` (expo-sharing) with an `isAvailableAsync` fallback alert — the **same path as the scorecard share** (`round/[id]/index.tsx`). **No new deps** (both already pinned: `react-native-view-shot@3.8.0`, `expo-sharing@~12.0.1`).
- Co-located `ShotPatternsShareCard` reuses the in-file RN `DispersionPlot` (new optional `size` prop for a fixed, viewport-independent render). Mirrors the web card's **verified** layout (350px plot, same paddings) using the registered mobile fonts (`Fraunces-Medium` / `-MediumItalic`; kickers use the screen's system-font kicker style).
- Off-screen target uses the scorecard's pattern: `position: absolute; left: -10000` + `collapsable={false}`.
- Export button uses an `opacity` press-feedback in the `style` callback (the iOS `android_ripple` cross-platform rule). **No PII** — club + dispersion stats only.

File stays at 758 lines (under the 1000-line ceiling), so the card is co-located rather than extracted (unlike web, where the line budget forced a split).

## Verification

- ✅ `npm run typecheck` (mobile). No mobile lint script exists; typecheck is the CI gate.

## ⚠️ Testing notes — needs on-device QA (could not verify here, no emulator)

1. **react-native-svg inside a view-shot capture** — this is the app's **first** such case (the scorecard card is text/views, no SVG). view-shot has historically rendered SVG **blank** on some Android configs. **Verify the dispersion plot actually appears** in the shared image, not a blank panel.
2. **1200-wide off-screen layout** — confirm the card isn't clipped to screen width when captured.
3. **Output dimensions** — capture is at natural device-scaled size (1200×630 logical × density), matching the scorecard path; aspect is exactly 1200:630.
4. Fonts render (Fraunces medium/italic) and the footer isn't clipped (it was on web before the fix — same layout math here, but RN flexbox can differ).

Implements #374 (mobile half). Pairs with #428 (web).

### DIFF
--- a/apps/mobile/app/(app)/patterns.tsx
+++ b/apps/mobile/app/(app)/patterns.tsx
@@ -1,6 +1,8 @@
-import { useEffect, useMemo, useState } from 'react'
-import { Pressable, ScrollView, Text, useWindowDimensions, View } from 'react-native'
+import { useEffect, useMemo, useRef, useState } from 'react'
+import { Alert, Pressable, ScrollView, Text, useWindowDimensions, View } from 'react-native'
 import Svg, { Circle, Ellipse, Line, Rect, Text as SvgText } from 'react-native-svg'
+import { captureRef } from 'react-native-view-shot'
+import * as Sharing from 'expo-sharing'
 import {
   CLUBS,
   LIE_SLOPES,
@@ -12,6 +14,7 @@ import {
   type Club,
   type DispersionPoint,
   type DispersionStats,
+  type DistanceUnit,
   type LieSlope,
   type LieType,
   type Shot,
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — feat(mobile): Shot Patterns share card + native share (#374)

**Concerns**
- `apps/mobile/app/(app)/patterns.tsx:320` — `shareVerdict` is a pure format helper: it maps `DispersionStats` → a human-readable verdict string with no UI dependency. CONTRIBUTING.md is explicit: "format helpers … live in `packages/core/src/`." The paired web PR (#428) almost certainly has an equivalent function; if it lands separately, the same conversion will exist in two app layers with no shared source of truth. Moving `shareVerdict` to `packages/core/src/shot-patterns.ts` (alongside `getAimCorrection`, which already lives there with tests) would also require a test, but the logic is straightforward to cover.

**Nits**
- `apps/mobile/app/(app)/patterns.tsx:143` — `// eslint-disable-next-line no-console` added without an inline explanation. CONTRIBUTING.md: "No `// eslint-disable-next-line` without a comment explaining why." A brief note on the same line (e.g., `// intentional — no toast primitive on mobile, console is the only fallback`) would satisfy this.

---
Looked good: no new dependencies, no migrations, no RLS gaps, no secrets, PR targets `dev`, `nativewind` untouched, dep ranges not widened, `getAimCorrection` sourced from `@oga/core` with existing tests, `isAvailableAsync` fallback handled, off-screen capture pattern matches the existing scorecard share, scope held to one concern.

---

cner-smith:
Addressed review:

- **`shareVerdict` → `@oga/core`** — now consumes the shared `dispersionVerdict` from `packages/core/src/shot-patterns.ts` (the same helper #428 adds); the local copy is gone, so web and mobile share one source of truth.
- **`eslint-disable` without explanation** — annotated both `no-console` disables (the flagged share-failure one and the shot-fetch one) with `-- <reason>`; ESLint clean.

Verified: mobile `tsc --noEmit` clean, `eslint app/(app)/patterns.tsx` clean, `@oga/core` shot-patterns 17/17.

**Merge order / dependency:** this PR needs `dispersionVerdict`, which lands in **#428**. To keep CI green now, this branch temporarily includes the core commit (`feat(core): dispersionVerdict …`) via cherry-pick. **Merge #428 first;** I'll then rebase this branch onto `dev` and that shared commit drops out by patch-id, leaving only the mobile delta.

Issue #18, opened 2026-05-01
AI practice plan generation (Phase 5)


cner-smith:
Supabase Edge Function calling claude-sonnet-4-6 to generate personalized practice plans based on strokes gained data and skill level. Requires ANTHROPIC_API_KEY. Currently the practice plan page is a placeholder — this is the backend that powers it.

cner-smith:
## Implementation analysis 2026-05-22

### Current state on `dev`

- `apps/web/src/pages/practice/PracticePlanPage.tsx` is a static placeholder card ("Coming in phase 5"). No data hookup, no fetch.
- `apps/mobile/app/(app)/practice.tsx` mirrors this.
- `supabase/functions/` directory does not exist — **this would be the project's inaugural Edge Function** (per #317, which has the same prerequisite — bundle the Edge Functions infra work across both tickets).
- `@oga/core` already has the SG breakdown math (`computeRoundSG` in `packages/core/src/sg.ts`) and drill matchers + skill brackets (`@oga/core` exports `FACILITIES`, `GOALS`, `SKILL_LEVELS`).
- Drills live in `supabase/seed.sql` — 24 drills tagged with `category`, `facility`, `skill_levels[]`.

### Two implementation paths

#### Path A — On-demand generation (Edge Function called from the page)

The page mounts → calls `supabase.functions.invoke('practice-plan', { body: { userId, lookback: 5 } })` → Edge Function loads the last N rounds' SG aggregates, builds an Anthropic prompt (Sonnet 4.6 per `claude-api` skill defaults), returns markdown + a drill-id list → client renders.

**Pros**
- Plans are always current with the latest rounds (no staleness).
- Simplest pipeline — no scheduled job, no caching layer.
- Cheap initial implementation: Edge Function + Anthropic call + thin client renderer.

**Cons**
- Latency hit on every page open (~3-6s Anthropic call). Visible spinner.
- Cost scales with page opens, not rounds played. A bored user reloading the page is paying for tokens.
- No persistence — if Anthropic is down, the page is broken.

**Files**
- New: `supabase/functions/practice-plan/index.ts`
- New: `packages/core/src/practice-plan.ts` (prompt builder + drill matcher — pure, testable)
- Update: `apps/web/src/pages/practice/PracticePlanPage.tsx`, `apps/mobile/app/(app)/practice.tsx` (fetch + render)
- Update: `.env.example` to document `ANTHROPIC_API_KEY` as a Supabase Edge Function secret

[...]

PR #430, opened 2026-05-26
feat: per-round strokes-gained nudge in round summary (#18)

cner-smith:
## Summary

Adds a deterministic (no-AI) per-round **"Today's focus"** nudge. After a round, the web round summary surfaces the player's worst strokes-gained category plus up to two facility-appropriate drills from the existing library.

- **`@oga/core` — new `round-nudge.ts`** (pure, 9 unit tests):
  - `pickRoundFocus` — worst SG category for the round, gated on a clear leak (`-0.5` threshold); tie-break order putting > approach > around_green > off_tee.
  - `roundFocusHeadline` — coach-voice line (`"Approach cost you about 2.1 strokes this round."`).
  - `selectNudgeDrills` — keeps drills doable at the player's facilities (no-facility = anywhere), caps at 2.
- **`RoundSummary.tsx`** — co-located `RoundNudge` component rendering the focus block (headline + drill chips) immediately before the existing "Generate practice plan" CTA, via `useProfile` + `useDrills`. Shows from round 1; absent when no category is below the threshold.

No schema change, no AI call. This is **track 0** of the practice-plan feature (#18) — it proves the drill-retrieval gate against current data before any engine work. (Mobile parity is a separate follow-up; same `@oga/core` functions, different render surface.)

## Test plan

- `pnpm --filter @oga/core test` — **407 pass** (9 new).
- `pnpm typecheck` — clean (3/3). `pnpm --filter web lint` + `build` — clean.
- Verified rendering locally against the real component with mock profile/drills:
  - Leak round (`sg_approach = -2.1`) → headline + 2 chips (net-only drill correctly excluded; capped at 2), rendered above the CTA.
  - No-leak round (all categories > `-0.5`) → block absent.
- [ ] **QA on the Vercel preview, signed-in:** open a past round with an SG category < -0.5 → confirm the "Today's focus" block + chips appear (chips limited to drills doable at your facilities); confirm the block is **absent** on a balanced round; confirm no regression to score / SG cells / best-worst / the >=3-rounds CTA.

Refs #18 (track 0 — does not close the broader feature).


cner-smith:
## Review — feat: per-round strokes-gained nudge in round summary (#18)

**Concerns**
- `apps/web/src/components/rounds/RoundSummary.tsx:175` — `useDrills` fires immediately with `skillLevel: undefined` while the profile is still loading, because `useDrills` has no `enabled` guard. React Query will re-fetch once the profile resolves (the `queryKey` includes `skillLevel`), but during that window `picks` is drawn from an unfiltered-by-skill-level result set. The coincidental `facilities: []` guard (`profile?.facilities ?? []` also undefined while loading) limits exposure to "anywhere" drills only, but the pattern is fragile — it silently returns partially-correct data rather than waiting. The straightforward fix is to pass `enabled: !!profile` to the `useQuery` inside `useDrills`, or to skip rendering `RoundNudge` until `profile` has settled.

- `apps/web/src/components/rounds/RoundSummary.tsx:211` — every drill chip links to `/practice/drills` regardless of `drill.id`. If a drill detail route exists (or is planned), the chips should deep-link. If the generic list is genuinely the right destination, it would help to note that intent so the next reader doesn't assume it's an oversight.

**Nits**
- `apps/web/src/components/rounds/RoundSummary.tsx:146` — IIFE in JSX to scope a single variable. `const focus = pickRoundFocus(round)` hoisted to the `RoundSummary` function body is idiomatic React and reads more naturally; the IIFE adds visual noise without benefit here.

- `packages/core/src/round-nudge.ts:73` — `void _categoryCheck` is an unusual lint-suppression idiom. A type-level assertion (`type _Check = NudgeCategory extends ShotCategory ? true : never`) expresses the same invariant without a runtime object, or the check could live as a `satisfies` expression on the `NUDGE_COLUMNS` constant (which already has one) and skip the separate variable entirely.

- `packages/core/src/round-nudge.test.ts` — no test for the exact threshold boundary (`sg_approach: -0.5`). The current tests bracket the boundary (-0.3 → null, -0.6 → focus) but don't pin which side `-0.5` falls on. The code is `> NUDGE_THRESHOLD`, so `-0.5` triggers; a boundary test would make that intent explicit and catch if the operator is ever changed to `>=`.

---
Looked good: core logic correctly placed in `@oga/core` (not `apps/web/src/lib`), all three exports have matching tests, `selectNudgeDrills` correctly treats `facility: null` as "anywhere", tie-break logic verified by test, PR targets `dev`, single concern, no `any`, no bare `eslint-disable`.


---

cner-smith:
Addressed review:

- **`useDrills` `enabled` guard** — added an optional `options.enabled` to `useDrills`; `RoundNudge` now passes `enabled: !!profile` so the drills query waits for the profile to settle (no more transient fetch with `skillLevel: undefined`). Headline still renders immediately; only the chips wait.
- **IIFE → hoisted `const focus`** — `pickRoundFocus(round)` hoisted to the component body; render is now `{focus && <RoundNudge focus={focus} />}`.
- **`void _categoryCheck`** → type-level `type _NudgeIsShotCategory = NudgeCategory extends ShotCategory ? true : never` (no runtime object).
- **Threshold boundary test** — added a test pinning `sg_approach: -0.5` → fires (gate is `>` not `>=`).
- **`/practice/drills` link** — kept (no `/practice/drills/:id` route exists; only the list). Added an intent comment so it doesn't read as an oversight.

Verified: `@oga/core` 10/10 tests, `pnpm typecheck` clean, web lint + build clean.

PR #431, opened 2026-05-26
feat(db): drill corpus schema — 0030 (#18)

cner-smith:
## Summary

Schema for the AI practice plan's drill corpus (#18) — **PR 1 of 2** (the sourced drill seed is a separate follow-up).

**`drills` enrichment** so retrieval can gate precisely and contributions/provenance are tracked:
- `goals text[]` (goal bands a drill suits), `targets text[]` (weakness tags), each GIN-indexed
- `drill_type text` (checked: warmup/technical/skill_game/pressure_game/putting)
- `target_template jsonb` (skill-scaling for deterministic targets — resolved in `@oga/core`, engine track; left null for now)
- `source` + `source_url` (instructional origin — PGA pro / book / video / study; `source_url` nullable for books/seminars)
- `contributor` (who added the row to OGA — GH handle) + `verified boolean default false` (maintainer-vetted; the retrieval gate will require `verified = true`, so unvetted community PRs never auto-reach players)

**`practice_plans`** columns the generator reads/writes: `feedback` (one untrusted note per plan, next-gen loop), `coach_note` (editorial), `raw_model_output jsonb` (spot-check/regression).

Additive `add column if not exists` (idempotent, per 0029). RLS unaffected — `drills` stays publicly readable; `practice_plans` keeps its user-owned policies.

## Out of scope (follow-up)
- The sourced drill corpus itself (replacing the 24 AI-generated drills in `seed.sql`) — separate PR after a research deep-dive.
- `resolveTarget` / `target_template` authoring and the `verified` retrieval filter — engine track.

## Test plan
- `supabase migration up --local` applies 0030 cleanly; `gen types --local` regenerates `types.ts` (new columns present, diff purely additive).
- `pnpm typecheck` — clean (3/3).
- No app code consumes the new columns yet, so no runtime surface to exercise.

Refs #18.

### DIFF
--- a/packages/supabase/src/types.ts
+++ b/packages/supabase/src/types.ts
@@ -39,6 +39,24 @@ export type Database = {
   }
   public: {
     Tables: {
+      course_submissions: {
+        Row: {
+          created_at: string
+          id: string
+          user_id: string
+        }
+        Insert: {
+          created_at?: string
+          id?: string
+          user_id: string
+        }
+        Update: {
+          created_at?: string
+          id?: string
+          user_id?: string
+        }
+        Relationships: []
+      }
       course_tees: {
         Row: {
           course_id: string
@@ -157,36 +175,60 @@ export type Database = {
       drills: {
         Row: {
           category: string | null
+          contributor: string | null
           created_at: string
           description: string | null
+          drill_type: string
           duration_min: number | null
           facility: string[] | null
+          goals: string[]
           id: string
           instructions: string | null
           name: string
           skill_levels: string[] | null
+          source: string | null
+          source_url: string | null
+          target_template: Json | null
+          targets: string[]
+          verified: boolean
         }
         Insert: {
           category?: string | null
+          contributor?: string | null
           created_at?: string
           description?: string | null
+          drill_type?: string
           duration_min?: number | null
           facility?: string[] | null
+          goals?: string[]
           id?: string
           instructions?: string | null
           name: string
           skill_levels?: string[] | null
+          source?: string | null
+          source_url?: string | null
+          target_template?: Json | null
+          targets?: string[]
+          verified?: boolean
         }
         Update: {
           category?: string | null
+          contributor?: string | null
           created_at?: string
           description?: string | null
+          drill_type?: string
           duration_min?: number | null
           facility?: string[] | null
+          goals?: string[]
           id?: string
           instructions?: string | null
           name?: string
           skill_levels?: string[] | null
+          source?: string | null
+          source_url?: string | null
+          target_template?: Json | null
+          targets?: string[]
+          verified?: boolean
         }
         Relationships: []
       }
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — feat(db): drill corpus schema — 0030 (#18)

**Concerns**
- `supabase/migrations/0030_drill_corpus.sql:24` — `feedback` is explicitly labelled "untrusted" in the comment, and the PR description states it feeds the next-gen loop. Two issues follow: (1) No DB-level size constraint enforces the `<=500 chars` note — a direct Supabase client call bypasses app-layer validation. A `CHECK (char_length(feedback) <= 500)` costs nothing and matches the stated intent. (2) Once this column is interpolated into a model prompt, it's a prompt-injection surface. The schema alone doesn't cause the problem, but it establishes the channel; recording the sanitization requirement in a comment here or a linked issue now — before the consuming code is written — prevents it from being an afterthought.

**Nits**
- `supabase/migrations/0030_drill_corpus.sql:16` — `source_url text` accepts any string. When this eventually lands in a `<Link>` or a `fetch()` call, a `javascript:` prefix or a non-URL value causes problems. `CHECK (source_url IS NULL OR source_url ~ '^https?://')` is one line and closes the gap at the boundary where data enters.

- `packages/supabase/src/types.ts:39–55, 695–704` — `course_submissions` and `insert_synthetic_hole` appear in the regenerated types diff but aren't created in this migration and aren't mentioned in the PR description. Both are pre-existing objects that `gen types --local` picked up. A one-line note in the description ("types regeneration also surfaced `course_submissions` and `insert_synthetic_hole` from prior migrations") would make `git blame` on this file less puzzling for future readers.

---
Looked good: `course_submissions` RLS is properly established in migration 0027 (SELECT policy + `security definer` trigger-only inserts); `drills` public-read and `practice_plans` user-owned RLS are unaffected by additive column changes; `drill_type` CHECK constraint covers the full value set; GIN indexes on `goals`/`targets` are the right choice for array-contains queries; scope held to one concern and PR targets `dev`.

---

cner-smith:
Addressed review:

- **`feedback` length** — added `check (char_length(feedback) <= 500)` so a direct Supabase client write can't bypass the app-layer cap (the column is user-writable under practice_plans RLS). Also expanded the comment to record the **prompt-injection / sanitization requirement** (delimit + treat as untrusted, preference-signal-only — spec D14) so it's not an afterthought when the consuming code lands.
- **`source_url`** — added `check (source_url is null or source_url ~ '^https?://')` to block `javascript:`/non-URL values at the boundary (matters once community-contributed drills render the link). Verified against the local DB: `https://…` passes, `javascript:alert(1)` is rejected.

Note on the `types.ts` diff (your suggestion): the regeneration also surfaced **`course_submissions`** (migration 0027) and **`insert_synthetic_hole`** (migration 0026) — pre-existing objects that the committed `types.ts` was missing, i.e. types on `dev` were stale and hadn't been regenerated after 0026/0027. Kept them (types now match the DB); not removing correct entries to tidy the diff.

CHECK constraints don't alter the generated types, so `types.ts` is unchanged by this fix. The new constraints apply on the next `db reset`/deploy (my local DB already had the columns from the first apply).

PR #432, opened 2026-05-27
feat(db): seed drill corpus v1 — 41 sourced drills (#18)

### LINKED ISSUE #18: AI practice plan generation (Phase 5)

cner-smith:
## Drill corpus v1 — seed (#18, Track 1)

Replaces the 24 placeholder drills in `supabase/seed.sql` with **41 sourced, fact-checked drills**, populating the `0030` enrichment columns.

### Drills — 41 total
**off_tee 10 · approach 10 · around_green 10 · putting 11**

Each row: `verified=true`, `contributor='oga'`, plus `drill_type`, `goals[]`, `targets[]`, `skill_levels[]`, `facility[]` (empty = anywhere), `source`, `source_url` (http(s)-only per the 0030 CHECK; NULL for the one book-only drill).

Every numeric/attribution claim was verified against live sources before seeding. Corrections made during review:
- Two mis-attributed off_tee "step" drills (wrong footwork; one mislabeled a "pump") merged into one accurate **Step-Through Speed** (Schauffele step-in + Harrington gather-step).
- AimPoint "1% ≈ 1 ball-width" corrected to AimPoint's real finger-count rule (the old claim contradicted its own cited source).
- Pelz "circle of love" → his actual "good zone" term + 17″-past / 34″ geometry.
- Approach near-duplicates merged (wedge carry-ladder ×2 → 1; towel low-point ×2 → 1).

### Bundled pre-existing seed fix (so `db reset` works again)
The demo-courses block had been failing `supabase db reset` independent of the drills:
- `0013` dropped `courses.location` but the seed still inserted it → now uses `city`/`state`.
- `0027`'s `courses_rate_limit_before_insert` trigger raises with no auth context → the seed (run as table owner) now disables/re-enables it around the three fixed demo courses. Transactional; trigger confirmed re-enabled (`ENABLED`) after.

### Verification
`supabase db reset` end-to-end: 30 migrations + seed apply clean → **41 drills, 3 courses, 54 holes**. All CHECK constraints (`drill_type`, `source_url ^https?://`) pass; rate-limit trigger restored.

Part of #18 (AI practice plan — corpus track). Does not close #18.

### Note on scope (CONTRIBUTING.md §5)
This PR intentionally bundles the 41-drill feature with a pre-existing courses-seed fix (`location`→`city`/`state` after 0013; disable/re-enable the 0027 rate-limit trigger around the demo inserts), deviating from one-concern-per-PR. The courses fix is the prerequisite for `supabase db reset` to verify the seed loads end-to-end, so it's grouped here deliberately.


cner-smith:
## Review — feat(db): seed drill corpus v1 — 41 sourced drills (#18)

**Concerns**
- `supabase/seed.sql:424` — The `ALTER TABLE ... DISABLE TRIGGER` / `ENABLE TRIGGER` pair is bare top-level DDL with no exception handler between them. The PR description calls this "transactional," but the diff shows no explicit `BEGIN`/`COMMIT` or `DO $$...$$` block wrapping the three statements. If the INSERT at lines 426–429 raises (e.g., a constraint violation or a future schema mismatch), the `ENABLE TRIGGER` at line 438 never executes and `courses_rate_limit_before_insert` stays disabled for the rest of the session. Whether Supabase wraps the seed file in an implicit transaction is a tooling detail, not a guarantee in the diff. A `DO $$ BEGIN … EXCEPTION WHEN OTHERS THEN ALTER TABLE public.courses ENABLE TRIGGER …; RAISE; END $$;` block would make this safe regardless.

- `supabase/seed.sql` (file-level) — This PR carries two distinct concerns: a new feature (41-drill corpus) and a pre-existing bug fix (courses `location` → `city`/`state` column rename + the trigger bypass). CONTRIBUTING.md says ["One concern per PR. UI restyle, bug fix, and new feature do not belong in the same PR."](CONTRIBUTING.md#opening-a-pr) The courses fix is presented as a prerequisite for `db reset` to work, which is a reasonable pragmatic argument, but the rule doesn't carve out that exception. If this is intentional, a note in the PR body acknowledging the deviation would help future reviewers understand the decision.

**Nits**
- `supabase/seed.sql:106,117` — "Fairway Target Game" and "Driver Dispersion Mapping" carry the identical `source_url` (`'https://practical-golf.com/driver-dispersion'`). May be intentional if both drills appear in that one article, but worth confirming there isn't a more specific link for the second entry.

---
Looked good: single-file change; PR targets `dev`; no new tables, so no missing RLS; no secrets or service-role keys in the diff; `verified=true` and `contributor='oga'` consistent across all 41 rows; the one NULL `source_url` (Hinge & Hold, line 238) is documented and valid per the `0030` CHECK constraint.

---

cner-smith:
Addressing the review:

**① Trigger safety** — the `disable`/`enable trigger` statements are *inside* the `do $$ … begin … end $$` block (`seed.sql`: `begin` L305 → `disable trigger` L310 → `insert courses` L312 → `enable trigger` L350 → `end $$`), not bare top-level DDL — the cited L424/L438 don't exist. A DO block runs atomically as a single statement, so if the INSERT raises, the whole block rolls back *including* the trigger disable; it can't be stranded off. The `EXCEPTION` wrapper would be redundant — no change.

**② One concern per PR (CONTRIBUTING.md §5)** — acknowledged, intentional; added a scope note to the PR body. The courses-seed fix (`location`→`city`/`state` after 0013; 0027 trigger bypass) is bundled because it's the prerequisite for `supabase db reset` to verify the 41-drill seed loads end-to-end.

**③ Shared `source_url`** — intentional. Both "Fairway Target Game" and "Driver Dispersion Mapping" derive from Jon Sherman's `practical-golf.com/driver-dispersion`; sharing the source is accurate.

PR #433, opened 2026-05-27
feat(mobile): round-summary nudge — parity with web (#430 follow-up)

cner-smith:
## Mobile round-nudge parity (Track 0 follow-up)

Mirrors the web `RoundSummary` "Today's focus" nudge (merged in #430) on the mobile read-only round screen (`apps/mobile/app/(app)/round/[id]/index.tsx`).

### Behaviour (identical to web)
- `pickRoundFocus` picks the worst SG category if it's below −0.5; otherwise nothing renders.
- Headline via `roundFocusHeadline` (e.g. "Putting cost you about 0.8 strokes this round.").
- Up to 2 facility-filtered drills via `selectNudgeDrills` + the shared `getDrills` query.
- All logic reused from `@oga/core` + `@oga/supabase` — **no new business logic, no `@oga/core` change** (its round-nudge tests already cover it).

### Mobile-specific choices
- **Non-interactive pills** — mobile has no drill-library route yet (practice tab is a teaser), so a tap has nowhere honest to land. Revisit when the mobile drill library exists.
- **No "Generate practice plan" CTA** — targets the unbuilt engine; deferred.
- **Lazy fetch** — no react-query on mobile; a gated `useEffect` fetches profile + drills only when a leak exists, mirroring web's `useProfile`→`useDrills` gating.

### Verification
`npm run typecheck` (mobile) green. **⚠️ Device QA pending** — RN render, Android-first, no mobile e2e; not device-verified yet.

Follow-up to #430 (web nudge). Part of #18, Track 0.

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -128,11 +128,9 @@ export default function RoundIndex() {
     }
   }, [id])
 
-  // "Today's focus" nudge (parity with web RoundSummary). Only fetch the
-  // profile + a couple of drills when the round actually has a leak —
-  // pickRoundFocus returns null on a balanced round, so nothing queries.
-  // No react-query on mobile; this mirrors web's useProfile→useDrills
-  // gating by hand.
+  // "Today's focus" nudge (parity with web RoundSummary). No react-query on
+  // mobile — gate the profile + drills fetch by hand, like web's
+  // useProfile→useDrills.
   useEffect(() => {
     if (!round || round.total_score == null || !user) return
     const focus = pickRoundFocus(round)
@@ -677,7 +675,7 @@ function RoundNudge({ focus, picks }: { focus: RoundFocus; picks: DrillRow[] }) 
         marginBottom: 18,
       }}
     >
-      <Text style={{ ...KICKER, marginBottom: 8 }}>Today&apos;s focus</Text>
+      <Text style={{ ...KICKER, marginBottom: 8 }}>Today's focus</Text>
       <Text
         style={{
           color: '#1C211C',
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -11,18 +11,27 @@ import { captureRef } from 'react-native-view-shot'
 import * as Sharing from 'expo-sharing'
 import { useFocusEffect, useLocalSearchParams, useRouter } from 'expo-router'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
-import { formatSG } from '@oga/core'
+import {
+  formatSG,
+  pickRoundFocus,
+  roundFocusHeadline,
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — feat(mobile): round-summary nudge — parity with web (#430 follow-up)

**Concerns**
- `apps/mobile/app/(app)/round/[id]/index.tsx:128-132` — The five-line comment block narrates what the code already shows ("only fetch the profile + a couple of drills when the round actually has a leak", "pickRoundFocus returns null on a balanced round, so nothing queries"). CONTRIBUTING.md is explicit: "Don't narrate what the code does." The one non-obvious fact worth keeping is that there's no react-query on mobile and the gating is manual. Everything else is code re-statement. Trim to a single line, e.g. `// No react-query on mobile; gate profile + drills fetch manually like web's useProfile→useDrills.`

**Nits**
- `apps/mobile/app/(app)/round/[id]/index.tsx:678` — `Today&apos;s focus`: in a JSX text node the apostrophe doesn't need HTML-entity encoding — `'` is valid directly. `&apos;` works because Metro/Babel decodes it, but it reads like an HTML artifact. Write `Today's focus`.
- `apps/mobile/app/(app)/round/[id]/index.tsx:133` vs `302` — The effect gates on `round.total_score == null` (plus `user`) before deciding to fetch, but the render-side `focus = pickRoundFocus(round)` and the JSX conditional at line 509 don't carry that gate — they render the headline the moment `pickRoundFocus` returns non-null, regardless of `total_score`. In practice this is probably moot because the redirect-to-live path handles in-progress rounds before this code is reached, but the asymmetry between the two guards is worth a brief inline note if it's intentional, or aligning the render gate to match the fetch gate if it isn't.

---
Looked good: package boundary clean (all logic from `@oga/core` / `@oga/supabase`, nothing added to `apps/mobile/lib/`), no new `@oga/core` exports so the test requirement doesn't trigger, `active` flag teardown in the async IIFE is correct, `user?.id` as a stable primitive in the dep array, single concern, targets `dev`.

---

cner-smith:
Addressing the review (pushed in 0ec7e52):

**① Comment over-narration** — fixed. Trimmed to the non-obvious bit (no react-query on mobile → manual gating, like web's `useProfile`→`useDrills`); dropped the lines restating the `if (!focus) return` logic. (CONTRIBUTING.md §"Don't narrate what the code does".)

**② `Today&apos;s focus`** — fixed → `Today's focus`. Confirmed mobile has no `react/no-unescaped-entities` rule active, so the raw apostrophe is safe; `&apos;` was a reflex against a lint rule that isn't even in play here.

**③ Guard asymmetry (effect gates `total_score == null`, render doesn't)** — left as-is; it's correct. An in-progress round sets `redirectToLive` and returns `<LiveRoundSession>` (line ~218) *before* the summary render, so `round.total_score` is guaranteed non-null where `focus` is computed — the render-side gate would be dead code. The effect carries the guard only because it runs *before* the early returns and can transiently see an in-progress round. Intentional asymmetry.

PR #434, opened 2026-05-27
feat(core): AI practice-plan engine helpers (Phase A, #18)

cner-smith:
## Phase A — pure `@oga/core` helpers for the practice-plan engine

Adds the deterministic logic the practice-plan generator (#18) builds on. **Pure logic only** — no Edge Function, no migration, no UI in this PR. New `packages/core/src/practice-plan/` module:

- **`playFrequencyPlan`** — maps `play_frequency` → sessions/week + the 7-day plan window.
- **`resolveTarget`** — resolves a drill's target template + the player's stats into a concrete, clamped target number (targets are computed deterministically, never model-emitted).
- **`buildPlayerDigest`** — assembles the player digest from stored per-round SG + dispersion: per-category averages (null-aware), worst-first ranked weaknesses with a fixed tie-break, per-category trend, and a self-consistency check.
- **`validatePlanDraft`** — validates a generated plan against the candidate set: drill-reference resolution, block-type consistency, per-session minute bounds, session count, top-2 weakness coverage, anti-repetition vs the prior plan, note-length — and rejects malformed (non-positive / negative / non-finite) minute values.
- **`selectBaselinePlan`** — selects a static low-data baseline plan by skill × goal with weekly rotation and worst-category preference (seed content; full variants to follow).

### Notes
- Authored dependency-free (type-only imports) so the helpers can be reused server-side later without a bundler.
- New exports surfaced via `@oga/core`.

### Tests
TDD throughout — **438/438 `@oga/core` tests green** (40 new across the 5 helpers), workspace typecheck clean.

Part of #18. Pure-logic foundation; the generation service + UI land in follow-up PRs.


cner-smith:
## Review — feat(core): AI practice-plan engine helpers (Phase A, #18)

**Concerns**
- `packages/core/src/practice-plan/targets.ts:14` — `digest.dispersion[0]?.cone68.lateral` takes the first dispersion entry without filtering by club. `digest.ts`'s `consistencyWarnings` searches for `d.club === 'driver'`, and the `digest.test.ts` fixtures pass `club: 'driver'` at index 0 — so in production the first entry is likely driver data. When `scales_with === 'approach_dispersion'`, this would silently scale approach targets using driver lateral spread. The test fixture sidesteps the issue by inventing `club: 'approach'` (not a real club name used elsewhere), which means the test passes but does not reflect what `buildPlayerDigest` actually puts in the array. Either document a binding contract that an aggregated approach entry always sits at index 0, or filter explicitly (e.g. `dispersion.find(d => d.club === 'approach')`), or define a named constant for the expected key — whatever the intended semantics are.

**Nits**
- `packages/core/src/practice-plan/validation.test.ts` (five locations, e.g. the "out-of-range drill_ref" and "negative block minutes" tests) — each `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion` has no accompanying comment explaining why the assertion is safe. CONTRIBUTING.md: "No `// eslint-disable-next-line` without a comment explaining why." In all five cases the reason is the same and brief: `structuredClone(okDraft)` preserves the full shape of `okDraft`, so the indexed fields are guaranteed non-null; a short inline note would satisfy the rule.
- `packages/core/src/practice-plan/validation.ts:40` — `b.minutes < 0` accepts `b.minutes === 0`, whereas the session-level guard at line 46 uses `s.total_minutes <= 0`. The PR description says the validator "rejects malformed (non-positive / negative / non-finite) minute values", which reads as non-positive for both levels. If zero-minute blocks are intentional (e.g. for placeholder blocks planned in Phase B), a brief comment here would clarify the discrepancy rather than leaving it looking like an off-by-one with the session check.

---
Looked good: all new logic lands in `packages/core/src/` (not in app `lib/` dirs), every exported symbol has a corresponding test, PR targets `dev`, no migrations or UI changes, no secrets or domain-pillar violations.

---

cner-smith:
Addressing the review (pushed in cf93c2e):

**Concern — `resolveTarget` dispersion scaling** — fixed by **deferring the dispersion scaling to Phase B.** You're right that there's no clean per-category "approach dispersion" in the digest (it's per-club), so scaling approach targets off `dispersion[0]` (likely driver) was wrong, and the test masked it with an invented `club: 'approach'`. v1 `resolveTarget` is now **skill baseline → clamp only**; `target_template.scales_with` stays on drills as data and gets consumed in Phase B once the digest exposes a real per-category dispersion aggregate. Tests reworked to baseline/fallback/clamp (no invented club).

**Nit — block minutes `< 0` vs session `<= 0`** — aligned: block minutes now reject `<= 0` (a zero-minute block is malformed too), message updated to "> 0".

**Nit — `eslint-disable` without reason** — annotated all of them with `-- structuredClone(okDraft) preserves the full shape, so this indexed access is non-null`.

Full `@oga/core` suite green (438), typecheck clean.

PR #435, opened 2026-05-27
feat: AI practice plan engine — Phase B (generate-practice-plan Edge Function)

cner-smith:
Phase B of the AI practice-plan engine (#18, Phase 3): the `generate-practice-plan` Supabase Edge Function — on-demand, RLS-scoped, turning a player's data into a stored weekly plan via deterministic retrieval + a Claude tool-use call, with hard cost ceilings and a baseline fallback. **No UI button is wired** (that's Track 3 / Practice UI).

Built subagent-driven (implementer → spec review → code-quality review per task) from `ai-practice-plan-engine-phase-b-plan.md`, which was first **hardened by a 3-agent review panel** (LLM-integration / prompt-engineering / security) that folded in 6 blockers before any code.

## What's in here (13 commits)
- **`@oga/core` pure helpers** (vitest, then vendored import-free into `_shared/practice-plan/`): `sanitizeFeedback` (untrusted-input scrub: C0+DEL+C1 strip, angle-bracket forge-proofing), `buildPlanPrompt` + `PLAN_TOOL` (0-based index refs, enum/`additionalProperties:false`/`strict:true` schema, randomized-id feedback delimiter), `resolvePlanForStorage`, and an `article_ref` bounds-check added to `validatePlanDraft`.
- **Edge modules:** `retrieval.ts` (§7 candidate-pool query with deterministic `id asc` tiebreak + warmup/putting guarantee; published-article snapshot), `claude.ts` (Anthropic tool-use call: `maxRetries:1`, forced tool, temp gated, `max_tokens:4096`, `stop_reason` truncation handling, `tool_result` repair contract, typed error taxonomy), `index.ts` (the thin 8-step orchestrator).
- **Migrations:** `0031` cadence unique index (TOCTOU concurrency backstop, immutable `date_trunc(... AT TIME ZONE 'UTC')`); `0032` `plan_generations_this_month()` SECURITY DEFINER global-spend-ceiling counter (`search_path=''`, returns int only).
- **CI:** `check-vendor` drift-check job (byte-identity of the 9 vendored helpers) in `ci.yml` + `preview.yml`.

## Security / cost posture
- User-JWT, RLS-scoped client; **never** reads the service-role key. All cost guards (kill-switch → auto global ceiling → cadence → per-user cap → sufficiency) short-circuit **before** any paid Claude call.
- `dryRun` is gated to service-role callers only (closes an uncapped-paid-call abuse hole); transient Claude errors serve an **ephemeral** baseline (no INSERT) so the cap/cadence aren't burned; `raw_model_output` is persisted for forensics but **stripped from the HTTP response**.
- Untrusted feedback is sanitized + wrapped in a per-call randomized delimiter with a restated trust boundary.

## Verification (local, this branch)
- `pnpm test` → **472 pass** · `pnpm typecheck` → clean (3/3) · `node scripts/check-shared-vendor.mjs` → **OK (9 files)** · `supabase db reset` applies 0001→0032 cleanly.
- Every unit passed independent spec + code-quality review; a final whole-branch integration review confirmed the seams line up (same ordered `candidates`/`articles` arrays thread through prompt→validate→store; `PLAN_TOOL` schema == `PlanDraft` == validator == storage; `COACH_NOTE_MAX` single-sourced).
- Edge code has **no vitest/typecheck** here (Deno; outside the workspace) — correctness is by inspection + the deploy smoke below.

## ⚠️ Deploy + manual-invoke smoke — REMAINING (QA, not done in this PR)
1. Set Edge secrets on dev: `ANTHROPIC_API_KEY` (+ optional `PLAN_MODEL`, `MAX_GENERATIONS_PER_USER_PER_MONTH`, `GLOBAL_MONTHLY_CAP`, `PLAN_ENGINE_ENABLED`).
2. `supabase functions deploy generate-practice-plan --use-api`.
3. **Confirm extensionless-import resolution at deploy** — the vendored `_shared/practice-plan/*` files use extensionless TS imports (byte-identical to `@oga/core`, enforced by the drift-check). `--use-api` (esbuild) should resolve them; if anything uses raw `functions serve`, an `import_map`/`deno.json` may be needed. This is the #1 thing the smoke validates.
4. Manual-invoke against the dev e2e user (≥5 SG rounds): assert the stored plan validates, references only real drills, has server-filled targets, `valid_until` = +7d, `raw_model_output` persisted; invoke again → cadence guard returns the same plan, no 2nd Claude call. Force a validation failure → baseline served + logged. Check the repair turn's `tool_result` shape and the 4-session/4096-token output sizing on a daily player.
5. Pin `@anthropic-ai/sdk` to the bundled version once known; optionally regenerate types so `plan_generations_this_month` is typed.

## Deferred (tracked, graceful — not blockers)
- **§17 corpus floor:** approach/around_green/putting have 0 anywhere-doable (`facility='{}'`) drills. Pre-PROD gate — an empty pool cleanly serves a baseline (logged) today, but limited-facility players get silent baselines until ≥1 anywhere-doable drill per category lands.
- **Dispersion slice not vendored** → digest gets `dispersion: []` (weaknesses derived from SG alone; no crash). Follow-up: vendor the dispersion slice.
- **`weaknessTargets` map not built** → retrieval falls back to deterministic `id asc` (no overlap re-ranking; pool still gated/complete). Follow-up: add a category→target-tag map.
- **Dual stored block shape** (generated = resolved `drill_id`/`target`; baseline = raw `drill_ref`) — the Phase-C Practice UI must tolerate both.
- Full golden-set eval, the two remaining validator gaps (block-id uniqueness, intra-plan repetition), and wiring the UI button — all Phase C.

Part of #18 (Phase 3). This is Phase B of the engine; Phase C + Practice UI follow.


cner-smith:
## Review — feat: AI practice plan engine — Phase B (generate-practice-plan Edge Function)

**Concerns**

- `supabase/functions/generate-practice-plan/retrieval.ts` (`pgArray`) — `pgArray` builds `{val1,val2}` by joining with `,` with no escaping, and its output is interpolated directly into PostgREST `.or()` filter strings using `goal` and every element of `facilities` from the user's profile row. A stored value containing `,` or `}` would produce a malformed filter expression, causing the candidate-pool query to error and the `throw error` path to serve a hard 500 to that user. PostgREST parameterizes its SQL, so this is a robustness issue rather than a SQL-injection path, but the failure would be unrecoverable for that user (no baseline fallback — the error is thrown before the fallback logic runs). If `goal` and `facility` are constrained to their expected enum sets by a DB check constraint, document that dependency in a comment here; otherwise add a guard that rejects or sanitizes unexpected characters before building the filter string.

- `supabase/functions/generate-practice-plan/index.ts` (`decodeJwt` / `dryRun` gate) — The `dryRun` path is gated on `decodeJwt` returning a trusted `"role": "service_role"` claim, which relies on the API gateway having `verify_jwt = true`. There is no per-function `config.toml` in the diff to make that explicit. The Supabase default is `verify_jwt = true`, but `supabase functions serve --no-verify-jwt` (a common local shorthand) disables it; in that mode a crafted JWT with `"role": "service_role"` opens the `dryRun` path, bypasses the cadence guard and per-user monthly cap, and still makes a live Claude call. The deploy checklist in the PR description (items 1–5) doesn't call this out. Worth adding an explicit `config.toml` for this function that sets `verify_jwt = true`, or at a minimum adding a checklist item to verify it at deploy time.

**Nits**

- `supabase/functions/generate-practice-plan/retrieval.ts` (`ensurePresent`) — The fallback `selected.push(candidate)` is reached when the backward loop finds no evictable slot (all 25 occupants satisfy `protect`). This expands the pool past `POOL_LIMIT = 25`. The function's JSDoc says "≤25 always" — that claim is wrong in this edge case (e.g., a corpus where the 25 gated drills are all `putting`/`pressure_game` and the only available `warmup` isn't among them). `validatePlanDraft`'s `drill_ref < candidates.length` check would still catch any out-of-range model pick, so there's no silent data corruption, but the invariant comment should be corrected — or the pool should be capped at `POOL_LIMIT` after both `ensurePresent` calls.

- `supabase/migrations/0032_plan_global_count.sql` — `date_trunc('month', now())` truncates in the session timezone. Migration 0031's unique-index expression is careful to pin `AT TIME ZONE 'UTC'` for immutability; this function doesn't. If the Supabase project's default timezone is ever set to something other than UTC, the month-boundary shifts by up to ~12 hours and the global cap counter would drift relative to calendar month boundaries. Simple fix: `date_trunc('month', now() AT TIME ZONE 'UTC')` (or `date_trunc('month', timezone('UTC', now()))`) for consistency with 0031.

---
Looked good: all new `@oga/core` exports have corresponding tests; pure logic correctly placed in `packages/core/src/`; SECURITY DEFINER function has `search_path = ''` and returns only an integer count; `sanitizeFeedback` + randomized per-call delimiter is a solid layered approach to prompt injection; skill level, goal, and facilities correctly threaded digest → retrieval → baseline (domain pillar compliance); concurrent-insert race handled at DB level (0031 unique index) and code level (23505 re-SELECT); no `SUPABASE_SERVICE_ROLE_KEY` reachable from the caller path.

---

PR #436, opened 2026-05-27
fix(edge): add .ts extensions to vendored imports (Deno deploy fix)

cner-smith:
`supabase functions deploy generate-practice-plan --use-api` failed:

> Module not found "…/_shared/practice-plan/types". Maybe add a '.ts' extension

Deno's bundler **requires** explicit `.ts` extensions on relative imports. The vendored `_shared/practice-plan/*.ts` are byte-identical copies of `@oga/core`, which omits extensions per TS convention — so they don't resolve under Deno. (This was the #1 deploy risk flagged in #435's body.)

## Fix (contained to the vendoring mechanism — no `@oga/core` / web / mobile changes)
- `scripts/vendor-transform.mjs` — `addTsExt()` appends `.ts` to relative import/export specifiers (anchored on `from`, idempotent, leaves `npm:`/`jsr:` bare specifiers alone).
- `scripts/vendor-practice-plan.mjs` — the `vendor:practice-plan` script now copies **through** `addTsExt` instead of a raw `cp`.
- `scripts/check-shared-vendor.mjs` — compares `addTsExt(coreSource) === vendoredCopy` (modulo the extension transform) so it still catches real drift while allowing the Deno-required extensions.
- Re-vendored all 9 `_shared/practice-plan/*.ts` + fixed the barrel `index.ts` exports to carry `.ts`.

`@oga/core` stays extensionless (TS-correct, unchanged); the vendored Deno copies carry `.ts`.

## Verification
- `grep` for extensionless relative imports across `supabase/functions/` → **none**.
- `node scripts/check-shared-vendor.mjs` → **OK (9 files)**; transform is idempotent (re-running vendor keeps drift green).
- `pnpm test` → 472 pass (`@oga/core` untouched) · `pnpm typecheck` → clean.

**After merge:** re-run `supabase functions deploy generate-practice-plan --use-api` — it should bundle cleanly now. (Local Deno isn't installed here, so the deploy itself is the final confirmation.)


cner-smith:
## Review — fix(edge): add .ts extensions to vendored imports (Deno deploy fix)

**Nits**
- `scripts/vendor-practice-plan.mjs:9` — `readdirSync` is obtained via a top-level `await import('node:fs')` while `readFileSync`, `writeFileSync`, and `mkdirSync` are already statically imported from the same module on line 1. The dynamic import is unnecessary; just add `readdirSync` to the existing static import.

---
Looked good: scope held entirely to the vendoring layer (`@oga/core` untouched, no web/mobile changes), `addTsExt` shared between vendor script and drift checker so transform parity is guaranteed by construction, PR targets `dev`.

---

PR #437, opened 2026-05-27
fix(core): drop unsupported JSON-schema keywords from PLAN_TOOL (Anthropic tool 400)

cner-smith:
The deployed `generate-practice-plan` function's Claude call returned a 400 (caught live in the dryRun smoke):

> `tools.0.custom: For 'integer' type, property 'minimum' is not supported`

`PLAN_TOOL.input_schema` used `minimum` (×5), `minItems` (×3), `maxLength`, and `strict: true` — none of which Anthropic's custom-tool `input_schema` accepts (`strict` is also incompatible with our optional `reason`/`article_ref` fields). Every generation was failing → silent baseline fallback.

## Fix
Per the `claude-api` skill's supported-keyword subset, `PLAN_TOOL` now uses only `type`, `enum`, `required`, `properties`, `items`, `additionalProperties: false`, and `description`. Removed: `strict: true`, all `minimum`/`minItems`/`maxLength`. The dropped bounds were always **also** enforced server-side by `validatePlanDraft` (drill_ref/article_ref range, minutes > 0, session count, coverage, coach_note length) + the repair loop — so the schema only loses redundant model-side hints; the gate is unchanged. The relevant bounds are now stated in the field `description`s as guidance.

## Guard against regression
A test asserts the serialized schema contains no `minimum`/`maximum`/`minItems`/`maxItems`/`minLength`/`maxLength` and no top-level `strict`, so a future edit can't silently re-add an API-rejected keyword.

## Verification
- `pnpm test` → 472 pass · `pnpm typecheck` → clean · drift-check OK (9 files; vendored `prompt.ts` re-synced).
- Schema keys confirmed clean in both `@oga/core` source and the vendored copy.
- **Final confirmation is the live deploy:** after merge, re-deploy and re-run the dryRun smoke — it should now build the plan and return real `drill_id`s instead of `reason: generation_failed`.

Follow-up to the Phase-B engine (#435); surfaced by the deploy smoke. No UI changes.

### DIFF
--- a/packages/core/src/practice-plan/prompt.test.ts
+++ b/packages/core/src/practice-plan/prompt.test.ts
@@ -85,26 +85,31 @@ describe('PLAN_TOOL schema', () => {
     ])
   })
 
-  it('drill_ref and article_ref are integers with minimum 0', () => {
+  it('drill_ref and article_ref are integer-typed (bounds enforced server-side by validatePlanDraft)', () => {
     expect(blockItem.properties.drill_ref.type).toBe('integer')
-    expect(blockItem.properties.drill_ref.minimum).toBe(0)
     expect(focusItem.properties.article_ref.type).toBe('integer')
-    expect(focusItem.properties.article_ref.minimum).toBe(0)
+    // minimum/maximum are NOT in the schema — Anthropic tool API rejects them (400).
+    // validatePlanDraft owns all numeric bound enforcement.
+    expect((blockItem.properties.drill_ref as Record<string, unknown>)).not.toHaveProperty('minimum')
+    expect((focusItem.properties.article_ref as Record<string, unknown>)).not.toHaveProperty('minimum')
   })
 
-  it('minutes is an integer with minimum 1', () => {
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — fix(core): drop unsupported JSON-schema keywords from PLAN_TOOL (Anthropic tool 400)

**Nits**
- `packages/core/src/practice-plan/prompt.ts:49` (`focus_areas`), `:74` (`sessions`), `:82` (`sessions[].blocks`): `minItems: 1` was removed with no compensating description update. The integer fields that lost `minimum` each got "Positive integer." or "0-based" language to give the model equivalent guidance; these array fields get nothing. The repair loop handles an empty array server-side, but a note — e.g. `description: 'One or more focus areas (at least one required).'` — restores the model-side hint cheaply and reduces repair-loop frequency. Same applies to the vendored copy in `supabase/functions/_shared/`.

---
Looked good: correct diagnosis of the Anthropic API rejection (all five removed keywords verified against the error), regression guard serializes the full schema and covers `strict` separately, vendored copy kept in sync, scope held to one concern, targets `dev`.

---

PR #438, opened 2026-05-27
fix(core): derive session total_minutes server-side; drop ±20% model check

cner-smith:
The deployed engine rejected a generated plan (caught in the dryRun smoke):

> session "…Approach Deep Dive" minutes 85 vs total 65 >20% → baseline

The model declared a 65-min session whose blocks summed to 85; `validatePlanDraft`'s ±20% rule rejected it, the repair retry didn't fix it → baseline. `total_minutes` is a **derived** value (the sum of block minutes), so — exactly as §10 already does for *targets* — the server should compute it rather than have the model emit a value that can disagree with the blocks.

## Fix (2 files + tests; schema untouched)
- `storage.ts` — `resolvePlanForStorage` now sets each session's `total_minutes = Σ block.minutes`, overriding the model's value. The stored/displayed session length is correct-by-construction.
- `validation.ts` — removed the D4 `total_minutes` finite/positive + ±20% checks (and the now-unused `MINUTE_TOLERANCE`/`sum`). The per-block `minutes > 0 && finite` check stays — that's what keeps the derived sum valid. All other checks (drill_ref/article_ref range, drill_type, session count, top-2 coverage, anti-repetition, coach_note length) are unchanged.
- The model still emits `total_minutes` (harmless hint); it's now ignored by validation and overridden in storage — so no schema re-touch / no new deploy risk.

## Verification
- `pnpm test` → 473 pass (incl. a new storage test: draft total 65 / blocks summing 43 → stored `total_minutes === 43`; and a validation test: a divergent session now passes). `pnpm typecheck` → clean. Drift-check OK (9 files; `validation.ts` + `storage.ts` re-vendored).
- **Final confirmation is the live deploy:** after merge + re-deploy, the dryRun should return a stored generated plan (no `reason: generation_failed`).

Follow-up to #435/#437; surfaced by the deploy smoke. No UI changes.

### DIFF
--- a/packages/core/src/practice-plan/storage.test.ts
+++ b/packages/core/src/practice-plan/storage.test.ts
@@ -119,4 +119,13 @@ describe('resolvePlanForStorage', () => {
     expect(out.focus_areas[0]!.article).toBeUndefined()
     expect(spy).toHaveBeenCalled()
   })
+
+  it('sets stored total_minutes to the sum of block minutes, overriding the model-emitted value', () => {
+    // Draft declares total_minutes=65 but blocks sum to 8+20+15=43.
+    // Storage must derive the real total from block minutes, not copy the model hint.
+    const mismatch: PlanDraft = structuredClone(draft)
+    mismatch.sessions[0]!.total_minutes = 65 // deliberately wrong; blocks still sum to 43
+    const out = resolvePlanForStorage(mismatch, ctx)
+    expect(out.drills.sessions[0]!.total_minutes).toBe(43) // 8+20+15
+  })
 })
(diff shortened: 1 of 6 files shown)


cner-smith:
## Review — fix(core): derive session total_minutes server-side; drop ±20% model check

**Concerns**
- `packages/core/src/practice-plan/validation.ts:40` (and `supabase/functions/_shared/practice-plan/validation.ts:40`) — The removed D4 check did double duty: it also caught zero-block sessions, because a model-emitted `total_minutes ≤ 0` would fail validation before reaching storage. With the D4 removal, a session whose `s.blocks` array is empty passes `validatePlanDraft` and then `resolvePlanForStorage` silently stores `total_minutes: 0` (the `reduce` identity). If there is a session-structure check elsewhere that guarantees `blocks.length ≥ 1`, this is fine — but it isn't visible in the diff. Worth either confirming the assumption is covered or adding a `blocks.length === 0` guard in validation alongside the per-block check.

---
Looked good: server-side derivation is the right fix for a derived value, per-block `minutes > 0` check stays intact so the sum is guaranteed positive when blocks are non-empty, both new tests are meaningful (override case + surviving per-block positivity), the `supabase/_shared` copies are in sync with `packages/core`, and the scope is held to one concern.

---

PR #439, opened 2026-05-27
feat(web): Practice plan page — render generated plan + Generate button (Track 3)

cner-smith:
Track 3 of the AI practice plan (#18): the web Practice page now **renders the plan the engine generates** and wires the **Generate** button — replacing the "Coming in phase 5" placeholder. Built to the `docs/design/future-features.html #practice` mockup, in the paper-editorial system (`DESIGN.md`).

## Scope (decided with the owner)
**Core render + Generate**, **faithful-sessions** layout. Per-block completion checkboxes and the feedback box are a deferred follow-up; mobile + the drill-library page are untouched.

## What's in here
- **Data layer:** `getDrillsByIds` (`@oga/supabase`), `useGeneratePlan()` (the first `supabase.functions.invoke` from web — invalidates the plan query), `useDrillsByIds()` to resolve block `drill_id`s → names. `useLatestPracticePlan` already existed.
- **`PracticePlanPage`:** renders the reasoning panel (`coach_note`), focus areas (with `/learn/:slug` citations), and the engine's sessions as numbered drill cards (name + why + facility pills + minutes + type tag + server-filled target). Six states: loading / no-plan (Generate CTA) / generating / live (no regen within the weekly window) / soft-expired ("last week's" + Generate) / error (`toUserMessage`). Reads the canonical `@oga/core` `StoredSession`/`StoredFocusArea`/`StoredBlock` types; degrades safely on null/empty/missing-drill data.

## Cross-check (multi-agent, as requested)
- **Spec + design:** ✅ no `DESIGN.md` violations, all 6 states correct, date/no-regen logic sound (no UTC off-by-one), faithful-sessions correctly implemented, scope held.
- **Code quality:** the one Important finding (local types duplicating + drifting from `@oga/core`'s canonical stored shapes) is **fixed** — the page now imports them; dead `ACCENT` const removed; `target` rendering polished for legibility.
- **Visual:** rendered via a throwaway harness with a representative plan + screenshotted — strong match to the mockup (warm paper, Fraunces headline, mono kickers, numbered drill cards, accent used sparingly, hairlines, no shadows).
- `pnpm typecheck` + `pnpm --filter web build` clean.

## Deferred follow-ups
- Per-block completion (`completed_drill_ids`) + the feedback box (needs `saveFeedback` in `@oga/supabase`) — fast next pass; the feedback box activates the loop the engine already reads.
- Mobile Practice screen; the drill-library page.
- Authoring drill `target_template`s so blocks show numeric targets (most are null today).

Part of #18 (Phase 3). No closing keyword — completion/feedback land in the follow-up.

### DIFF
--- a/apps/web/src/hooks/useDrills.ts
+++ b/apps/web/src/hooks/useDrills.ts
@@ -1,6 +1,6 @@
-import { useQuery } from '@tanstack/react-query'
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import type { ShotCategory } from '@oga/core'
-import { getDrills, getLatestPracticePlan } from '@oga/supabase'
+import { getDrills, getDrillsByIds, getLatestPracticePlan } from '@oga/supabase'
 import { supabase } from '../lib/supabase'
 import { useAuth } from './useAuth'
 
@@ -28,6 +28,36 @@ export function useLatestPracticePlan() {
       const { data, error } = await getLatestPracticePlan(supabase, user!.id)
       if (error) throw error
       return data
+    },
+  })
+}
+
+export function useGeneratePlan() {
+  const { user } = useAuth()
+  const qc = useQueryClient()
+  return useMutation<void, Error>({
+    mutationFn: async () => {
+      const { error } = await supabase.functions.invoke('generate-practice-plan', { body: {} })
+      if (error) throw error
+    },
+    onSuccess: () => {
+      qc.invalidateQueries({ queryKey: ['practice-plan', user?.id] })
+    },
+  })
+}
+
+export function useDrillsByIds(ids: string[]) {
+  const sortedIds = [...ids].sort()
+  return useQuery({
+    queryKey: ['drills-by-ids', sortedIds],
+    enabled: ids.length > 0,
+    queryFn: async () => {
+      const { data, error } = await getDrillsByIds(supabase, ids)
+      if (error) throw error
+      const drills = data ?? []
+      const byId: Record<string, (typeof drills)[number]> = {}
+      for (const d of drills) byId[d.id] = d
+      return byId
     },
   })
 }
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — feat(web): Practice plan page — render generated plan + Generate button (Track 3)

**Concerns**

- `apps/web/src/pages/practice/PracticePlanPage.tsx:96–115` — `PrimaryButton` has exactly one caller (`GenerateButton`), and `ErrorNotice` has one caller (the main `PracticePlanPage` body). CONTRIBUTING.md's 3-caller rule: "Don't extract a helper, hook, or component into its own function or file unless it has three or more callers, or the inline version is genuinely unreadable." Neither case clears that bar. `PrimaryButton` is a style wrapper that could be folded directly into `GenerateButton` (keeping one named component for the two places it's used); `ErrorNotice` is four lines and straightforward to inline at the single call-site. Worth cleaning up rather than carrying the pattern into the follow-up PRs.

- `apps/web/src/pages/practice/PracticePlanPage.tsx:117–194` — `GenerateButton` (2 callers: `NoPlanState`, `PlanHeader`) and `PageHeading` (2 callers: `LoadingState`, `NoPlanState`) both fall short of the 3-caller threshold. The inline versions would be noisy enough that there's a genuine readability argument — this is closer to the line than the cases above — but flagging so the rule is visible if a reviewer is strict about it.

**Nits**

- `packages/supabase/src/queries/practice.ts:47–65` — `DrillCard` is declared after the function that casts to it, and the type is unexported. Moving it above the function and matching it to what Supabase actually infers from the `select()` columns would make both the early-return cast and the type relationship to the query result explicit rather than implicit. Low impact but can drift silently if the column list changes.

---
Looked good: correct `dev` base, no `any`, `@oga/core` canonical types imported (not re-declared), date logic sound (local ISO string construction avoids UTC off-by-one, lexicographic YYYY-MM-DD comparison is correct), sorted cache key in `useDrillsByIds` prevents key thrashing on reorder, `enabled: ids.length > 0` guard keeps the drill fetch from firing on empty plans, six states fully covered.

---

PR #440, opened 2026-05-27
fix(edge): add CORS to generate-practice-plan (browser invoke was preflight-blocked)

cner-smith:
Surfaced by driving the real Practice page in a headless browser: clicking **Generate** failed with
`blocked by CORS policy: No 'Access-Control-Allow-Origin' header`. The `generate-practice-plan` Edge Function shipped **no CORS** — no `OPTIONS` preflight handler and no `Access-Control-Allow-*` headers — so `supabase.functions.invoke` from the web app fails the preflight before the request is even sent. The button can't work in a browser without this.

The code reviews couldn't catch it (they read the function, which "mirrors `round-summary-email`" — but that's a server-to-server cron with no browser caller, so it legitimately has no CORS). Only a live browser invoke surfaces it. The engine itself is fine: invoking from Node (no preflight) generated + inserted a real plan, and the page renders it.

## Fix
- `corsHeaders` const (`Access-Control-Allow-Origin: *`, the supabase-js request headers, `POST, OPTIONS`).
- `OPTIONS` preflight handled first in the `Deno.serve` handler → 200 + CORS headers.
- `json()` helper merges `corsHeaders` into every response, so all 15 response paths (guards, live-plan, dryRun, insert success, 23505 race, transient 503, baselines) carry CORS. Audited — no raw `new Response` left uncovered.
- `*` origin is correct: the security boundary is the verified `Authorization: Bearer <user-jwt>` (auth in a header, not a cookie → no `Allow-Credentials`), and `*` doesn't break Vercel preview URLs. `round-summary-email` left as-is (server-to-server cron, no browser caller).

## After merge
Re-deploy: `supabase functions deploy generate-practice-plan --use-api`. Then the Practice page's Generate button (Track 3, PR #439) works end-to-end in the browser.

Follow-up to #435; required for #439's button to function. No web/`@oga/core` changes.

### DIFF
--- a/packages/core/src/practice-plan/digest.test.ts
+++ b/packages/core/src/practice-plan/digest.test.ts
@@ -75,4 +75,58 @@ describe('buildPlayerDigest', () => {
     })
     expect(d.warnings ?? []).toHaveLength(0)
   })
+
+  it('rounds averages to 2 decimal places in the digest output', () => {
+    const d = buildPlayerDigest({
+      profile: { skill_level: 'developing', handicap_index: 12.4, goal: 'break_80',
+        facilities: ['range'], play_frequency: 'weekly' },
+      // approach mean = (-1.2 + -1.0 + -1.133...) / 3  →  an ugly decimal
+      // Use values that produce a 3-dp raw mean to prove rounding
+      rounds: [
+        r({ sg_approach: -1.2, sg_putting: -0.4, sg_off_tee: 0.1, sg_around_green: 0.2 }),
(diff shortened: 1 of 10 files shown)


cner-smith:
## Review — fix(edge): add CORS to generate-practice-plan (browser invoke was preflight-blocked)

---
Looked good: `*` origin correctly paired with bearer-token auth and no `Allow-Credentials`; `json()` helper covers all response paths; `OPTIONS` path returns a raw `Response` (correct — preflight should not be JSON); single concern, targets `dev`.

---

PR #441, opened 2026-05-27
feat(web): per-block completion + plan feedback box (Track 3, final piece)

cner-smith:
The deferred half of Track 3 — makes the rendered plan interactive. Builds on the merged Practice page (#439).

## What's in here
- **Per-block completion:** an accessible checkbox (`role="checkbox"`, accent-fill when done) at the left of each drill card — distinct from the name (which still toggles the instructions). Click writes `completed_drill_ids` **optimistically** (`useUpdatePlanProgress`, cache snapshot + rollback-on-error + invalidate-on-settled). Completed blocks dim their name (no strikethrough — editorial); each session kicker shows "N of M done".
- **Feedback box:** a "BEFORE NEXT WEEK" section below the sessions — coached prompt + textarea (≤500, live counter), pre-filled from `plan.feedback`, **saves on blur** (`useSaveFeedback`) with a subtle "Saved" indicator. One note per plan; **this is what next week's generation reads** (the engine's feedback loop is now reachable from the UI).
- Data: `saveFeedback` (`@oga/supabase`) + the two web hooks. `updatePlanProgress` already existed.

## Cross-check (both ✅ ready to merge)
- **Code review:** optimistic update correctness, a11y (checkbox independent of expand), feedback-save logic, DESIGN.md, non-regression — all pass.
- **Visual (real app, e2e plan):** toggle flips optimistically **and persists on reload** (`completed_drill_ids: ["s1b1"]`); session progress updates; done-state dims; feedback saves + persists + pre-fills on reload; checkbox and expand are independent; on-brand, no layout/contrast issues.

## Non-blocking follow-up notes (from review — optional)
- Rapid double-toggle: standard optimistic tradeoff; `onSettled` re-syncs from server.
- `role="checkbox"` on a `<button>` works with modern AT; a native hidden checkbox would be more idiomatic.
- Accent appears on checkboxes + type-tags + links on a populated plan (visual confirmed on-brand; worth an eyeball).
- `FeedbackSection` keeps its `initial` if a plan is replaced while mounted (gated by the no-regen-while-current rule; `key={plan.id}` would bulletproof it).

Test data left on the dev e2e plan (`2a8d6bdf`): one completed block + a feedback note — harmless. Part of #18 (Phase 3). No engine/mobile changes.

### DIFF
--- a/apps/web/src/hooks/useDrills.ts
+++ b/apps/web/src/hooks/useDrills.ts
@@ -1,6 +1,12 @@
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import type { ShotCategory } from '@oga/core'
-import { getDrills, getDrillsByIds, getLatestPracticePlan } from '@oga/supabase'
+import {
+  getDrills,
+  getDrillsByIds,
+  getLatestPracticePlan,
+  saveFeedback,
+  updatePlanProgress,
+} from '@oga/supabase'
 import { supabase } from '../lib/supabase'
 import { useAuth } from './useAuth'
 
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — feat(web): per-block completion + plan feedback box (Track 3, final piece)

**Concerns**
- `apps/web/src/hooks/useDrills.ts:117-122` — `useSaveFeedback.onSuccess` calls `setQueryData` then immediately `invalidateQueries`. The invalidation marks the cache stale and schedules a background refetch that overwrites whatever `setQueryData` just put there — the `setQueryData` call is dead, with no observable effect. The result is a redundant network round-trip on every blur-save, on top of the mutation itself. `useUpdatePlanProgress` in the same file has the right shape: `onMutate` for the optimistic write, `onSettled` for the invalidation. `useSaveFeedback` should follow the same pattern — either drop `setQueryData` and rename `onSuccess` → `onSettled`, or drop `invalidateQueries` if you want to skip the refetch entirely.

**Nits**
- `apps/web/src/pages/practice/PracticePlanPage.tsx:716` — The second guard in `handleBlur` (`if (next.length === 0 && savedValue.trim().length === 0) return`) is unreachable. When both sides are empty, `next === savedValue.trim()` is `'' === ''` → true, so the first guard exits first. Dead branch.
- `apps/web/src/pages/practice/PracticePlanPage.tsx:864` — `<FeedbackSection>` has no `key`. The PR description already flags `key={plan.id}` as a one-prop fix that would bulletproof against stale state when the plan is replaced while the component is mounted. Worth including here rather than carrying forward as a known gap.

---
Looked good: RLS on `practice_plans` (`for all using (auth.uid() = user_id)`) covers `saveFeedback`'s update-by-id-only pattern without needing an explicit `user_id` predicate in the query; `saveFeedback` lands in `packages/supabase` not `apps/web/lib`; scope held to one feature area.

---

PR #442, opened 2026-05-27
feat(db): drill_type → practice-mode taxonomy (foundation)

cner-smith:
## What

Redefines `drills.drill_type` as the practice **mode** and re-tags the corpus to match the practice pedagogy:

`warmup · blocked · random · skill_game · pressure_game · on_course` (category stays the area).

- **`technical` retired** — it collapsed the blocked↔random distinction. Split into `blocked` (same shot repeated) and `random` (a different shot every ball).
- **`on_course` added**; **unused `putting` type dropped** (putting is a category, not a mode).
- **"AimPoint Express Green Reading" removed** — a taught green-reading method, not a drill.
- **4 score-to-beat drills** (Tipping-Point Speed Test, Towel Landing-Zone Pitch, Tee-Corridor Lag, Eyes-Closed Speed Calibration) moved `technical → skill_game`; the rest `technical → blocked`.

## Result — 40 drills

`warmup 5 · blocked 19 · skill_game 12 · pressure_game 4` · `random 0` · `on_course 0` (the empty modes are authored in a follow-up).

## Files
- `supabase/migrations/0033_drill_type_practice_modes.sql` — drops the old CHECK (constraint-name-robust), re-tags existing rows, swaps the default + CHECK.
- `supabase/seed.sql` — same re-tag for a fresh `db reset`.

## ⚠️ Deploy note
**Foundation only.** The plan engine still uses the old block-type vocab (`BlockType`, baseline, validation). **Do not apply `0033` to a project's database until the engine follow-up ships** — otherwise plan generation degrades to the baseline fallback (the model's allowed block types won't match the re-tagged drills). Merging this PR is safe; migrations apply to the DB manually, not on merge.


cner-smith:
## Review — feat(db): drill_type → practice-mode taxonomy (foundation)

**Concerns**
- `supabase/migrations/0033_drill_type_practice_modes.sql:21` — The AimPoint row is deleted by `name`, not by a stable identifier. If that row's `name` was ever edited in a production database, the `DELETE` silently no-ops. Step 4 then re-tags it to `blocked` (its `drill_type` is still `'technical'`), so the drill survives in the corpus with the wrong semantics and the intended removal doesn't happen. Not a crash, but an invisible correctness issue. Matching on a UUID, or at least asserting `GET DIAGNOSTICS row_count = ROW_COUNT` and raising if zero, would make the failure visible.

- `supabase/migrations/0033_drill_type_practice_modes.sql:42–44` — The new `CHECK` excludes `'putting'`, but no step before it re-tags or deletes rows where `drill_type = 'putting'`. The migration relies on the assertion that `putting` is unused, but doesn't enforce it. If any production database has a `putting` row (from the old allowed-values list in 0030), `ADD CONSTRAINT` at step 5 will fail with a constraint-violation error. The failure is loud rather than silent, which is acceptable, but a one-liner `UPDATE public.drills SET drill_type = 'blocked' WHERE drill_type = 'putting';` between steps 4 and 5 — or at minimum a comment asserting the invariant — removes the deploy surprise. The PR's own warning ("do not apply until the engine follow-up ships") reduces the blast radius, but the guard costs one line.

---
Looked good: constraint-name discovery via `pg_constraint` handles the auto-named constraint from 0030 robustly without assuming a name; migration and seed.sql updated in tandem so both the `db reset` and the migration path stay consistent; PR targets `dev`; scope held to one concern (taxonomy data only); no TypeScript/app code touched, no RLS or package-boundary issues.

---

PR #444, opened 2026-05-28
feat(db): v2 corpus fill across all 6 practice modes (40 → 125 drills)

cner-smith:
## What this PR does

Fills the practice-mode taxonomy (migration 0033, PR #442) with sourced, fact-checked drills across every mode and area. **40 → 125 drills (+85).**

### Mode counts (was → now)
| mode | before | after | new |
|---|---|---|---|
| `warmup` | 5 | 14 | +9 |
| `blocked` | 19 | 44 | +25 |
| `random` | 0 | 15 | +15 (honest ceiling) |
| `skill_game` | 12 | 27 | +15 |
| `pressure_game` | 4 | 15 | +11 (honest ceiling) |
| `on_course` | 0 | 10 | +10 (honest ceiling) |

Realistic total ceiling at the source bar ≈ 135–170 (not 300). Three modes (`random` / `on_course` / `pressure_game`) cap around 15 each — they're protocol-bound, so distinct well-sourced drills dedup hard once you fetch and verify each candidate.

### Source bar (enforced on every drill)

- **Real, currently-live source:** exact URL that loads + a named PGA pro, teaching authority, or reputable publication (Golf.com, Golf Digest, PGA.com, Golf Monthly, Titleist, Today's Golfer, GolfTEC, Golf Insider UK, Practical Golf, etc.).
- **Every numeric/attribution claim verified** against the source — no hallucinated drills, pros, or URLs.
- **No branded teaching SYSTEMS** that require the pro's curriculum to execute (Short Game Chef, Joe Mayo Spectacular Strike, Sieckmann finesse-wedge framework, AimPoint Express, SeeMore RST, Haney WIDE-WIDE-NARROW). Test: can a player execute from the drill copy alone?
- **Honest ceiling at the source bar** — a mode lands below 50 (reported) rather than padded with dupes/fakes.

### Mode-boundary rules (locked + enforced)

- **`skill_game` = "a recorded score you replay to beat"** (owner rule, 0033).
- **`pressure_game` = consequence:** restart-on-miss / consecutive-make quotas / pass-fail Tour standards. Many candidates with "challenge" / "test" / "scoring game" in their source title turned out to be pressure_game on inspection.
- **`blocked` = universal techniques with clear what+when.** No curriculum-locked systems.
- **`random` = different shot every ball.** Protocol-bound; a random-protocol drill with a kept score is `skill_game`, not `random`.
[...]

PR #445, opened 2026-05-29
feat(db): migration 0034 — v2 drill corpus data (85 inserts, 40 → 125)

cner-smith:
## What this PR does

Adds migration `0034_drill_corpus_v2_data.sql` that bulk-inserts the 85 v2 drills committed in PR #444 (which only touched `seed.sql`). This makes the v2 corpus portable to remote Supabase projects, not just `npx supabase db reset` locally.

## Why a migration

`seed.sql` is a local-only convenience — it runs on `supabase db reset` but doesn't auto-roll to remote projects. Without a migration, every PR that changes the corpus produces drift between the repo's `seed.sql` and what's actually in the dev/prod `drills` tables. Migration 0034 closes that gap permanently.

## Changes

- Adds a UNIQUE constraint on `drills.name` (the corpus has always satisfied this — every drill name is distinct; making it a constraint enables idempotent re-runs).
- Bulk-inserts the 85 v2 drills (random + on_course + pressure_game + warmup + blocked + skill_game additions) using `ON CONFLICT (name) DO NOTHING`.

## State after this migration applies

| mode | count |
|---|---|
| `warmup` | 14 |
| `blocked` | 44 |
| `random` | 15 |
| `skill_game` | 27 |
| `pressure_game` | 15 |
| `on_course` | 10 |
| **total** | **125** |

Matches `seed.sql` exactly.

## Applied to dev

Already applied to `oga-dev` (`txquvfeyvkaetqlamqlz`) via MCP earlier in the same session:
- Migration `0034_drill_corpus_v2_data` registered (constraint piece)
- 85 INSERT rows applied via `execute_sql` in 5 batches of 17
- Dev DB row counts verified against `seed.sql` exactly (125 total, mode counts match)

## Prod (`oga`, `znjmkptblswxststjxwc`)

Not yet applied. Apply at launch-time via the standard migration flow (`supabase db push` or Supabase Dashboard). The full SQL in this file produces the correct final state from a fresh project that already has 0033 applied. Prod's drill bulk rollout was historically deferred to launch — this PR is what makes that rollout a one-command operation.

## Type of change
- [x] New feature (DB migration + data)
- [ ] Bug fix
- [ ] Design/UI change
- [ ] Refactor
- [ ] Docs

## Testing
- [x] Applied to dev, verified 125 drill rows materialize with correct mode counts
- [x] `ON CONFLICT (name) DO NOTHING` enables safe re-runs
- [x] Constraint `drills_name_unique` registered on dev
- [ ] pnpm typecheck — N/A (data-only change)
- [ ] pnpm test — N/A (data-only change)

## Notes for reviewer

- **Pure data migration** — no schema changes beyond the unique constraint. Slots into the existing taxonomy from migrations 0030 + 0033.
- **219KB file** because each drill includes a verbose 7-part `instructions` body (~500 words). That's by design — the data is the deliverable.
- **Idempotent** via `ON CONFLICT (name) DO NOTHING`. Safe to re-apply on a project that already has any subset of these drills.
- **Branches off `dev`**, independent of PR #444 (which committed the same content to `seed.sql`). Both files are now in sync; the migration is the portable copy.


cner-smith:
## Review — feat(db): migration 0034 — v2 drill corpus data (85 inserts, 40 → 125)

**Concerns**
- `supabase/migrations/0034_drill_corpus_v2_data.sql:1` — `ALTER TABLE public.drills ADD CONSTRAINT drills_name_unique UNIQUE (name)` will error and abort the entire migration if any existing rows have duplicate `name` values. For a freshly seeded environment this is fine, and the dev application described in the PR description confirms it passed there. But the risk surface is any project that accumulated drill rows outside of `seed.sql` — manual inserts during testing, partial resets, experimental data. PostgreSQL will scan the full table and reject the `ALTER TABLE` with no rollback path short of manual deduplication. The migration file itself doesn't document this pre-condition anywhere. At minimum, a SQL comment at the top stating the assumption ("assumes all existing rows in `public.drills` have distinct `name` values") makes the operational requirement explicit for whoever runs `supabase db push` to prod.

**Nits**
- `supabase/migrations/0034_drill_corpus_v2_data.sql` (INSERT column list) — The tail of the INSERT values appears to include columns — a source attribution string, a URL, a source tag (`'oga'`), and a boolean — that are not listed in CONTRIBUTING.md's drill contribution guide under "Adding drills." The documented schema there shows `id, name, description, duration_min, category, facility, skill_levels, instructions`. Since this migration is the largest corpus expansion to date, it's the natural moment to note that CONTRIBUTING.md's drill schema reference will mislead future contributors who use it to write their own `INSERT`s. Not this PR's responsibility to fix, but a follow-up issue or a one-liner in the PR description pointing to the discrepancy would help.

---
Looked good: pure data migration, targets `dev` (not `main`), idempotent `ON CONFLICT (name) DO NOTHING`, no secrets or credentials, no new tables requiring RLS, no non-SQL changes, single concern held throughout.

---

cner-smith:
Addressed in 3fb7fb9:

- **Concern (pre-condition):** added a `⚠️ PRE-CONDITION` block to the migration header that documents the assumption + provides the exact dedupe query to run before `supabase db push` against a project you don't fully control. Comment-only; no SQL behavior change.
- **Nit (CONTRIBUTING.md drill schema discrepancy):** skipping per the review's "not this PR's responsibility" — will file separately.

Tacked on a stale-test fix (`apps/web/e2e/practice.signed-in.spec.ts`) for a Playwright failure that came up while smoke-testing today's work. The test was written when `/practice` was a placeholder (literal `<h1>Practice plan</h1>` + a "Browse drill library" link); PR #441 replaced both. Test now verifies URL stability on `/practice` (no auth redirect) and the "The full set" heading on `/practice/drills` via direct nav. 12/12 chromium-auth tests pass.

PR #446, opened 2026-05-29
feat(web): generation progress bar + fix(engine): treat APIConnectionError as transient

cner-smith:
## Why

Surfaces a real bug from today's smoke test: a Claude API timeout during plan generation (`APIConnectionTimeoutError` after 61s wall-clock) was misclassified as a non-transient error → the orchestrator **persisted** a baseline plan with `valid_until = today + 7` → the cadence guard then **locked the user out of regenerating for 7 days**.

Manually deleted the broken baseline row from dev for the affected user (TestGolfer); without (C), this lockout would recur on any future Claude blip.

## (B) Generation progress bar

Claude calls can legitimately take 15–45s. Static "Generating…" button text gave zero signal that the call was alive — users (myself included) thought the click did nothing.

- New `GenerateProgressBar` component (local to `PracticePlanPage.tsx`). Thin (3px) indeterminate sliding gradient, only renders when `visible=true`.
- Placed above the GenerateButton on **both** entry points:
  - **First-time generation** — top of the `NoPlanState` card
  - **Regenerate** (after expiry) — top of the `PlanHeader`
- Pure CSS animation (`@keyframes oga-progress-slide`), inlined in the component. No library, no global stylesheet change.

## (C) Classify `Anthropic.APIConnectionError` as transient

`classifyClaudeError` already matches `RateLimitError` / `InternalServerError` / `overloaded_error` / `status>=500` as transient — but a **connection-level failure** (no HTTP response at all; the SDK never reached Anthropic) is just as retryable.

\`\`\`diff
 const transient =
   err instanceof Anthropic.RateLimitError ||
   err instanceof Anthropic.InternalServerError ||
+  err instanceof Anthropic.APIConnectionError ||
   type === 'overloaded_error' ||
   err.status === 429 ||
   (typeof err.status === 'number' && err.status >= 500)
\`\`\`

`APIConnectionError` is the SDK's base class for connection failures and covers `APIConnectionTimeoutError` (today's failure) via inheritance.

With this fix, a timeout returns **503 ephemerally** — no INSERT, the per-user cap isn't burned, and the cadence guard isn't tripped, so the user can retry immediately.

## Scope

- **2 files, +79 / −22 lines**
- ≤30 LOC web edit (new component + 2 mount points)
- ~2 LOC + comment block on the engine
- No schema change, no contract change
- Typecheck green across the workspace

## Testing
- [x] pnpm typecheck — green (3/3)
- [ ] pnpm test — N/A (no test coverage on engine error classification; would be a follow-up)
- [ ] Visual verify in browser — pending preview deploy
- [ ] Engine smoke test — pending Edge function redeploy + a fresh Generate click

## Deployment

- Web side ships with the Vercel preview / dev merge.
- **Engine side (C) requires an Edge function redeploy.** I'll fire it via MCP right after this PR opens so dev's `generate-practice-plan` v6 picks up the classification change.

### DIFF
--- a/apps/web/src/pages/practice/PracticePlanPage.tsx
+++ b/apps/web/src/pages/practice/PracticePlanPage.tsx
@@ -206,6 +206,44 @@ function GenerateButton({
   )
 }
 
+/** Indeterminate sliding progress bar shown while generation is in flight.
+ *  Claude calls can take 15–45s; static "Generating…" text gives no signal
+ *  the call is alive. The bar is the only motion cue and lives at the top of
+ *  whichever generate surface (NoPlanState card or PlanHeader) the user
+ *  clicked. Keyframes inlined so the change is contained to this file. */
+function GenerateProgressBar({ visible }: { visible: boolean }) {
+  if (!visible) return null
+  return (
+    <div
+      aria-hidden="true"
+      style={{
+        position: 'relative',
+        height: 3,
+        width: '100%',
+        background: '#EBE5D6',
+        overflow: 'hidden',
+        borderRadius: 2,
+      }}
+    >
+      <style>{`
+        @keyframes oga-progress-slide {
+          0%   { transform: translateX(-100%); }
+          100% { transform: translateX(400%); }
+        }
+      `}</style>
+      <div
+        className="bg-caddie-accent"
+        style={{
+          position: 'absolute',
+          inset: 0,
+          width: '25%',
+          animation: 'oga-progress-slide 1.4s ease-in-out infinite',
+        }}
+      />
+    </div>
+  )
+}
+
 function PageHeading() {
   return (
     <>
@@ -271,8 +309,18 @@ function NoPlanState({
       </div>
       <div
         className="bg-caddie-surface"
-        style={{ border: `1px solid ${LINE}`, borderRadius: 4, padding: '40px 32px', maxWidth: 640 }}
+        style={{
+          border: `1px solid ${LINE}`,
+          borderRadius: 4,
+          padding: '40px 32px',
+          maxWidth: 640,
+          position: 'relative',
+        }}
       >
+        {/* Top-of-card progress bar while generation is in flight. */}
+        <div style={{ position: 'absolute', top: 0, left: 0, right: 0 }}>
+          <GenerateProgressBar visible={isPending} />
+        </div>
         <div className="kicker" style={{ marginBottom: 12 }}>
           No plan yet
         </div>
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — feat(web): generation progress bar + fix(engine): treat APIConnectionError as transient

**Concerns**

- `apps/web/src/pages/practice/PracticePlanPage.tsx` + `supabase/functions/generate-practice-plan/index.ts` — Two distinct concerns are bundled here: a UI feature (progress bar) and an engine bug fix (error classification). CONTRIBUTING.md §"Opening a PR": "One concern per PR. UI restyle, bug fix, and new feature do not belong in the same PR." I understand the bug fix is urgent and both changes are small, but the rule is unambiguous. Noting it so the pattern doesn't carry forward.

- `apps/web/src/pages/practice/PracticePlanPage.tsx:402` — The `<div style={{ marginBottom: 8 }}>` wrapper inside `PlanHeader` is always rendered, even when `GenerateProgressBar` returns `null` (i.e., when `!isPending`). An empty block with `marginBottom` still shifts subsequent siblings in the document flow, so the flex row containing the kicker, title, and button gains a permanent 8 px top-offset in the normal (non-generating) state — a layout regression. The `NoPlanState` placement doesn't have this problem because its wrapper is `position: absolute`. Fix: move `marginBottom` inside `GenerateProgressBar` so the gap only exists when the bar renders, or conditionally render the outer wrapper (`{isPending && <div …><GenerateProgressBar visible /></div>}`).

**Nits**

- `apps/web/src/pages/practice/PracticePlanPage.tsx:209` — The four-line JSDoc narrates what the component does and how it's used ("The bar is the only motion cue and lives at the top of whichever generate surface…"). CONTRIBUTING.md: "Comments only when the *why* is non-obvious from the name and types alone. Don't narrate what the code does." The name `GenerateProgressBar` and the `visible` prop are self-explanatory; the only genuinely non-obvious detail is why keyframes are inlined rather than in a stylesheet. One sentence covering that would satisfy the rule.

---
Looked good: correct `dev` target, no schema changes, no security issues, `APIConnectionError`→`APIConnectionTimeoutError` inheritance is accurate, and `GenerateProgressBar` is co-located in the parent file rather than split out.

---

cner-smith:
Addressed in 84a1a97:
- **Concern (idle-state 8px shift):** wrapper now conditional on `isPending` — when no generation is in flight, neither the wrapper nor its `marginBottom` are in the DOM, so the title row sits at its original position.
- **Nit (JSDoc density):** trimmed to one sentence (the keyframes-inlined detail). Name + `visible` prop carry the rest.
- **Two-concern bundling:** acknowledged; future PRs split.

PR #449, opened 2026-05-29
feat(plan): resolve baseline drill_id by corpus name (#18 Track D)

cner-smith:
## What

When the orchestrator falls back to a baseline plan (insufficient SG rounds, transient Claude error, generation failure), the persisted `drills.sessions[*].blocks[*].drill_id` was undefined — the UI fallback name was the literal string "Drill" with `canExpand = false`. Deferred Phase B work from the engine ship.

## Approach (locked plan, Approach A)

Hardcode a specific corpus drill **name** per baseline block. At `serveBaseline` time do ONE query `select id, name from drills where name in (...)`, build a name→id map, fill `drill_id` on each stored block, drop the `drill_name` hint from the persisted shape.

Editorial control + deterministic + obvious in source.

## Changes

- `PlanBlock` (`packages/core/src/practice-plan/types.ts`): adds optional `drill_name?: string` — baseline-only hint, never emitted by Claude tool calls.
- `baseline.ts`: sets `drill_name` on each of the 11 blocks across the two default variants (putting-focus + approach-focus). All 11 picks come from the v2 corpus seed (migration 0034). Names verified verbatim against `supabase/seed.sql`.
- `baseline.test.ts`: asserts every baseline block carries a non-empty `drill_name`.
- `_shared/practice-plan/{types,baseline}.ts`: vendor copies via `scripts/vendor-practice-plan.mjs`; drift-check OK (9 files).
- `generate-practice-plan/index.ts` `serveBaseline()`: collects distinct names, queries `drills`, builds a name→id map, produces `resolvedSessions` carrying `drill_id` with `drill_name` stripped. **Graceful fallback:** unresolved name (e.g., a prod project without migration 0034 applied yet) passes through with `drill_id` undefined — same as the pre-D behavior, so a missing corpus never blocks a baseline. The lookup error logs but doesn't fail.

All three `serveBaseline` exit paths (dryRun, transient-ephemeral, INSERT) carry the resolved sessions because resolution happens before the branch.

## Picks

Variant A (putting focus):
| Block | Drill |
|---|---|
| `b-put-warmup` | Speed Station (15-Inches-Past Lag Primer) |
| `b-put-blocked` | Gate Drill |
| `b-put-game` | Golden Eight (Walters) |
| `b-atg-blocked` | Bump-and-Run |
| `b-atg-game` | Up-and-Down Scoring Game |

Variant B (approach focus):
| Block | Drill |
|---|---|
| `b-app-warmup` | 7-Iron Calibration Warm-Up |
| `b-app-blocked` | Tee-Gate Center-Face |
| `b-atg-updown-game` | Up-and-Down Scoring Game *(intentional reuse)* |
| `b-ott-warmup` | Rhythm/Tempo Warm-Up |
| `b-ott-blocked` | Path Gate |
| `b-put-lag-game` | Distance Ladder |

## Tests / typecheck / drift

- `pnpm --filter @oga/core test` — 490 pass (was 489, +1 for the drill_name assertion).
- `pnpm typecheck` — green across `@oga/core`, `@oga/supabase`, `@oga/web`.
- `node scripts/check-shared-vendor.mjs` — OK (9 files).

## Deploy

After merge (or immediately for dev validation):
```
npx supabase functions deploy generate-practice-plan --use-api \
  --project-ref txquvfeyvkaetqlamqlz
```

## QA

On dev, force a baseline fallback (temporarily set `MIN_SG_ROUNDS=999` or use a fresh user with <5 SG rounds) → generate a plan → confirm every block has a real drill name + the expand toggle works.

### DIFF
--- a/packages/core/src/practice-plan/baseline.test.ts
+++ b/packages/core/src/practice-plan/baseline.test.ts
@@ -64,7 +64,6 @@ describe('selectBaselinePlan', () => {
           for (const b of s.blocks) {
             expect(b.drill_name, `block ${b.id} has no drill_name`).toBeTruthy()
             expect(typeof b.drill_name).toBe('string')
-            expect(b.drill_name!.length).toBeGreaterThan(0)
           }
         }
       }
--- a/packages/core/src/practice-plan/baseline.test.ts
+++ b/packages/core/src/practice-plan/baseline.test.ts
@@ -52,4 +52,22 @@ describe('selectBaselinePlan', () => {
       }
     }
   })
+
+  it('every block carries a drill_name that serveBaseline can resolve', () => {
+    // Each baseline block names a real corpus drill by name; the Edge fn
+    // queries `drills.name in (…)` and fills `drill_id` before INSERT.
+    // A missing/empty hint means the UI would show "Drill" and `canExpand`
+    // would be false — the regression D fixes.
+    for (const variants of Object.values(BASELINE_PLANS)) {
+      for (const v of variants ?? []) {
+        for (const s of v.sessions) {
+          for (const b of s.blocks) {
+            expect(b.drill_name, `block ${b.id} has no drill_name`).toBeTruthy()
+            expect(typeof b.drill_name).toBe('string')
+            expect(b.drill_name!.length).toBeGreaterThan(0)
+          }
+        }
+      }
+    }
+  })
 })
--- a/packages/core/src/practice-plan/baseline.ts
+++ b/packages/core/src/practice-plan/baseline.ts
@@ -2,8 +2,10 @@ import type { SkillLevel, Goal } from '../constants'
 import type { PlanDraft, PlanCategory } from './types'
 
 // Baseline plans are static content (same `PlanDraft` shape the UI renders).
-// NOTE: block `drill_ref` is a placeholder here — baseline drill references are
-// resolved to real drill ids when the baseline content is authored (Phase B).
+// Block `drill_name` is the canonical corpus name; `serveBaseline` resolves
+// each one to a real `drill_id` at storage time and strips the hint before
+// INSERT. `drill_ref` stays 0 (placeholder — baselines don't go through the
+// retrieval/Claude path that consumes it).
 type BaselineVariant = Omit<PlanDraft, 'sessions'> & {
   primaryCategory: PlanCategory
   sessions: PlanDraft['sessions']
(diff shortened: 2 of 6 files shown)


cner-smith (commented):
## Review — feat(plan): resolve baseline drill_id by corpus name (#18 Track D)

**Concerns**
- `supabase/functions/generate-practice-plan/index.ts` (the `nameToId` map build) — the map has last-write-wins semantics: if the `drills` table ever has two rows with the same name, the resolved `drill_id` is whichever row the DB returned last. CONTRIBUTING.md's drill schema only enforces `id uuid primary key`; there is no unique constraint on `name`. The 11 hardcoded names are verified against `seed.sql` today, but nothing prevents a future contributor from adding a drill that collides with one of these names and silently changes which ID gets resolved. A `UNIQUE` constraint on `drills.name` in a follow-up migration, or resolving via a stable `slug` or `corpus_key` column that's explicitly unique, would close this. Low urgency now (small corpus, hardcoded names), but worth tracking before the corpus grows.

**Nits**
- `packages/core/src/practice-plan/baseline.test.ts` (inner-loop assertions) — `expect(b.drill_name!.length).toBeGreaterThan(0)` is redundant: the `.toBeTruthy()` two lines above already fails on empty strings (empty string is falsy). The `!` non-null assertion is also unnecessary in the same position for the same reason. The line can be dropped without weakening the test.

---
Looked good: targets `dev`, `drill_name` field and the new test both live in `@oga/core` as CONTRIBUTING.md requires for new exports, `resolvedSessions` is built before the `dryRun`/transient-ephemeral/INSERT branch so all three exit paths carry the resolved IDs.

---

cner-smith:
Thanks for the review.

**Concern (drills.name collision)** — already closed. Migration `0034_drill_corpus_v2_data.sql:27` adds `alter table public.drills add constraint drills_name_unique unique (name);`, so a future contributor can't silently change which `drill_id` resolves — the second-insert attempt would 23505. The migration's own header even calls it out:

> Idempotency: adds a UNIQUE constraint on drills.name (which the data relies on).

Prod hasn't received 0034 yet (it's in the migrations bucket in `project_baseline_drill_mapping_and_v1_remaining`), but the graceful fallback handles that — pre-corpus prod, the lookup returns 0 rows, no name maps to an id, and every block stores with `drill_id` undefined (same as today's behavior). Once 0034 ships to prod alongside the next release, the constraint goes live there too.

A `slug` / `corpus_key` column would still be more robust against a future rename — the migration would preserve the slug while the name could be edited — but that's a corpus-evolution concern, not a Track D blocker. Filing as a follow-up if/when a rename actually comes up.

**Nit (redundant length check)** — fixed in `eb3758b`.

PR #450, opened 2026-05-29
release: dev → main (v1.0-rc) — AI practice plan engine + Track 3 UI + share cards + email scaffold

cner-smith:
**Promote dev → main for v1.0-rc.** 25 merged PRs, 6 migrations (0029–0034), 2 new Edge Functions. This is the biggest dev cycle since the v0.6.0 backmerge (#425) and brings v1.0's largest unshipped feature — the AI practice plan — to production-ready on dev.

> **⚠️ This release is the codebase only.** Prod-side rollout (apply migrations 0029–0034 to prod DB, deploy `round-summary-email` v5 + `generate-practice-plan` v7 to prod with `--use-api`, set `ANTHROPIC_API_KEY` + `PLAN_MODEL` on prod, backfill `summary_email_sent_at` before scheduling the email cron) is **tracked separately and not done by merging this PR**. See the post-merge prod runbook at the bottom.

**Closes:** #18, #201, #374

Email cron go-live (#317) and the mobile Practice screen are **intentionally** NOT closed — scaffold + Edge fn shipped but the cron isn't scheduled in prod; mobile parity is TBD.

---

### 🧠 AI practice plan engine (#18 — the headline)

**Engine Phase A** — pure `@oga/core` helpers
- #434 `feature/ai-plan-engine-core-18` — 5 helpers (`playFrequencyPlan`, `resolveTarget`, `buildPlayerDigest`, `validatePlanDraft`, `selectBaselinePlan`), 438 tests

**Engine Phase B** — Edge Function `generate-practice-plan`
- #435 `feature/ai-plan-engine-phase-b-18` — Edge fn + vendored `_shared/practice-plan/` + CI drift-check + migrations 0031 (cadence unique) + 0032 (global plan count)
- #436 `fix/edge-vendor-ts-extensions` — Deno needs `.ts` extensions in vendored imports
- #437 `fix/plan-tool-schema-keywords` — drop `minimum` / `strict` (Anthropic tool API 400)
- #438 `fix/derive-session-minutes` — server-derive `total_minutes`, drop ±20% check
- #440 `fix/edge-cors` — CORS so the browser invoke doesn't preflight-block
- #443 `feature/practice-mode-engine` — engine speaks the mode taxonomy + every session closes on the green
- #446 `feature/practice-loading-and-timeout-fix` — generation progress bar + treat `APIConnectionError` as transient
[...]

PR #451, opened 2026-05-29
chore(edge): explicit verify_jwt = true for round-summary-email (#450 review)

cner-smith:
PR #450 review follow-up.

## Why

`round-summary-email`'s [`isServiceRole`](supabase/functions/round-summary-email/index.ts) decoder reads `role` from the JWT payload **without re-checking the signature** — it trusts that the Supabase gateway has already validated it. The comment at `index.ts:40–46` explicitly acknowledges this.

That trust is sound while `verify_jwt = true`, but the function had no entry in `supabase/config.toml` — it relied on the platform default. A future upstream change to that default would silently let unsigned tokens reach the handler.

`generate-practice-plan` already declares `verify_jwt = true` explicitly (`config.toml:62–63`) with a comment explaining exactly why it matters. This PR mirrors that for `round-summary-email`.

## Change

Adds a new block to `supabase/config.toml`:

\`\`\`toml
[functions.round-summary-email]
verify_jwt = true
\`\`\`

Plus a 5-line comment explaining why the explicit declaration is load-bearing for the cron-only service-role guard.

## Behavior

**No-op today** — matches the current Supabase default. The value is making the dependency visible to maintainers and immune to platform defaults shifting.

## Once merged

PR #450 (dev → main release) picks this up via its `dev` head — nothing to re-trigger there manually.

### DIFF
--- a/supabase/config.toml
+++ b/supabase/config.toml
@@ -61,3 +61,11 @@ enabled = false
 # the dryRun service-role gate inside the function is only safe while this is on.
 [functions.generate-practice-plan]
 verify_jwt = true
+
+# round-summary-email is cron-only. Its `isServiceRole` decoder reads `role`
+# from the JWT payload without re-checking the signature — it trusts the
+# Supabase gateway has already validated it. Make that dependency explicit
+# (instead of relying on the platform default) so a future upstream change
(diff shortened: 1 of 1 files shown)


cner-smith:

---

PR #453, opened 2026-05-29
feat(db): migration 0035 — v1 corpus drills (40 missing rows on prod)

cner-smith:
## Why

The v1 drill corpus (40 rows from PR #432) only ever lived in `supabase/seed.sql` — dev-only via `supabase db reset`. Migration **0034 ported only the v2 fill (85 rows)** on the assumption the 40 v1 rows were already present (true on dev, false on prod).

Result on prod today: **drills count is 109** = 85 from 0034 + 24 pre-existing v0 placeholder drills, missing **all 40 v1 entries**.

Track D's `serveBaseline` resolves `drill_id` by corpus name. Five of the 11 hardcoded baseline picks are v1 drills:

- **Gate Drill**
- **Up-and-Down Scoring Game**
- **Rhythm/Tempo Warm-Up**
- **Path Gate**
- **Distance Ladder**

On prod those name queries returned 0 rows and the D graceful fallback rendered the literal "Drill" with no expand. The other 11 v1 drills not used by baselines were equally invisible to Claude's candidate-pool retrieval (smaller pool, less variety).

## Change

`supabase/migrations/0035_v1_corpus_drills.sql` — 69 lines, ~111KB. Single `insert into public.drills (…) values …  on conflict (name) do nothing;` with 40 row tuples captured verbatim from dev on 2026-05-29.

Idempotent via `drills_name_unique` (added in 0034). **No-op on dev** (all 40 already present). On prod, inserts the 40 missing rows.

## What this does NOT touch

24 pre-existing v0 placeholder drills on prod (lowercase / generic names like `Gate drill`, `Lag putting clock`, `Sand-shot 10s`) are left alone. Deleting them risks orphaning practice_plans JSONB references that may cite those uuids. Cleanup is deferred to a separate migration once we've confirmed no live plans depend on them.

## After deploy

- prod drills: **109 → 149** (still 24 v0 pollution left for a follow-up)
- All 11 baseline picks resolve cleanly — baseline practice plans render real drill names with the **+** expand
- Claude's candidate pool gains 40 curated v1 drills

## Verification on the diff

- 40 row tuples (`grep -c '^(\$drill\$' = 40`)
- Ends with `on conflict (name) do nothing;`
- Five baseline-needed names all present
- Sourced from dev's current `public.drills` via PostgreSQL `format(… %L …)` — escaping is guaranteed by construction

## Deploy

After merge:
```
git checkout dev && git pull
npx supabase db push --linked   # prod-linked; applies 0035
```

Then regenerate a baseline plan in prod (signed in as a user with no SG rounds, or clear the cadence guard) and confirm all six baseline blocks render real drill names.


cner-smith (commented):
## Review — feat(db): migration 0035 — v1 corpus drills (40 missing rows on prod)

**Concerns**
- `supabase/migrations/0035_v1_corpus_drills.sql:67` — `Upside-Down Whoosh` sets `facility = '{}'`. CONTRIBUTING.md enumerates `anywhere` as a valid facility value (`['range','short_game','putting','sim','anywhere']`), and the drill's own instructions are explicit: *"No balls, no tee, no range — a living room or hotel corridor works."* An empty array won't satisfy any `&&`-style facility filter the planner uses, so this drill will never be retrieved despite being in the corpus. Should be `'{anywhere}'`.

**Nits**
- Throughout the file, `goals` uses the quoted array-string form (`'{break_80,...}'`) while `targets` uses the unquoted form (`{chip_contact,...}`). Both are accepted by PostgreSQL when the column type is inferable from the INSERT target, but the inconsistency could trip up a contributor hand-authoring a future row. If the rows were generated with `format('%L', …)` as the description states, the two columns appear to have been extracted differently — worth checking the extraction query for the next corpus migration.

---
Looked good: `on conflict (name) do nothing` idempotency is correct and relies on the right constraint (`drills_name_unique` from 0034); `public.drills` is a shared read-only corpus, not user-owned data, so no RLS concern; migration number is sequential; PR targets `dev`; single concern.

---

PR #454, opened 2026-05-29
fix(db): quote targets array in 0035 inserts

cner-smith:
Followup to #453. The targets column in 0035 was emitted with %s instead of %L in the format() call, producing unquoted array literals like {a,b,c} that hit a syntax error on apply.

## Caught by

Apply attempt against prod returned ERROR: 42601: syntax error at or near "{" at the first targets value. Full transaction rolled back; prod unchanged at 109 drills.

## Fix

Regenerated 0035 with the corrected format() string — position 9 is now %L and the argument is targets::text without coalesce. All other content is byte-identical.

## Already applied to prod via MCP

After this patch, the corrected migration was applied to prod (znjmkptblswxststjxwc) via mcp apply_migration. Drill count went 109 → 149. The 5 baseline-needed v1 drills (Gate Drill, Up-and-Down Scoring Game, Rhythm/Tempo Warm-Up, Path Gate, Distance Ladder) now resolve.

This PR is the source-of-truth backfill for dev so the file stops diverging from what landed on prod.


cner-smith:

---
Looked good: `%L` fix matches the stated 42601 error exactly; `ON CONFLICT (name) DO NOTHING` makes a double-apply safe (covers both the prod-first apply described in the PR body and any dev env that already has these rows via seed); diff is +40/−40 one-line-per-drill at the `targets` position, consistent with "all other content byte-identical"; targets `dev` not `main`; single concern; no secrets, no new tables or RLS changes, no `@oga/core` or mobile deps touched.

---

Issue #469, opened 2026-05-30
Mobile Learn: tapping a not-yet-written article shows a 'check back soon' stub


cner-smith:
Found during on-device QA (2026-05-30).

**Symptom:** On mobile, tapping some Learn article links lands on a stub — "This guide is being written. Check back soon." — instead of content.

**Mechanism (not a crash, working as coded):** `apps/mobile/app/(app)/learn/[article].tsx` renders `<StubBody>` when `findLearnArticle(slug).article.status === 'soon'` (also the `default` case of `LiveArticle` for any id missing from its switch). Seven articles ARE fully written and render: strokes-gained, benchmarks, glossary, how-to-practice, course-management, mental-game, skill-games-pressure-games. The remaining catalog entries are `soon`/draft placeholders that the Learn **index still links to**, so they feel "not wired up."

**Decision needed:** either (a) hide `status:'soon'` entries from the mobile Learn index so we never link to a stub, or (b) finish/flip those articles. Mirror whatever the web Learn index does. Per CLAUDE.md the catalog is intentionally mostly draft pre-launch, so (a) is likely the quick win.

Scope: mobile only (`apps/mobile/app/(app)/learn/index.tsx` + `[article].tsx`). Low priority / content-UX gap.

cner-smith:
Resolved — the stub is unreachable. Verified against current code:

- All 20 catalog articles are `status: 'published'` (no `'soon'`), so the `status === 'soon'` → StubBody branch in `apps/mobile/app/(app)/learn/[article].tsx:30` never fires.
- `MOBILE_ARTICLES` registers a component for every published article id — 20 ids ↔ 20 components, exact match — so the `LiveArticle` → StubBody fallback (`[article].tsx:47`) never fires either.

#513 published the full set, so 'tapping a not-yet-written article' can't happen. The `StubBody` stays as graceful handling if a future `'soon'` article is ever added. Closing as completed.

PR #479, opened 2026-05-31
feat(mobile): Learn parity — port all 20 articles to native

cner-smith:
Ports the 13 missing Learn articles to mobile so **all 20 render** (was 7). Per-article RN components under `apps/mobile/components/learn/articles/` dispatched by id via a registry; shared primitives in `components/learn/primitives.tsx` replace the per-article boilerplate the web articles hand-roll; inline web `<svg>` figures re-authored in `react-native-svg` with matching viewBoxes. `[article].tsx` imports the shared styles and routes registry articles after the existing inline 7. Addresses #469.

_Split from the original combined PR per CONTRIBUTING "one concern per PR" — the crawler work is now #480, the live-round map fixes #481._

## On the flagged "unused" primitives
`H4` / `NumberList` / `Kv` / `ResourceList` are exported but not yet consumed by the 13 ported articles. **Kept intentionally:** the planned article unification (migrate the inline 7 → these primitives) will use them, so removing now would just be churn re-adding identical code shortly.

## QA
Mobile typecheck green; on-device render QA of the 20 articles pending.

### DIFF
--- a/apps/mobile/app/(app)/learn/[article].tsx
+++ b/apps/mobile/app/(app)/learn/[article].tsx
@@ -2,37 +2,8 @@ import { Linking, Pressable, ScrollView, Text, View } from 'react-native'
 import { useLocalSearchParams, useRouter } from 'expo-router'
 import { findLearnArticle } from '@oga/core'
 import { AppBar } from '../../../components/ui/AppBar'
-
-const KICKER: import('react-native').TextStyle = {
-  color: '#8A8B7E',
-  fontSize: 10,
-  fontWeight: '500',
-  letterSpacing: 1.4,
-  textTransform: 'uppercase',
-}
-
-const TITLE: import('react-native').TextStyle = {
-  color: '#1C211C',
-  fontSize: 26,
-  fontStyle: 'italic',
-  fontWeight: '500',
-  lineHeight: 32,
-  marginBottom: 14,
-}
-
-const BODY: import('react-native').TextStyle = {
-  color: '#1C211C',
-  fontSize: 15,
-  lineHeight: 22,
-  marginBottom: 14,
-}
-
-const SUBKICKER: import('react-native').TextStyle = {
-  ...KICKER,
-  color: '#5C6356',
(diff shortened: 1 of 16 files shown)


cner-smith:
## Review — Mobile Learn parity (all 20 articles) + OSM hole-geometry crawler + enrich budget

**Blockers**
- PR as a whole — CONTRIBUTING.md §"One concern per PR": "UI restyle, bug fix, and new feature do not belong in the same PR. PRs that add scope beyond their stated concern will be asked to split." Three independent concerns are bundled here: the mobile Learn article port (16 files under `apps/mobile/`), the OSM hole-geometry crawler pass (`scripts/crawl/holes-fetcher.ts`, new db-writer functions, new types), and the enrich budget cap + daily cron (`scripts/crawl/enricher.ts`, `scripts/enrich-prod-daily.sh`). The author offers to split — please do. The two crawler concerns are close enough to share a PR; the mobile UI articles should be a separate one.

**Concerns**
- `apps/mobile/components/learn/primitives.tsx` — `H4`, `NumberList`, `Kv`, and `ResourceList` are exported but not imported by any of the 13 article files added in this PR. CONTRIBUTING.md: "Default to no. Ship the minimum that satisfies the requirement, then stop." If no current article uses them, they should not ship yet; add them when the first article that needs them lands.

- `scripts/crawl/util.ts` — `haversineMeters` is pure geographic math. The crawler writes the result (yards, tee/pin coords) into the `holes` table, and if an app surface ever needs client-side distance (nearest-course finder, hole layout), it will have to re-implement it. Worth lifting to `@oga/core` in a follow-up.

---
Looked good: no secrets hardcoded (cron script sources from gitignored `.env.crawl.prod.local`), no new migrations without RLS, no `any` added, no `eslint-disable` without comment, `nativewind` unchanged, PR targets `dev`, `budgetReached` correctly breaks the outer state loop, `primitives.tsx` callers across the 13 new articles clear the 3-caller bar for every primitive that is actually used.

---

Issue #470, opened 2026-05-30
Course data is low-quality on prod + dev — wipe and re-import


cner-smith:
The courses table on both prod and dev has degraded — stale, duplicate, and malformed rows that hurt search and round setup. We should wipe the course data clean and re-import from scratch via the crawler (scripts/crawl-courses.ts, OpenGolfAPI/OSM).

Needs a plan, not a blind truncate:
- rounds / hole_scores / shots FK to courses. Re-import must preserve existing user rounds (remap to re-imported course ids, or keep course rows referenced by rounds). Cannot simply TRUNCATE courses.
- Crawler upserts on external_id (migration 0019), so a clean re-crawl should converge — but the rotten rows that predate external_id / were hand/synthetic-created need an identification + cleanup pass first.
- Synthetic holes: courses materialized mid-round (ensureRealHole) are user-specific; decide whether they survive.
- Run prod and dev as separate, sequenced operations. Dev first to validate the wipe+reimport flow end to end.

Acceptance: search returns clean, de-duplicated, correctly-located courses on both environments; existing rounds still resolve to a valid course.

cner-smith:
### Triage (prod data, 2026-06-08) — the "wipe + re-import" premise doesn't hold

Measured prod (`courses` = 15,535 rows):

| Metric | Count | Note |
|---|---|---|
| dup **external_id** groups | **0** | upsert-on-external_id (0019) works — a re-import converges to the *same* rows and removes **zero** dups |
| **true** duplicate rows | **238** (229 clusters) | same normalized name **and** same ~1km location; 1.5% of the table |
| same-name groups total | 1,239 | but ~1,010 are legit different courses sharing generic names ("Municipal GC" in different towns) — **not** dups |
| missing coords | 115 (0.7%) | unmappable/unsearchable |
| no external_id | 4 | legacy hand/synthetic rows |
| missing city/state | 5,282 (34%) | completeness gap (OSM sources lack it), not rot — backfillable via reverse-geocode |
| missing name | 0 | — |
| **courses referenced by rounds** | **1** | the "preserve user rounds across re-import" risk is a single FK |

**Conclusion: don't wipe + re-import.** A clean re-crawl would reproduce the same rows (external_id dedup already converges), so it wouldn't fix anything — and it carries needless risk. The real rot is small and specific.

**Reframed plan (no build, mostly SQL, dev-first):**
[...]

Issue #267, opened 2026-05-09
Crawler: handle multi-course facilities with duplicate hole numbers


cner-smith:
Multi-course facilities (36-hole, 54-hole clubs) have duplicate hole ref numbers 1-18 in OSM. The crawler currently deduplicates blindly, mixing holes from different courses.\n\nOSM typically tags hole ways with a name that distinguishes courses (e.g. 'Earlywine North' vs 'Earlywine South').\n\nFix: in the crawler's dedup step, if duplicate hole numbers exist, check if OSM name tags distinguish them. If so, create separate course records per named group rather than deduplicating.\n\nThe manual --course-filter flag on import-osm-course.ts is the workaround for now.

cner-smith:
## Notes

### Real-world examples
- **Earlywine Park GC, OK** (the example in the body): 36-hole facility, North + South courses, both with holes ref 1-18
- **Cherokee Hills GC, OK**: split with East/West
- **Bandon Dunes Resort, OR**: 6 separate 18s (Bandon Dunes, Pacific Dunes, Old Macdonald, Bandon Trails, Sheep Ranch, Punchbowl)
- **Pebble Beach area**: Pebble Beach + Spyglass Hill + Spanish Bay + Bayonet/Black Horse all in close proximity

This is far more common than the issue body suggests — most "club" facilities have multiple courses. The crawler is currently mis-merging at probably hundreds of US facilities.

### OSM tag patterns to detect
- `name=Earlywine Park GC North` (most common)
- `course=North` (less common but cleanest)
- `name:full=...` or `loc_name=...` (rare)
- `network` or `operator` tags identifying the parent facility

### Dedup algorithm
```ts
// pseudo:
function dedupHoles(holes: OsmHole[]): Course[] {
  const byRef = groupBy(holes, h => h.tags.ref)
  // If any ref has multiple holes with distinguishing name tags, split:
  for (const [ref, group] of byRef) {
    if (group.length > 1) {
      const distinctNames = new Set(group.map(h => h.tags.name ?? h.tags.course))
      if (distinctNames.size > 1) {
        // It's a multi-course facility — split into separate courses by name
        return splitByName(holes)
      }
    }
  }
  // Otherwise, single-course path (current behavior)
  return [{ holes }]
}
```

### Validation
Spot-check after re-crawl:
- Earlywine should have 2 course rows, each with 18 holes
- Bandon should have 6 (or 5; Punchbowl is a 19-hole putting course, technically not 18)
[...]

PR #480, opened 2026-05-31
feat(crawler): OSM hole-geometry extraction + enrich daily-budget cron

cner-smith:
Two related course-data crawler improvements, split out of #479 per CONTRIBUTING "one concern per PR".

## OSM hole geometry (`--source osm-holes`)
The crawler was centroid-only — it never fetched the `golf=hole`/`tee`/`green` geometry OSM actually carries, so only ~0.1% of hole rows had coordinates. New per-state Overpass `out geom` pass assigns each hole to its nearest course centroid (≤1500m), orients tee↔green (snapping to explicit tee/green features), derives par (clamped to the 3–6 CHECK) + yards, and writes `tee_lat/lng` + `pin_lat/lng`. Skips courses that already have geometry — idempotent + resumable via `crawl_state`.

New `scripts/crawl/holes-fetcher.ts`; `upsertHoleGeometry` / `fetchOsmCoursesGeoForState` / `fetchCoursesWithHoleGeometry` in db-writer; `haversineMeters` in util; types.

**Validated on dev** (Lincoln Park West 0→18 holes with coords) and **run on prod: 8,885 courses gained geometry, 125,928 holes, 0 errors.** Relates to #470 / #267.

## Enrich rate-limit budget + daily cron
`--max-courses N` global budget so a daily `enrich` run stops under the OpenGolfAPI ~900/day cap (state left in-progress to resume via the already-teed skip). `scripts/enrich-prod-daily.sh` is a portable crontab wrapper (sources gitignored `.env.crawl.prod.local`).

## Follow-up (review note)
Lift `haversineMeters` to `@oga/core` if an app surface ever needs client-side distance — deferred.


cner-smith (commented):
## Review — feat(crawler): OSM hole-geometry extraction + enrich daily-budget cron

**Concerns**
- `scripts/crawl-courses.ts` / `scripts/crawl/enricher.ts` / `scripts/enrich-prod-daily.sh` — This PR contains two independently deployable changes: the `osm-holes` geometry pass and the `--max-courses` enrich budget with cron wrapper. CONTRIBUTING.md: "One concern per PR." The author notes this was already split from #479, but these two features touch disjoint code paths and could be reverted or bisected independently. Worth splitting so the OSM geometry work (which is the heavier lift) can be reviewed and rolled back on its own.

**Nits**
- `scripts/crawl/db-writer.ts` (`fetchOsmCoursesGeoForState`) — The name implies OSM-sourced courses, but the function fetches all courses with lat/lng regardless of source. The comment immediately above it clarifies this ("Includes ALL coord courses (OSM, OpenGolfAPI-enriched, and manually added), not just osm_%."), but the name will mislead a caller scanning the export list. `fetchCourseGeoForState` matches the actual contract.
- `scripts/enrich-prod-daily.sh:38–41` — `STAMP` is captured once at script start and reused in both the opening and closing log lines. The "finished" entry will always show the start time, not the completion time. `$(date ...)` inline on line 41 fixes it.
- `scripts/crawl/holes-fetcher.ts` (`fetchHoleFeaturesInState`) — On the fourth and final Overpass attempt (`attempt === BACKOFFS_MS.length`), when every endpoint in that round fails, the loop `break`s without printing a warning. The thrown error reaches the outer `catch` and logs `fatal:`, but there is no log line between the last endpoint failure and the throw — so you can't tell from the log that a final retry even happened. Printing the warning unconditionally before checking `backoff == null` would make the failure sequence visible.

---
Looked good: `tee_lat/lng/pin_lat/lng` are already defined in `0001_initial_schema.sql`, so no migration gap; idempotency guard via `fetchCoursesWithHoleGeometry` is solid; the `budgetReached` state machine correctly leaves the current state `in_progress` and breaks the outer loop; par clamping comment matches the `CHECK (par between 3 and 6)` in the schema.

---

PR #481, opened 2026-05-31
fix(mobile): live-round map fixes — AimGhosts, first-ball, drag hit target

cner-smith:
Three live-round map fixes (split out of #479 per "one concern per PR").

## 1. AimGhosts not rendering (regression)
`4670661` gated ghost promotion on the map phase, but committed shots and Re-place-ball both collapse to `PLACE_BALL`, so the promote branch was dead. Now keys off an explicit `aimCommitted` signal (`roundState` SHOT_DETAIL/PUTTING) AND the live aim surviving the transition — committed shot → one ghost, skip-aim / re-place → none (no duplicate).

## 2. First ball seeded on the tee box
Shot 1 with no ball yet defaults the ball marker to the hole's tee coords; GPS-follow still refines it.

## 3. Aim/ball drag hit target too small
`@rnmapbox` `PointAnnotation` hit-tests the **opaque pixels of the rendered bitmap**, not the View frame — so the 80pt transparent hit-pad did nothing and only the ~9px dot was grabbable (touches missed → map panned). Replaced with a visible translucent grab disc (aim 48pt, ball 44pt) so the hittable area is ~5× larger. Interim — proper rewrite tracked separately.

## QA
Mobile typecheck green; on-device validation in progress.

### DIFF
--- a/apps/mobile/components/round/HoleMap.tsx
+++ b/apps/mobile/components/round/HoleMap.tsx
@@ -458,12 +458,30 @@ export function HoleMap({
     },
     [onSetAim],
   )
-  // Static child — memoized so the PureComponent never re-bitmaps it. 80pt
-  // transparent hit area around a small center grab-dot.
+  // Static child — memoized so the PureComponent never re-bitmaps it.
+  // @rnmapbox PointAnnotation hit-tests the OPAQUE pixels of the rendered
+  // bitmap, NOT the React View frame — so a fully transparent pad is not
+  // grabbable and only the tiny dot was draggable. A visible translucent 48pt
+  // disc gives a real ~48pt opaque touch target (and shows the grab zone). The
+  // outer 80pt box just centers it. Proper fix is a MarkerView + gesture-handler
+  // rewrite — see the tracking issue.
   const aimHandle = useMemo(
     () => (
       <View style={{ width: 80, height: 80, alignItems: 'center', justifyContent: 'center' }}>
-        <Marker color="#A66A1F" border="#FBF8F1" size={9} />
+        <View
+          style={{
+            width: 48,
+            height: 48,
+            borderRadius: 24,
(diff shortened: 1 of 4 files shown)


cner-smith (commented):
## Review — fix(mobile): restore AimGhosts on committed shots + seed first ball on tee

**Concerns**
- CONTRIBUTING.md "One concern per PR": the AimGhost promotion fix and the tee-seed default are independent bugs with separate mechanisms. The ghost fix spans `AimGhost.tsx` / `HoleMap.tsx` / `LiveRoundSession.tsx`; the tee-seed spans `useHoleState.ts:327` / `LiveRoundSession.tsx`. The description acknowledges two fixes but argues proximity. Strictly, two bugs are two concerns — CONTRIBUTING.md's one-concern rule doesn't carve out an exception for same-area fixes. Whether a further split is worth the overhead is the maintainer's call.

**Nits**
- `apps/mobile/components/round/markers/AimGhost.tsx:15–22` — The `aimCommitted` interface comment re-narrates the implementation in detail (phase enum collapse, three-way discriminator table, effect-ordering rationale). That same walkthrough already appears verbatim in the phase-exit effect comment at lines 68–79. CONTRIBUTING.md: "Don't narrate what the code does." A single sentence naming the phase-collapse problem is sufficient at the interface level; the effect comment is the right home for the full explanation.

---
Looked good: targets `dev`, no package-boundary violations, no new `@oga/core` exports, `aimCommitted`+`aimRef.current` guard correctly handles skip-aim without an async-reset race, `lastAimSnapshotRef` unconditionally cleared on every phase exit.

---

cner-smith:
On-course validation (2026-06-01): the translucent grab-disc made the aim + ball markers easy to drag during a real round — the interim hit-target fix works. So #482 (the MarkerView + gesture-handler rewrite) is now lower-priority/optional rather than urgent, unless we want the fully bulletproof target later.

PR #489, opened 2026-06-01
feat(mobile): in-app account deletion (#306, #314)

### LINKED ISSUE #306: iOS: implement in-app account deletion flow (App Store Guideline 5.1.1v mandate)
**Severity:** blocker for App Store submission (mandatory since June 30, 2022)
**Audit source:** certification-planner agent, 2026-05-13

## Background

Apple Guideline 5.1.1(v): apps that support account creation must allow account deletion **inside the app**. Cannot be email/support-only. Cannot be web-only. Must permanently delete the account and associated data.

Currently OGA has Sign Out but no Delete Account flow. This is a guaranteed rejection until added.

## Required components

1. **UI entry point.** Settings (Profile page) → \"Delete Account\" button. Reachable in one or two taps from any screen the user is in.
2. **Confirmation dialog** with clear warning copy.
3. **Backend deletion endpoint.** Supabase RPC that wipes the user's rounds/shots/profile + calls \`auth.admin.deleteUser\` (requires service role — must run as RPC, not from client).
4. **Logout + redirect** to login screen after deletion succeeds.

### Settings entry

> Profile → Account section
> \"Delete Account\" (red text)

### Confirmation dialog

> **Title:** Delete your OGA account?
>
> **Message:** This will permanently delete your account and all rounds, shots, and saved data. This cannot be undone.
>
> **Cancel button:** Cancel
> **Confirm button:** Delete Account (destructive red)

### Post-deletion message

> Toast or alert: \"Your account has been deleted.\"

## Backend (Supabase)

Create a SECURITY DEFINER RPC named \`delete_my_account\`. It must:

1. Take no parameters (uses \`auth.uid()\`)
2. Inside a transaction:
   - DELETE FROM shots WHERE user_id = auth.uid()
   - DELETE FROM hole_scores WHERE round_id IN (SELECT id FROM rounds WHERE user_id = auth.uid())
   - DELETE FROM rounds WHERE user_id = auth.uid()
   - DELETE FROM user_clubs WHERE user_id = auth.uid()
   - DELETE FROM practice_plans WHERE user_id = auth.uid()
   - DELETE FROM profiles WHERE id = auth.uid()
3. Call \`auth.admin.delete_user(auth.uid())\` — requires the service role, hence SECURITY DEFINER

Migration goes in \`supabase/migrations/0025_account_deletion_rpc.sql\` (or whatever the next number is).

## Files

- New: Settings page Delete Account UI (likely \`apps/mobile/app/(app)/profile.tsx\` or settings sub-page)
- New: \`supabase/migrations/0025_account_deletion_rpc.sql\`
- Update: web equivalent if web also has account management (filed separately if needed)

## Notes

Apple specifically watches for token / data residue after deletion. Test by:
1. Delete account in iOS app
2. Try to log in with same email → should fail
3. Try to log in with same email and reset password → should fail (no auth.users row)
4. Try to create a new account with the same email → should succeed

### LINKED ISSUE #314: iOS: Supabase RPC backend for account deletion (supports App Store mandate)
**Severity:** blocker (paired with the in-app account deletion flow — backend half of #306)
**Audit source:** certification-planner agent, 2026-05-13

## Scope

Migration that adds a SECURITY DEFINER Postgres function which:

1. Wipes all user-owned data across our tables (rounds, shots, hole_scores, user_clubs, practice_plans, profile)
2. Calls \`auth.admin.delete_user(auth.uid())\` to remove the auth.users row
3. Returns success/failure to the client

## Suggested migration

\`supabase/migrations/0025_account_deletion_rpc.sql\`:

```sql
-- Account deletion RPC. Called by the mobile app's Delete Account
-- button. Wipes all user-owned data in a single transaction and then
-- removes the auth.users row via auth.admin.delete_user.
--
-- SECURITY DEFINER so the function runs with elevated privileges
-- (needed for auth.admin.* and for cross-table deletes that might
-- be RLS-restricted at the user level). The inner uid check ensures
-- callers can only delete their own account.

CREATE OR REPLACE FUNCTION public.delete_my_account()
RETURNS void
LANGUAGE plpgsql
SECURITY DEFINER
SET search_path = public
AS \$\$
DECLARE
  uid uuid;
BEGIN
  uid := auth.uid();
  IF uid IS NULL THEN
    RAISE EXCEPTION 'not authenticated';
  END IF;

  -- Delete in FK-safe order (children → parents).
  DELETE FROM shots WHERE hole_score_id IN (
    SELECT hs.id FROM hole_scores hs
    JOIN rounds r ON r.id = hs.round_id
    WHERE r.user_id = uid
  );
  DELETE FROM hole_scores WHERE round_id IN (
    SELECT id FROM rounds WHERE user_id = uid
  );
  DELETE FROM rounds WHERE user_id = uid;
  DELETE FROM user_clubs WHERE user_id = uid;
  DELETE FROM practice_plans WHERE user_id = uid;
  DELETE FROM profiles WHERE id = uid;

  -- Final step: remove the auth.users row.
  PERFORM auth.admin_delete_user(uid);
END;
\$\$;

GRANT EXECUTE ON FUNCTION public.delete_my_account TO authenticated;
```

Note: \`auth.admin.delete_user\` may not be directly callable from within a Postgres function. If that's the case, the alternative is to leave the auth.users row, mark the email as deleted via a Supabase Edge Function that runs after the RPC. Verify which approach works against the current Supabase version before finalizing.

## Client integration

In \`apps/mobile/lib/supabase.ts\` or new auth helpers:

```ts
async function deleteMyAccount() {
  const { error } = await supabase.rpc('delete_my_account')
  if (error) throw error
  await supabase.auth.signOut()
}
```

## Files

- New: \`supabase/migrations/0025_account_deletion_rpc.sql\` (or next sequential number)
- Update: \`apps/mobile/lib/\` — add deleteMyAccount helper
- Hooks into: settings/profile UI from #306

## Test plan

1. Create test account, add some rounds/shots
2. Call \`delete_my_account()\`
3. Verify: no rows in shots/hole_scores/rounds/profiles/user_clubs/practice_plans for that user_id
4. Verify: auth.users row is gone (or marked deleted)
5. Verify: signing in with the deleted email fails
6. Verify: signing up fresh with the same email works

cner-smith:
## In-app account deletion (#306 + #314)

App Store Guideline 5.1.1(v) (mandatory since June 2022) requires apps that support account creation to allow account deletion inside the app. OGA had Sign Out but no delete flow. Decisions: SQL RPC, immediate hard delete, mobile now (web fast-follow).

### Key simplification
Every user-owned table cascades off auth.users:
```
auth.users -> profiles -> rounds -> hole_scores -> shots
              profiles -> shots, practice_plans
auth.users -> user_clubs
```
So removing the single auth row wipes all data with no per-table DELETEs. The audit's auth.admin.delete_user() SQL call was dropped (that's the service-role GoTrue API, not SQL-callable); a direct DELETE FROM auth.users is the supported path.

### Backend — 0036_account_deletion_rpc.sql
SECURITY DEFINER public.delete_my_account(), search_path='', auth.uid() guard. anon EXECUTE explicitly revoked (Supabase default privileges auto-grant it; revoking from public alone leaves the per-role grant).

### Mobile — Profile screen
"Delete account" (destructive) -> ConfirmDialog -> supabase.rpc('delete_my_account') -> supabase.auth.signOut() (JWT stays valid after the row removal, so the session is cleared; the auth listener redirects to login).

### Verified on a local Postgres
- Migration applies; function is SECURITY DEFINER (owner postgres).
- Grants after migration: authenticated only — anon removed (caught + fixed an over-broad default grant during testing).
- Seeded user + round + clubs, called the RPC with that user's JWT claim: BEFORE users=1 profiles=1 rounds=1 clubs=1 -> AFTER all 0 (txn rolled back). No Supabase Storage in OGA, so the Storage-owner caveat doesn't apply.

### Status
- Mobile + workspace typecheck clean.
- Migration + RPC + cascade proven on local Postgres.
- Device tap-through QA pending; migration applied to dev/prod on the normal deploy.

Web Settings parity (same shared RPC) is a fast-follow.

Closes #306
Closes #314

### DIFF
--- a/apps/mobile/app/(app)/profile.tsx
+++ b/apps/mobile/app/(app)/profile.tsx
@@ -78,9 +78,20 @@ export default function ProfileTab() {
       Alert.alert('Could not delete account', error.message)
       return
     }
-    // Deleting the auth row leaves the current JWT valid until sign-out, so
-    // clear the session explicitly; the auth listener then redirects to login.
-    await supabase.auth.signOut()
+    // The account row is already gone; the JWT just stays valid until
+    // sign-out, so clear the session. If signOut itself errors we must STILL
+    // release the modal — otherwise the user is stranded in a disabled
+    // "Deleting…" dialog after an irreversible delete. On success the auth
+    // listener unmounts this screen and redirects to login.
+    const { error: signOutError } = await supabase.auth.signOut()
+    if (signOutError) {
+      setDeleting(false)
+      setDeleteOpen(false)
+      Alert.alert(
+        'Account deleted',
+        'Your account has been deleted. Restart the app to finish signing out.',
+      )
+    }
   }
 
   const trimmedUsername = username.trim()
--- a/apps/mobile/app/(app)/profile.tsx
+++ b/apps/mobile/app/(app)/profile.tsx
@@ -2,6 +2,7 @@ import { useEffect, useRef, useState } from 'react'
 import {
   Alert,
   Linking,
+  Modal,
   Pressable,
   ScrollView,
   Text,
@@ -15,7 +16,6 @@ import type { Database } from '@oga/supabase'
 import { supabase } from '../../lib/supabase'
 import { useAuth } from '../../hooks/useAuth'
 import { AppBar } from '../../components/ui/AppBar'
-import { ConfirmDialog } from '../../components/ui/ConfirmDialog'
 
 type Profile = Database['public']['Tables']['profiles']['Row']
 type SkillLevel = Profile['skill_level']
@@ -467,12 +467,8 @@ export default function ProfileTab() {
         </Pressable>
       </ScrollView>
 
-      <ConfirmDialog
+      <DeleteAccountModal
         visible={deleteOpen}
-        title="Delete your OGA account?"
-        message="This will permanently delete your account and all rounds, shots, and saved data. This cannot be undone."
-        confirmLabel="Delete account"
-        destructive
         busy={deleting}
         onConfirm={deleteAccount}
         onCancel={() => setDeleteOpen(false)}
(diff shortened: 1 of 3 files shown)


cner-smith (commented):
## Review — feat(mobile): in-app account deletion (#306, #314)

**Concerns**
- `apps/mobile/app/(app)/profile.tsx:78` — `await supabase.auth.signOut()` result is silently discarded. If `signOut` returns an error (transient network issue, SDK error), `deleting` stays `true` and `deleteOpen` stays `true`. `ConfirmDialog` disables **both** the Cancel and Confirm buttons when `busy={true}` (see `ConfirmDialog.tsx:66,75`), so the user ends up locked in a "Working…" modal with no escape — after an account deletion that is irreversible. The happy path works because the auth listener fires and unmounts the component before the user notices, but that's a race the error path loses.

  Suggested fix — capture the signOut result and reset state on failure:
  ```ts
  const { error: signOutError } = await supabase.auth.signOut()
  if (signOutError) {
    setDeleting(false)
    setDeleteOpen(false)
    Alert.alert(
      'Account deleted',
      'Your account has been deleted. Please restart the app to finish signing out.',
    )
  }
  ```
  The message is intentionally different from the RPC error copy — at this point the account *is* gone; the user needs to know that before they see the stuck-looking screen.

---
Looked good: `SECURITY DEFINER` + pinned `search_path = ''` + explicit `REVOKE … FROM public, anon` (catching the Supabase auto-grant edge case) on the migration; `auth.uid()` guard makes the function self-scoping with no input surface to validate; single concern; targets `dev`; no new `@oga/core` exports and no package boundary violations.

---

PR #486, opened 2026-06-01
fix(mobile): Learn navigation — own stack + Home discoverability

cner-smith:
## Mobile Learn — fix Back + surface on Home

Two coupled problems, both rooted in how Learn was wired into the tab navigator. Cross-checked against expo-router 3.5.24 / @react-navigation/bottom-tabs 6.5.20 source before implementing.

### 1. Back bug (fix)
`learn/index` and `learn/[article]` were flat **sibling** `Tabs.Screen` routes. `@react-navigation/bottom-tabs` defaults to `backBehavior: 'firstRoute'`, and `href: null` only hides the tab button — the routes still participate in tab history. So `router.back()` from an article popped the **tab** history to the first route = **Home**, not the list.

**Fix:** add `learn/_layout.tsx` (a `Stack`) so the folder collapses to a single `learn` tab route and `index → [article]` is a real stack push. Back now pops to the **list**.

### 2. Discoverability (feature)
Learn was `href: null` with its only entry a link buried on the Practice screen — "can't find Learn at all."

- New **"From the yardage book"** preview at the bottom of Home: first 3 readable articles (catalog order → surfaces the two flagship published pieces) + a **"See all →"** into the full list.
- A muted top-of-Home **hint** (`↓ New to a stat? The yardage book's at the bottom.`) that smooth-scrolls down to the section.

### Notes
- Article selection computed inline — no new `@oga/core` field/helper (single caller).
- The Practice → Learn link still resolves to the list (verified).
- Edge case (cross-checked): opening an article straight from the Home preview → Back returns to Home in the common case; if the Learn stack was already visited that session it returns to the list. Both acceptable; inherent to tab+nested-stack.

### Verification
- ✅ Mobile typecheck clean.
- ⏳ **Device QA pending** (author at work, no device): Back-from-article → list, preview placement/spacing, scroll-to-hint feel.

Files: `learn/_layout.tsx` (new), `components/home/LearnPreview.tsx` (new), `(app)/_layout.tsx` (route collapse), `(app)/index.tsx` (hint + preview).

PR #491, opened 2026-06-02
fix(mobile): Learn back button returns to the list, not Home

cner-smith:
Split from the original combined Learn-nav PR (#486) per one-concern — this is the **bug fix half**.

`learn/index` and `learn/[article]` were flat sibling `Tabs.Screen` routes. `@react-navigation/bottom-tabs` defaults to `backBehavior: 'firstRoute'`, and `href: null` only hides the tab button (routes still live in tab history), so `router.back()` from an article popped the **tab** history to the first route = Home, not the list.

**Fix:** `learn/_layout.tsx` (a `Stack`) collapses the folder to one `learn` tab route; `index → [article]` becomes a real stack push → Back pops to the list. Cross-checked against expo-router 3.5.24 / bottom-tabs 6.5.20 source.

✅ Mobile typecheck clean. The Home 'yardage book' preview ships separately (feature/learn-home-preview).

### DIFF
--- a/apps/mobile/app/(app)/_layout.tsx
+++ b/apps/mobile/app/(app)/_layout.tsx
@@ -220,8 +220,10 @@ export default function AppLayout() {
           ),
         }}
       />
-      <Tabs.Screen name="learn/index" options={{ href: null }} />
-      <Tabs.Screen name="learn/[article]" options={{ href: null }} />
+      {/* Learn is a nested stack (learn/_layout.tsx owns index + [article]),
+          so it collapses to a single hidden tab route. This is what makes
+          "← Back" from an article pop to the list instead of the Home tab. */}
+      <Tabs.Screen name="learn" options={{ href: null }} />
       <Tabs.Screen name="bag" options={{ href: null }} />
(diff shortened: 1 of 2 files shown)


cner-smith (commented):
## Review — fix(mobile): Learn back button returns to the list, not Home

**Nits**
- `apps/mobile/app/(app)/learn/_layout.tsx:1–7` — The block comment narrates what the Stack does ("pressing '← Back' in an article pops to the list instead of falling through to the Home tab") rather than just stating the non-obvious *why*. CONTRIBUTING.md: "Don't narrate what the code does." The last sentence ("Screens render their own AppBar, so the stack header stays hidden") is exactly the kind of non-obvious why that warrants a comment; the rest can go. The inline JSX comment in `_layout.tsx` already carries the routing-model explanation for anyone reading the tabs file — the learn layout comment can afford to be just that one line about `headerShown`.

---
Looked good: correct target branch (`dev`), single concern (split cleanly from #486 per the description), no package-boundary violations, no `any`, no dependency changes, no migrations.

---

PR #492, opened 2026-06-02
feat(mobile): 'From the yardage book' preview on Home

cner-smith:
Split from the original combined Learn-nav PR (#486) per one-concern — this is the **feature half**.

Learn was hard to find (not a tab; only a buried link on Practice). Adds a 'From the yardage book' section at the bottom of Home — first 3 readable articles in catalog order (surfacing the flagship published pieces) + 'See all' — plus a top-of-Home hint that scrolls down to it.

- Per your review: kept **first-3 incl. the draft Glossary** card (draft articles are tappable + render fine).
- `LearnPreview` lives in `components/home/` alongside the other Home sections (RoundCTAs, SGBreakdown, …) — the established pattern for Home blocks.
- Article selection computed inline (no new `@oga/core` export, single caller).

✅ Mobile typecheck clean (standalone — independent of the back-nav fix). Recommended merge order: back-nav fix first, then this.

### DIFF
--- a/apps/mobile/app/(app)/index.tsx
+++ b/apps/mobile/app/(app)/index.tsx
@@ -1,5 +1,5 @@
-import { useCallback, useEffect, useMemo, useState } from 'react'
-import { ScrollView, Text, View } from 'react-native'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { Pressable, ScrollView, Text, View } from 'react-native'
 import { formatSG } from '@oga/core'
 import { deleteRound, getProfile, getRecentSGData } from '@oga/supabase'
 import type { Database } from '@oga/supabase'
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — feat(mobile): 'From the yardage book' preview on Home

**Nits**
- `apps/mobile/components/home/LearnPreview.tsx:5` — `KICKER` is annotated with an inline import type (`import('react-native').TextStyle`) rather than adding `TextStyle` to the existing named import on line 1. The inline form works but is inconsistent with the rest of the file; just add `TextStyle` to the `react-native` import and use it directly.

---
Looked good: targets `dev`, no new `@oga/core` exports (test-coverage rule doesn't trigger), `LearnPreview` placed in `components/home/` following the established Home-block pattern, scope held to one concern, `learnYRef` / `onLayout` approach correctly captures offset within the scroll content container.

---

PR #501, opened 2026-06-02
fix(mobile): play the brand splash on every login (#500)

### LINKED ISSUE #500: [mobile] Brand splash never plays after login (only on a later cold launch)
## Symptom
The animated brand splash (the "OGA / Track every shot." screen on ink) is supposed to play **after every login**. It doesn't — it only appears on a later cold app-launch, seemingly at random, never right after you log in.

## Root cause (code-confirmed)
The splash was an overlay in the **root layout** (`app/_layout.tsx`), and whether it animates was decided **once**, in a `useEffect([])` that runs only when the root layout mounts (cold start). It read an AsyncStorage flag `oga.pending-splash`.

login/signup set that flag and then `router.replace('/(app)')`:
```
login.tsx:45  await AsyncStorage.setItem('oga.pending-splash', '1')
login.tsx:46  router.replace('/(app)')
```
But `router.replace` is **in-session navigation** — it does **not** remount `app/_layout.tsx`. So the mount-time `[]` effect never re-runs, the overlay is already dismissed, and the flag just sits in storage until the **next cold launch**. AsyncStorage isn't reactive, so setting it mid-session is invisible to the already-mounted layout.

## Fix
Move the splash to its own route (`app/(auth)/welcome.tsx`); login/signup `router.replace` there, it plays the staged fade, then forwards into `/(app)`. Plays on every login by construction; no flag. The root layout keeps only a cold-start loading gate.

label: bug — observed (user-reported).

cner-smith:
Closes #500

## The bug
The animated brand splash ("OGA / Track every shot." on ink) was supposed to play after every login but only appeared on a later cold launch.

**Root cause (code-confirmed):** the splash was an overlay in the **root layout** (`app/_layout.tsx`); whether it animates was decided **once** in a mount-time `useEffect([])` reading the AsyncStorage flag `oga.pending-splash`. login/signup set the flag then `router.replace('/(app)')` — but in-session navigation **doesn't remount the root layout**, so the flag was never re-read and the splash deferred to the next cold start. AsyncStorage isn't reactive, so a mid-session write is invisible to the mounted layout.

## The fix (Option B — splash as its own route)
- **New `app/(auth)/welcome.tsx`** — the staged-fade brand splash (logo → tagline → support), tap-to-skip, reduced-motion aware, then `router.replace('/(app)')`. The `(auth)` group has no auth guard, so it renders fine post-login.
- **`login.tsx` / `signup.tsx`** — `router.replace('/(auth)/welcome')` instead of `/(app)`; removed the now-dead `AsyncStorage` flag (+ its import).
- **`_layout.tsx`** — stripped the overlay + flag + animation (now in the route); keeps only the native-splash-for-fonts gate and a brief loading gate over the cold-start auth-resolve window.

Now the splash plays after **every** login by construction (you pass through the route). A resumed session (cold-reopen while already logged in) shows just the native "o." splash, not the full animation — matching "after every login."

## Verification
- `tsc --noEmit` green under CI conditions (`.expo/types` absent → expo-router's permissive `Href`; the new `/(auth)/welcome` route resolves). The local-only error was a stale generated `router.d.ts`, regenerated on next `expo start`.
- **Not device-verified** — the on-device animation/timing is QA on the next build. Logic is a faithful port of the previous overlay choreography.

### DIFF
--- a/apps/mobile/app/(auth)/login.tsx
+++ b/apps/mobile/app/(auth)/login.tsx
@@ -9,7 +9,6 @@ import {
   View,
 } from 'react-native'
 import { Link, useRouter } from 'expo-router'
-import AsyncStorage from '@react-native-async-storage/async-storage'
 import { WebView } from 'react-native-webview'
 import { supabase } from '../../lib/supabase'
 
@@ -42,8 +41,9 @@ export default function Login() {
       setCaptchaToken(null)
       return
     }
-    await AsyncStorage.setItem('oga.pending-splash', '1').catch(() => {})
-    router.replace('/(app)')
+    // Route through the brand splash (plays on every login), which then
+    // forwards into the app. See app/(auth)/welcome.tsx (#500).
+    router.replace('/(auth)/welcome')
   }
 
   return (
--- a/apps/mobile/app/(auth)/signup.tsx
+++ b/apps/mobile/app/(auth)/signup.tsx
@@ -9,7 +9,6 @@ import {
   View,
 } from 'react-native'
 import { Link, useRouter } from 'expo-router'
-import AsyncStorage from '@react-native-async-storage/async-storage'
 import { WebView } from 'react-native-webview'
 import { supabase } from '../../lib/supabase'
 
@@ -44,8 +43,9 @@ export default function Signup() {
       setCaptchaToken(null)
       return
     }
-    await AsyncStorage.setItem('oga.pending-splash', '1').catch(() => {})
-    router.replace('/(app)')
+    // Route through the brand splash (plays on every login), which then
+    // forwards into the app. See app/(auth)/welcome.tsx (#500).
+    router.replace('/(auth)/welcome')
   }
 
   return (
--- a/apps/mobile/app/(auth)/welcome.tsx
+++ b/apps/mobile/app/(auth)/welcome.tsx
@@ -1,0 +1,127 @@
+import { useEffect, useRef } from 'react'
+import { Pressable, StyleSheet } from 'react-native'
+import { useRouter } from 'expo-router'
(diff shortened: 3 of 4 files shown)


cner-smith (commented):
## Review — fix(mobile): play the brand splash on every login (#500)

**Blockers**
- `apps/mobile/app/(auth)/welcome.tsx:63` — `// eslint-disable-next-line react-hooks/exhaustive-deps` has no inline explanation. CONTRIBUTING.md: "No `// eslint-disable-next-line` without a comment explaining why." Something like `-- intentional mount-only; router and reducedMotion are stable refs, started.current gates re-arms` on the same line would satisfy the rule.

**Concerns**
- `apps/mobile/app/(auth)/welcome.tsx:45` — `router.replace('/(app)')` runs without an auth check. The PR description notes `(auth)` carries no guard, so an unauthenticated user who arrives at `/(auth)/welcome` via deep link would be forwarded straight into `/(app)`. Can't tell from the diff whether `/(app)` has its own auth redirect that catches this; if it doesn't, this is a hole.
- `apps/mobile/app/(auth)/welcome.tsx:42` — `reducedMotion` is read once at mount because the effect dependency array is `[]`. The old `_layout.tsx` had a separate effect with `[reducedMotion, ...]` that canceled in-flight animations if the user toggled the accessibility preference mid-splash. That reactive behavior is silently dropped here. The window is narrow (~3.5 s), but it's a regression against the prior implementation that's worth a conscious call one way or the other.

---
Looked good: clean removal of the AsyncStorage flag pattern from both login and signup, `_layout.tsx` reduced to a pure loading gate with no leftover animation state, PR targets `dev`, no dependency changes.

---

Issue #475, opened 2026-05-31
Extract useClubDispersion grouping/selection into @oga/core (web↔mobile dedup)


cner-smith:
The web `useClubDispersion` hook (`apps/web/src/pages/rounds/hooks/useClubDispersion.ts`) is a hand-mirror of the mobile one (`apps/mobile/components/round/hole/useClubDispersion.ts`). The pure, platform-agnostic logic — `rowToShot`, `median`, the by-club grouping, and `selectClub` (nearest-median match + longest-club fallback) — is now duplicated in two places and will silently drift.

Per CLAUDE.md (pure domain logic lives in `@oga/core`, never in app code), promote the pure parts into `@oga/core` (e.g. `buildClubDispersions(shots)` + `selectClubDispersion(byClub, distance)`) with tests, leaving each platform a thin React + Supabase-client wrapper.

Both hook files already carry a comment flagging this. Surfaced by the PR #474 agent review. Under the 3-caller threshold (2 callers) so it's a judgment call, but the logic is non-trivial.

Effort: small. No user-facing change.

cner-smith:
Scope addendum (from PR #474 review): also lift the duplicated constants `AIM_AUTOSPAWN_FRACTION = 0.65` (web `useRoundActions`, mobile `useHoleState`) and `FEET_PER_YARD = 3` (web `MapView`, mobile `LiveRoundSession`) into @oga/core alongside the dispersion logic.

PR #474, opened 2026-05-31
feat: Shot-pattern live-round map — mobile redesign + web parity (#471)

cner-smith:
## Shot-pattern live-round map — mobile redesign + web parity

Rebuilds the live-round map into a shotpattern.app-style planning surface
(drag-the-aim + dispersion overlay + live readouts) on **mobile**, then ports
the visible features to **web**'s existing round map. Shared geometry/dispersion
math lives in `@oga/core` so both platforms compute identically.

Closes #471. Also folds in #472 (course-add gate) and a round-holes RLS fix.

### Shared foundation
- `@oga/core`: aim-relative dispersion + great-circle arc/circle/scatter geo
  (`arcGeoJSON`, `circleGeoJSON`, `scatterGeoJSON`, `destinationYards`,
  `computeAimRelativeDispersion`). `@oga/supabase`: `getShotsForUser`.

### Mobile (live-round redesign — device-validated across sessions)
- Drag-the-aim with auto-spawn (65% ball→pin), white aim bend + dotted
  reference + crosshair, carry/remaining + best-case live SG.
- Left toolbar (dispersion-dots toggle, place pin/tee), Tee/Appr overlay + distance
  rail (fixed-geometry arc/circle), floating bottom chrome (replaced the cream strip).
- Fixes: holes materialized via `insert_synthetic_hole` RPC (RLS); aim-drag
  stability (PointAnnotation prop stability); css-interop static styles; aim UX
  (drag/long-press conflict, no aim prompt, ⋮ menu); re-place-ball dup aim.
- **State-machine fixes (this round, NOT yet device-validated):** hole switch
  now resets ball/aim/roundState (were frozen across holes); aim/ball stay
  draggable across pin placement (annotation no longer unmounts→remounts).

### Web (parity port onto the existing round map — NOT yet browser-QA'd)
- **A** — live auto-spawn aim + draggable aim marker; white solid bend + dotted
  reference + CARRY/REMAINING on the latest aimed shot; past round keeps the
  explicit aim prompt.
- **B** — Tee/Appr toggle + distance rail (right edge); arc band (tee) / approach
  circle centered on the aim (appr), drawn beneath the aim line.
- **C** — left-edge "Pattern" toggle → single-color dispersion dots (web
  `useClubDispersion`, club picked by ball→pin distance; sparse clubs → none).
- **D** — expected-strokes HUD pill + best-case SG sublabel on the carry pill,
  calibrated to the player's handicap.

### QA status / notes
[...]

PR #502, opened 2026-06-02
fix(core): frame Patterns dispersion aim-relative, not compass N/E (#464)

### LINKED ISSUE #464: [core] BUG: Patterns dispersion is compass-framed, not aim-relative
Found during the Shot-Pattern map review. computeDispersion (packages/core/src/shot-patterns.ts:66-67) measures offsets as North/East — (endLat-aimLat), (endLng-aimLng) — NOT relative to each shot's aim line. So cone95.lateral is east-west spread, not perpendicular-to-shot. The Patterns plot looks plausible because it's direction-agnostic, but aggregating shots aimed in different compass directions smears the ellipse and mislabels lateral vs distance spread. Fix: rotate each shot's (end-aim) vector into the aim-relative frame (perp/along) using the start→aim bearing before computing spreads. The mobile live-map redesign introduces the correct aim-relative helper in @oga/core; the Patterns page should adopt it.

cner-smith:
Closes #464.

## The bug

`computeDispersion` measured each shot's end-vs-aim offset as raw **North/East** — `(endLat−aimLat)`, `(endLng−aimLng)` — so `lateralOffsetYards` was east-west spread, not perpendicular-to-the-shot. Aggregating shots aimed on different compass bearings smeared the ellipse and mislabeled lateral vs distance spread. Everything downstream that reads `avgLateralOffset` (dominant-miss, shot-shape, and the **aim-correction text**) was wrong too, not just the plot.

The correct rotation already existed in the same module — `aimRelativeOffsets()`, added for the live-round overlay (#474) — but the Patterns pages never adopted it.

## The fix

Rotate each shot's `(end−aim)` **and** `(start−aim)` vectors into the aim-relative frame by the `start→aim` bearing, via a shared `rotateAroundAim` core that `aimRelativeOffsets` now also uses (one bearing implementation, can't drift). `DispersionPoint`'s shape and field names are unchanged — `lateralOffsetYards` now correctly means perpendicular-to-aim (+ = right), `distanceOffsetYards` means along-aim (+ = long) — so **every consumer is corrected with zero consumer-side edits**: web + mobile dispersion plots, both 1200×630 share cards, the ball-flight chart, `getAimCorrection`, and `dispersionVerdict`.

Aim-relative framing requires a start position (to derive the line), so `computeDispersion` now skips start-less shots. **Data check: all 727 dev shots that have aim+end already have a start** — the post-round placement flow always writes it — so nothing drops in practice.

## Tests

TDD: added two `#464` regressions first (an east-aimed straight shot must read ~0 lateral; identical perpendicular misses flown on different bearings must read identically), watched them fail against the compass code, then fixed. Re-derived the existing `computeDispersion` assertions for the aim frame.

- `@oga/core`: **521 tests pass**
- web + mobile + supabase typecheck: clean

## Visual QA

Signed-in capture of `/patterns` (driver, 128 seeded shots) on the dev e2e user: 68/95 cones compute and center on AIM; summary reads Avg Lateral 0.2 yd / Straight / Straight / ±15.2/12.5 yd / "well centered on target." No regression; ball-flight chart renders.

Scope note: `computeAimRelativeDispersion` (live-round overlay) is left untouched — the web↔mobile dedup is already tracked in #475.

### DIFF
--- a/packages/core/src/shot-patterns.test.ts
+++ b/packages/core/src/shot-patterns.test.ts
@@ -32,9 +32,11 @@ describe('computeDispersion', () => {
   })
 
   it('converts lat/lng deltas to yard offsets', () => {
-    // 0.001 deg lat ≈ 121 yards north → distanceOffset ~ +121
-    // 0.001 deg lng at lat 40 ≈ 121000 * cos(40°) ~ 92,690 yards/deg
+    // Aimed due north (start south of aim). End 0.001 deg lat ≈ 121 yds north
+    // of aim = straight down the line → distanceOffset ~ +121, lateral ~0.
     const s = shot({
+      startLat: AIM_LAT - 0.001,
+      startLng: AIM_LNG,
       aimLat: AIM_LAT,
       aimLng: AIM_LNG,
       endLat: AIM_LAT + 0.001,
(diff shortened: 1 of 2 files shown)


cner-smith (commented):

Looked good: rotation math verified correct for north- and east-aimed cases, `hasAimGeometry` type predicate necessary for TypeScript narrowing so 2-caller extraction is justified, `rotateAroundAim` has 3 call sites satisfying the 3-caller rule, two TDD regression tests are geometrically sound, all changes stay in `packages/core/src/`, targets `dev`.

---

PR #504, opened 2026-06-02
fix(mobile): iOS press feedback via PressableTouch (#303)

### LINKED ISSUE #303: iOS: Pressable uses android_ripple with no iOS fallback visual
**Severity:** low (UX polish — inconsistent tap affordance between platforms)
**Audit source:** mobile-app-developer agent, 2026-05-13

## Mechanism

Two sites use \`android_ripple\` for tap feedback:
- \`apps/mobile/app/(app)/round/[id]/index.tsx:458\`
- \`apps/mobile/components/round/PastHoleShotsSheet.tsx:208\`

\`android_ripple\` is silently ignored on iOS. The Pressable still dims via default opacity behavior, but the user gets no parallel ripple/highlight. Not a rejection risk, just inconsistent feel.

## Proposed fix

Pair each \`android_ripple\` with an iOS-compatible visual via the \`style\` callback or \`hitSlop\` adjustments. A simple cross-platform pattern:

```tsx
<Pressable
  android_ripple={{ color: '#xxx' }}
  style={({ pressed }) => [
    styles.base,
    pressed && Platform.OS === 'ios' && { opacity: 0.7 },
  ]}
>
```

Or extract a shared \`<TapButton>\` component that renders ripple on Android and an opacity transition on iOS.

## Files

- \`apps/mobile/app/(app)/round/[id]/index.tsx:458\`
- \`apps/mobile/components/round/PastHoleShotsSheet.tsx:208\`

cner-smith:
Closes #303.

`android_ripple` is silently ignored on iOS, so 11 `Pressable`s across 5 live-round files had **no press-down feedback** there (persistent active/selected fills still rendered; only the momentary press state was missing).

## Fix
New `apps/mobile/components/ui/PressableTouch.tsx` — a `Pressable` that keeps the native `android_ripple` on Android and dims opacity (0.75, UIKit's press alpha) on iOS via `onPressIn`/`onPressOut` + a **static style array**.

The textbook `style={({ pressed }) => …}` callback **cannot** be used here: NativeWind 4 css-interop silently drops function `style` on wrapped RN components ([nativewind #847](https://github.com/nativewind/nativewind/issues/847)) — the exact trap every one of these sites already dodged with static style objects. The dim is gated to genuinely-interactive presses (`!disabled && onPress != null`) so non-played scorecard rows and disabled buttons don't flash.

All 11 `android_ripple` sites swapped `Pressable`→`PressableTouch`; their static styles are unchanged. The `active:` className variant was rejected (open NativeWind bugs [#1583](https://github.com/nativewind/nativewind/issues/1583)/[#1678](https://github.com/nativewind/nativewind/issues/1678) + it would reintroduce css-interop on a surface deliberately kept on plain styles).

## Sites (5 files)
`round/[id]/index.tsx` scorecard row · `PastHoleShotsSheet` row · `HoleMapOverlays` ToolbarButton + RailPill · `LiveRoundSession` header ⋮ + End/Delete menu · `MapBottomChrome` PrimaryCta/SecondaryPill/TextChip/NavChevron.

## Review & verification
- Plan cross-checked by **two agents** (RN/NativeWind correctness + repo-fit/regression) — both signed off; their two refinements (0.75 default, ref-not-forwarded JSDoc note) are incorporated.
- `npm run typecheck` clean (validates every tag swap + that all props forward).
- No automated test: mobile has no RN component-test harness and this is presentational/framework behavior (repo norm).

## Manual QA (device/simulator)
- [ ] iOS: each touched control dims on press; **tap a non-played scorecard hole row → no opacity flash**.
- [ ] Android: ripple unchanged on every control.

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -7,6 +7,7 @@ import {
   Text,
   View,
 } from 'react-native'
+import { PressableTouch } from '../../../../components/ui/PressableTouch'
 import { captureRef } from 'react-native-view-shot'
 import * as Sharing from 'expo-sharing'
 import { useFocusEffect, useLocalSearchParams, useRouter } from 'expo-router'
@@ -555,7 +556,7 @@ export default function RoundIndex() {
             const d = score != null && score > 0 ? score - h.par : null
             const hasShots = hs ? holeScoreIdsWithShots.has(hs.id) : false
             return (
-              <Pressable
+              <PressableTouch
                 key={h.id}
                 accessibilityRole={hasShots ? 'button' : 'text'}
                 accessibilityLabel={
@@ -634,7 +635,7 @@ export default function RoundIndex() {
                 >
                   {hasShots ? '→' : ''}
                 </Text>
-              </Pressable>
+              </PressableTouch>
             )
           })}
         </View>
--- a/apps/mobile/components/round/HoleMapOverlays.tsx
+++ b/apps/mobile/components/round/HoleMapOverlays.tsx
@@ -1,5 +1,6 @@
 import type { ComponentProps } from 'react'
-import { Pressable, Text, View } from 'react-native'
+import { Text, View } from 'react-native'
+import { PressableTouch } from '../ui/PressableTouch'
 import { MaterialCommunityIcons } from '@expo/vector-icons'
 
 // Shared pill chrome — dark translucent card on satellite, per DESIGN.md
@@ -287,7 +288,7 @@ function ToolbarButton({
   disabled?: boolean
 }) {
   return (
-    <Pressable
+    <PressableTouch
       accessibilityRole="button"
       accessibilityLabel={label}
       accessibilityState={{ disabled: !!disabled, selected: !!active }}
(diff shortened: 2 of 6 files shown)


cner-smith (commented):
## Review — fix(mobile): iOS press feedback via PressableTouch (#303)

**Concerns**
- `apps/mobile/components/ui/PressableTouch.tsx:16` — `pressedOpacity` is an exposed prop with no callers; none of the 11 swap sites pass it. CONTRIBUTING.md: "No config vars, options, or parameters added 'for hypothetical future use' — add them when there's a real caller." The constant `IOS_PRESSED_OPACITY` already pins the value; the prop adds API surface with no current use. Remove the prop and reference the constant directly in the `dim` expression.

**Nits**
- `apps/mobile/components/ui/PressableTouch.tsx:31` — The final JSDoc line ("Default tap target for the app; reach for this instead of a bare Pressable.") is a usage directive, not a *why*. CONTRIBUTING.md: "Comments only when the *why* is non-obvious from the name and types alone." The NativeWind css-interop explanation and the ref-forwarding note above it are justified; this line is not.

---
Looked good: `Pressable` imports in `MapBottomChrome` and `PastHoleShotsSheet` are legitimately kept — both files still use bare `Pressable` for elements without `android_ripple` that are correctly out of scope; `dim` guard correctly excludes disabled and no-`onPress` rows; `HoleMapOverlays` properly drops `Pressable` from its import entirely; PR targets `dev`, scope held to one concern, no security issues.

---

Issue #376, opened 2026-05-21
iOS: execute first development build end-to-end and verify on physical device


cner-smith:
The Apple Developer Program account is active as of 2026-05-21 and a physical iOS device is now available for testing. The codebase has never been built for iOS — the `ios/` directory is gitignored (Expo CNG) and no `eas build --platform ios` has ever run.

This issue coordinates the discrete pieces that are already filed and gets a runnable IPA onto the dev phone so VoiceOver / iOS-only bugs can actually be tested.

## Gating dependencies

- **#297** — Apple Developer Program enrollment + App Store Connect setup _(active, confirm + close if done)_
- **#298** — Add iOS build targets to `eas.json` (development / preview / production profiles)
- **#304** — `usesNonExemptEncryption=false` in `app.json` `ios.config` to skip the export-compliance prompt on every build
- **#305** — Stub `NSCameraUsageDescription` + `NSPhotoLibraryAddUsageDescription` in `app.json` pre-emptively (avoid second build cycle)
- **#313** — TestFlight setup (internal + external testing tracks)

## Current state of iOS scaffolding

- ✅ `app.json` `ios.bundleIdentifier` = `app.opengolf.oga`
- ✅ `app.json` `ios.infoPlist` has `NSLocationWhenInUseUsageDescription` + `NSLocationAlwaysAndWhenInUseUsageDescription` _(#301 still wants the Always version dropped + copy rewritten — separate issue)_
- ❌ `eas.json` has NO iOS-specific settings — all three profiles only have `android:` blocks (covered by #298)
- ❌ `eas credentials` never run for iOS — first build will prompt for team + distribution cert + provisioning profile (interactive)
- ❌ No App Store Connect app registered yet (EAS creates it on first build given correct bundle ID)
- ❌ No device UDID registered for ad-hoc distribution / no TestFlight tester invited

## Acceptance criteria

- `eas build --platform ios --profile development` (or `preview` if dev client adds friction) succeeds and produces an IPA
- IPA installs on the dev iOS phone (TestFlight internal tester invite OR ad-hoc with UDID registered)
- OGA opens, lands on splash, reaches login screen
- Mapbox tiles load on a live-round view (validates `MAPBOX_TOKEN` propagation and `@rnmapbox/maps` iOS config)
- VoiceOver enabled — can navigate the rounds list, hear the structured labels (verifies the in-flight #286 a11y work cross-platform)
[...]

Issue #299, opened 2026-05-13
iOS: declare App Privacy Manifest (PrivacyInfo.xcprivacy) — Apple May-2024 requirement


cner-smith:
**Severity:** blocker (App Store Connect rejects builds without it)
**Audit source:** certification-planner agent, 2026-05-13

## Background

Required for all App Store submissions since May 1, 2024. Apple enforces at upload time — ASC rejects missing manifests for known SDKs.

## Required declarations

Add to \`apps/mobile/app.json\` under \`ios.privacyManifests\`. Expo SDK 51 reads this and generates the \`PrivacyInfo.xcprivacy\` file at prebuild time (don't hand-edit — CNG model).

### NSPrivacyTracking
\`false\` — OGA has no cross-app tracking, no IDFA, no ad networks.

### NSPrivacyCollectedDataTypes

| Data type | Linked to user | Tracking | Purpose |
|-----------|----------------|----------|---------|
| Email address | Yes | No | App Functionality (auth) |
| Precise location | Yes | No | App Functionality (shot tracking) |
| Other user content (golf rounds, shots) | Yes | No | App Functionality |
| Crash data (when Sentry/similar added) | No | No | App Functionality |

### NSPrivacyAccessedAPITypes — Required Reason APIs we touch

| Category | Reason code | Why |
|---|---|---|
| FileTimestamp | \`C617.1\` or \`0A2A.1\` | expo-sqlite touches file mtimes |
| SystemBootTime | \`35F9.1\` | React Native core uses for performance timing |
| DiskSpace | \`E174.1\` | SQLite availability checks |
| UserDefaults | \`CA92.1\` | AsyncStorage uses NSUserDefaults on iOS |

### Suggested app.json snippet

```json
\"ios\": {
  \"privacyManifests\": {
    \"NSPrivacyTracking\": false,
    \"NSPrivacyCollectedDataTypes\": [
      { \"NSPrivacyCollectedDataType\": \"NSPrivacyCollectedDataTypeEmailAddress\",
        \"NSPrivacyCollectedDataTypeLinked\": true,
        \"NSPrivacyCollectedDataTypeTracking\": false,
        \"NSPrivacyCollectedDataTypePurposes\": [\"NSPrivacyCollectedDataTypePurposeAppFunctionality\"] },
      { \"NSPrivacyCollectedDataType\": \"NSPrivacyCollectedDataTypePreciseLocation\",
        \"NSPrivacyCollectedDataTypeLinked\": true,
[...]

PR #505, opened 2026-06-02
docs(ios): first iOS build + on-device smoke runbook (#376)

cner-smith:
Tees up the iOS build chain. Adds a "First iOS build & on-device smoke (#376)" section to `docs/ios-submission.md`.

## What I found verifying against current code
The chain is **further along than the issues say** — several gating tickets are already satisfied (the audit cited a deleted `app.json` and the wrong bundle id):
- **#298** (eas.json iOS targets) — ✅ already present in all three profiles. `development`/`preview` use `ios.simulator: false` (device builds, correct for a Linux+iPhone host); `production` omits a platform key (builds both). No eas.json change needed.
- **#304** (`ITSAppUsesNonExemptEncryption: false`) — ✅ in `app.config.ts`.
- **#305** (camera + photo-library usage strings) — ✅ in `app.config.ts`.
- Bundle id `golf.oga.app` (not the stale `app.opengolf.oga`, which is the *Android* package), `buildNumber:'1'` autoIncrement seed, privacy manifest (#299), Mapbox download-token plugin, `projectId` — all ✅.

## The runbook covers
- Device-build path for a **Linux + physical iPhone (no Mac)** host: EAS Cloud build + ad-hoc internal distribution (`eas device:create` → `eas build -p ios --profile preview`). Simulator builds are useless without a Mac.
- The **two Mapbox tokens** the build needs and how each is wired: `MAPBOX_DOWNLOADS_TOKEN` (secret, iOS CocoaPod) and `EXPO_PUBLIC_MAPBOX_TOKEN` (public, runtime, inlined at build time via `lib/maps.ts`).
- The #376 on-device smoke checklist.

## Flagged gap (not fixed here)
While verifying the font smoke: **`JetBrainsMono-Medium` is referenced in 2 places but never bundled** (no font file, not in `useFonts`) → those labels fall back to the system font on **both** platforms. Cosmetic, pre-existing, not a crash. Decision noted in the runbook: bundle it (OFL) or drop the 2 refs.

Docs-only. The actual build needs your Apple/Expo/Mapbox accounts (can't be done from here) — prerequisites are spelled out.

### DIFF
--- a/docs/ios-submission.md
+++ b/docs/ios-submission.md
@@ -41,3 +41,77 @@ A reviewer could question WebView use under Guideline 4.2.7 (HTML5 wrappers) or
 **If rejected anyway:** escalate to the App Review Board with the response
 above — high probability of acceptance. Budget one round trip in the submission
 timeline.
+
+---
+
+## First iOS build & on-device smoke (#376)
+
+OGA has never been built for iOS. This is the runbook to get a signed IPA onto
+the test iPhone. Build host is Linux + a physical iPhone (no Mac) → builds run
+on EAS Cloud and install on the device via **internal (ad-hoc) distribution**; a
+simulator build is not useful here (it only runs in the macOS Simulator).
+
+### Config status (already in place — #298/#304/#305 satisfied)
+- `eas.json` — all three profiles carry iOS keys. `development`/`preview` use
+  `ios.simulator: false` (device builds); `production` omits a platform key so
+  it builds both OSes. **No eas.json change needed.**
+- `app.config.ts` `ios` — bundle id `golf.oga.app`, `buildNumber: '1'` (seed for
+  `autoIncrement` with `appVersionSource: 'remote'`), `supportsTablet`,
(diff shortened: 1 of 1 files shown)


cner-smith (commented):
## Review — docs(ios): first iOS build + on-device smoke runbook (#376)

**Nits**
- `docs/ios-submission.md:68` — "memory says the account went active ~2026-05-21" is odd phrasing for a committed reference document. Future readers won't know whose memory or how reliable it is, and the date may be stale by the time someone runs this runbook. Either verify and commit a concrete fact ("confirmed active 2026-05-21 — see #297") or replace with a plain TODO: "confirm #297 is closed before proceeding."

---
Looked good: docs-only, no code touched, targets `dev`, single concern held throughout, config-status claims are cross-referenced against real tickets, the JetBrainsMono gap is flagged honestly rather than silently skipped.

---

cner-smith:
Addressed the nit: `docs/ios-submission.md:67` no longer relies on "memory says … ~2026-05-21" — replaced with a plain actionable TODO ("confirm #297 is closed before proceeding"), per the suggestion.

Issue #509, opened 2026-06-03
[mobile] Email confirmation redirects to localhost — never returns to the app


cner-smith:
## Symptom (observed)

Signing up on mobile (dev build, on a physical phone) sends a confirmation email, but tapping **Confirm** in the email opens **`localhost`** instead of returning to the app. The new account is never brought back into a signed-in session on the device.

**Repro:** sign up in the mobile app → receive confirmation email → tap the confirm link → browser lands on `localhost`, app is never reopened/authenticated.

## Root cause

The mobile email-confirmation deep-link flow was unimplemented end-to-end:

1. **`signUp()` sent no `emailRedirectTo`** (`apps/mobile/app/(auth)/signup.tsx`), so GoTrue built the confirmation link off the project **Site URL** (`localhost`). Web sets it (`emailRedirectTo: ${origin}/onboarding`); mobile didn't.
2. **No `oga://` entry in the Supabase Auth Redirect URLs allow-list** — GoTrue silently falls back to Site URL when a redirect isn't allow-listed (a second path to `localhost`).
3. **No deep-link handler in the app** — the native client is `detectSessionInUrl: false` and there was no `Linking`/`setSession`-from-URL handler, so even a correct `oga://` redirect would not establish a session.

Note: GoTrue's `/auth/v1/verify` endpoint confirms the email server-side *before* redirecting, so affected accounts are actually confirmed — the user can sign in manually. Only the auto-return was broken.

## Fix

PR: `fix/mobile-auth-deep-link`.

- Mobile `signUp()` now sends `emailRedirectTo: Linking.createURL('auth-callback')`.
- New `app/auth-callback.tsx` route parses the session from the redirect URL **fragment** (implicit flow) and calls `supabase.auth.setSession`, then routes to the post-login splash.
- Post-signup now shows a **"Check your email"** state (no session exists until confirmation) instead of bouncing to the splash → login.

## Remaining before this is resolved (launch blocker)

- [ ] **Dashboard:** add `oga://**` to **dev** project Auth → URL Configuration → Redirect URLs (and set a real Site URL, not localhost).
- [ ] **Dashboard:** same for **prod** project (`https://oga.golf` Site URL + `oga://**` redirect) before dual-store launch.
- [ ] **Device verification:** build an EAS dev/preview build and confirm the full tap-through on a physical phone (Expo Go won't exercise the `oga://` scheme).

Dual-store launch blocker — mobile signup confirmation must round-trip on both platforms.

[...]

PR #510, opened 2026-06-03
fix(mobile): deep-link email confirmation back into the app

cner-smith:
Refs #509.

Mobile signup's confirmation email redirected to `localhost` and never returned to the app, because `signUp()` sent no `emailRedirectTo` (GoTrue fell back to the project Site URL) and there was no deep-link handler to establish a session.

## Changes

- **`app/(auth)/signup.tsx`** — send `emailRedirectTo: Linking.createURL('auth-callback')`; after a successful `signUp()` show a **"Check your email"** state (no session exists until the email is confirmed) instead of bouncing to the post-login splash → login.
- **`app/auth-callback.tsx`** (new route) — receives `oga://auth-callback`, parses the session from the redirect URL **fragment** (implicit flow; native client is `detectSessionInUrl: false`), calls `supabase.auth.setSession`, then routes to `/(auth)/welcome` → app.

Flow: sign up → "Check your email" → tap link → `oga://auth-callback` → session set → brand splash → onboarding/app.

## Not in this PR (required before the blocker is closed)

- **Dashboard:** add `oga://**` to the Supabase Auth Redirect URLs allow-list (dev now, prod before launch) + set a real Site URL. The code does nothing until the redirect is allow-listed.
- **Device verification:** needs an EAS dev/preview build to test the real `oga://` tap-through — Expo Go won't exercise the custom scheme. **Not verified on device yet.**

## Test plan

- [x] `npm run typecheck` (mobile) green.
- [ ] On an EAS dev build: sign up → confirm via email → app reopens signed-in.

### DIFF
--- a/apps/mobile/app/(auth)/signup.tsx
+++ b/apps/mobile/app/(auth)/signup.tsx
@@ -103,21 +103,12 @@ export default function Signup() {
         keyboardShouldPersistTaps="handled"
       >
       <View
-        style={{
-          backgroundColor: '#FFFFFF',
-          borderRadius: 10,
-          borderWidth: 0.5,
-          borderColor: '#E4E4E0',
-          padding: 20,
-        }}
+        className="border-oga-border bg-oga-bg-card"
+        style={{ borderRadius: 10, borderWidth: 0.5, padding: 20 }}
       >
         <Text
-          style={[TYPE.bodyBold, {
-            color: '#111111',
-            fontSize: 22,
-            fontWeight: '600',
-            marginBottom: 16,
-          }]}
+          style={[TYPE.bodyBold, { fontSize: 22, fontWeight: '600', marginBottom: 16 }]}
+          className="text-oga-text-primary"
         >
           Create your OGA account
         </Text>
@@ -126,6 +117,7 @@ export default function Signup() {
           autoCapitalize="none"
           value={username}
           onChangeText={setUsername}
+          className="bg-oga-bg-input border-oga-border text-oga-text-primary"
           style={inputStyle}
         />
         <FieldLabel>Email</FieldLabel>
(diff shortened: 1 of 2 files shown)


cner-smith (commented):
## Review — fix(mobile): deep-link email confirmation back into the app

**Concerns**

- `apps/mobile/app/auth-callback.tsx:43` — `handled.current = false` on `sessionError` is effectively dead code. The `useEffect` dependency is `[url]`; if `setSession` fails, `url` hasn't changed, so the effect will not re-run regardless of what `handled.current` is set to. The intended "allow a retry" behaviour silently doesn't work. The options are: remove the reset and accept that an error is terminal (user taps "Back to sign in" and re-requests confirmation), or parse the tokens out of the fragment into state so a retry button can call `setSession` again without needing the effect to re-trigger.

- `apps/mobile/app/auth-callback.tsx` (general) — `oga://` is a custom URI scheme. On Android, any other app can register the same scheme and intercept the deep link, receiving the `access_token` and `refresh_token` directly from the URL fragment. Supabase/GoTrue supports PKCE for mobile (`flowType: 'pkce'` in the client config + exchange a code rather than receiving tokens in the fragment), which is the recommended pattern for native clients and removes tokens from the redirect URL entirely. This is likely out of scope for this fix, but worth tracking before launch — the current implicit-flow approach is the weaker option for a custom scheme on Android.

**Nits**

- `apps/mobile/app/(auth)/signup.tsx:60–78` — The "check your email" card hardcodes colour values (`#FFFFFF`, `#E4E4E0`, `#111111`, `#555550`, `#0F6E56`) inline while the surrounding code in the same file uses `className` with the design system. Keeping the card consistent with NativeWind classes (or whatever tokens the rest of the screen uses) would be cleaner.

---
Looked good: single-concern PR, targets `dev`, no `any` introduced, no `@oga/core` exports touched, mobile dependency pins (including `nativewind 4.1.23`) untouched, `handled` ref correctly prevents double-processing the same URL, error path surfaces a message rather than swallowing it silently.

---

Issue #514, opened 2026-06-04
[mobile] Past-round logging has no scorecard-first companion — falls back to the live map UI


cner-smith:
## Problem (found in on-device testing)

Mobile "log a past round" feels indistinguishable from a live round: you can End the round early / Delete it, there's no Share button, and nothing signals it isn't live. Root cause is architectural, not a few missing buttons.

There are **two separate mobile round UIs**, and past-round entry routes into the wrong one:

| | `LiveRoundSession` (hole/map) | Scorecard view (`round/[id]/index.tsx`) |
|---|---|---|
| End-early / Delete | ✅ | ❌ |
| Share scorecard | ❌ | ✅ (unconditional) |
| Editable | ✅ (map state machine) | ❌ read-only |
| Feels like | live round | post-round scorecard |

- `new?mode=past` → **LiveRoundSession**. The *only* thing `mode=past` changes is skipping GPS auto-place (two `if (isPastMode) return` guards in `components/round/hole/useHoleState.ts`, ~L158 and ~L337). So past mode = live UI minus GPS.
- The Share button lives only on the **completed** scorecard view (`round/[id]/index.tsx`, header ~L354), which is reached only when `total_score !== null` — and it's **read-only**.
- Net: while logging a past round you get live affordances (End/Delete), no Share, and no "this is a past round" framing — and there is no editable scorecard logger at all.

## Goal — mirror the web post-round structure

Web's past-round entry opens `RoundDetailPage` → `ScorecardView`: scorecard-first (not map), per-hole **editable** shot entry, **Share scorecard**, **Delete**, finalize/Save-SG. No "End round early." (`apps/web/src/pages/rounds/RoundDetailPage.tsx`, `components/RoundHeader.tsx`, `components/ScorecardView.tsx`.)

Mobile should get the same: a scorecard-first **editable** past-round screen, distinct from the live map flow.

## Proposed implementation

1. **Route past entry to a scorecard screen, not LiveRoundSession.** `new?mode=past` (and opening an in-progress past round) lands on an editable scorecard screen — extend `round/[id]/index.tsx` (today read-only) into the editable post-round logger rather than redirecting `total_score == null` rounds into LiveRoundSession.
2. **Per-hole editing.** Tap a hole → enter score + shots (reuse the existing shot-entry pieces; make `PastHoleShotsSheet` editable). Mirror web's `ShotEntryModal` (with the mini-map shown for past entry).
[...]

PR #515, opened 2026-06-06
fix(mobile): show in-progress rounds in Home + All Rounds lists

cner-smith:
## Problem
In-progress (live) rounds never appeared in the Home or All Rounds lists, so they couldn't be reopened or resumed.

## Cause
Both lists load via `getRecentSGData`, which filters `.not('sg_total','is',null)`. A live round has `sg_total = null` until it's finalized, so it was filtered out entirely.

## Fix
Add `getRecentRounds` (identical select, **without** the SG filter) and point the two mobile list screens at it. The SG dashboard keeps `getRecentSGData` (it should only show rounds with computed SG). Both list UIs already render `—` for null score/SG, so unfinished rounds show cleanly and tapping resumes the live session.

Pre-existing live-round bug, independent of the past-round logger (#514).

### DIFF
--- a/apps/mobile/app/(app)/index.tsx
+++ b/apps/mobile/app/(app)/index.tsx
@@ -1,7 +1,7 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { Pressable, ScrollView, Text, View } from 'react-native'
 import { formatSG } from '@oga/core'
-import { deleteRound, getProfile, getRecentSGData } from '@oga/supabase'
+import { deleteRound, getProfile, getRecentRounds } from '@oga/supabase'
 import type { Database } from '@oga/supabase'
 import { supabase } from '../../lib/supabase'
 import { useAuth } from '../../hooks/useAuth'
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — fix(mobile): show in-progress rounds in Home + All Rounds lists

**Nits**
- `packages/supabase/src/queries/rounds.ts:82–86` — The block comment partially narrates the code: the first sentence ("Same shape as `getRecentSGData` but WITHOUT the `sg_total IS NOT NULL` filter") restates what the diff already shows — there is no `.not()` call. CONTRIBUTING.md: "Don't narrate what the code does." The non-obvious *why* worth keeping is that two near-identical functions are intentional, and which surface owns which. Something like: `// List views include in-progress rounds; the SG dashboard intentionally keeps getRecentSGData (only rounds with computed SG).` covers it in one line.

---
Looked good: narrowly scoped to one concern, correct package placement for the new query (`packages/supabase` not `apps/mobile/lib`), `export * from './queries/rounds'` picks up the new export automatically so no index wiring needed, callers already have the `active`-guard cancellation pattern and render `—` for null SG fields.

---

cner-smith:
Addressed in `de3837d`: tightened the `getRecentRounds` comment to the one-line *why* (no longer narrates the absent filter).

PR #517, opened 2026-06-06
feat(mobile): editable scorecard-first past-round logger (#514)

cner-smith:
## What
Mobile editable past-round logger (web-parity). Scorecard ⇄ Map tabs: the **map creates shots** (tap to place ball, aim, pin), the **scorecard edits details** (club / lie / slope / result + two-axis putt). Past-round entry uses a `total_score=0` sentinel (no migration); reuses `HoleMap` (the renderer), not the live GPS state machine. Save-SG + Delete + Share; no End-early.

## On-device QA fixes (final commit)
- **Tab bar** — reverted the band to its pre-regression values + ~8px bottom clearance (Galaxy S23 QA).
- **Duplicate-key on saved-round shot placement** — shots load after hole_scores, so the seed could lock an empty draft and block the re-seed → a placement INSERTed a colliding `shot_number`. Now adopts late-arriving shots while nothing's placed yet (loop-safe, no clobber).
- **Tee box** — two dots flanking the tee shot (derived from the first ball, oriented toward aim/pin) replace the oversized badge. No migration.
- **Green flow** — placing within putting range now shows the live round's *"On the green?"* prompt instead of auto-opening a detail sheet: **Yes** marks a putt (lie/club + derived distance) and advances to the next ball; **No** goes to aim. Putt details (made / short-long / left-right / break) are edited on the scorecard, exactly like web.

Refs #514. QA'd on device across SDK52 + Learn combined builds this session.

### DIFF
--- a/apps/mobile/app/(app)/_layout.tsx
+++ b/apps/mobile/app/(app)/_layout.tsx
@@ -149,16 +149,17 @@ export default function AppLayout() {
           backgroundColor: '#FBF8F1',
           borderTopWidth: 1,
           borderTopColor: '#D9D2BF',
-          paddingTop: 8,
           // Explicit height + paddingBottom from safe-area insets (#300).
           // Removing height entirely shrank the bar to 49 px on Android and
           // older iPhones (no home indicator), so we set the visible content
           // band ourselves and add `insets.bottom` for iPhone X+ clearance.
           // Android gesture/3-button nav frequently reports insets.bottom≈0,
-          // so the label crowds the screen edge — add a fixed Android-only
-          // clearance band (height grows in step so the content band holds).
+          // so the label crowds the screen edge — grow the bar and split the
+          // extra between top and bottom so the icon+label stay CENTERED
+          // (an earlier bottom-only bump shoved them upward).
+          paddingTop: 8 + (Platform.OS === 'android' ? 6 : 0),
           height: 54 + insets.bottom + (Platform.OS === 'android' ? 12 : 0),
-          paddingBottom: insets.bottom + 10 + (Platform.OS === 'android' ? 12 : 0),
+          paddingBottom: insets.bottom + 10 + (Platform.OS === 'android' ? 6 : 0),
           elevation: 0,
           shadowOpacity: 0,
         },
--- a/apps/mobile/app/(app)/_layout.tsx
+++ b/apps/mobile/app/(app)/_layout.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react'
-import { ActivityIndicator, Pressable, Text, View } from 'react-native'
+import { ActivityIndicator, Platform, Pressable, Text, View } from 'react-native'
 import { Tabs, Redirect } from 'expo-router'
 import { MaterialCommunityIcons } from '@expo/vector-icons'
 import { StatusBar } from 'expo-status-bar'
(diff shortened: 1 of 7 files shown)


cner-smith:
## Review — feat(mobile): editable scorecard-first past-round logger (#514)

**Blockers**
- `apps/mobile/components/round/PastRoundMap.tsx` — `handleGreenYes`: `handleAddShot()` is called unconditionally after the `if (shot?.id)` block. When `shot?.id` is falsy (shot not yet persisted to the DB), the lie-type update is skipped but the flow advances anyway — silently dropping the putt marker on that shot. Fix: return early when `!shot?.id`, or restructure so `handleAddShot()` only runs on a successful (or confirmed no-op) path.
- `apps/mobile/components/round/PastRoundMap.tsx` — `start: { lat: s.start_lat as number, lng: s.start_lng as number }` in the shot-mapping expression: CONTRIBUTING.md says "Don't add `any` to silence an error — fix the type." That rule applies equally to `as T` casts that paper over `number | null`. The preceding `.filter()` should be a type-predicate guard — e.g. `(s): s is typeof s & { start_lat: number; start_lng: number } => s.start_lat != null && s.start_lng != null` — so the compiler narrows correctly without the cast.

**Concerns**
- `apps/mobile/app/(app)/round/[id]/index.tsx` — `persistHoleScore`: `setHoleScores(nextList)` and `setRound` (total_score) are updated optimistically before `await updateRound(...)`. If the DB write fails, the UI shows a total that contradicts the database with no rollback. Either move state updates after a successful write, or hold the previous values and restore them on error.
- `apps/mobile/components/round/PastRoundMap.tsx` — `handleRemoveShot`: `await syncScore(...)` at the end of the function has no error path — `syncScore` swallows its DB error internally. A `hole_scores` update failure is invisible to the user. `syncScore` should at minimum return a success flag so callers can surface the failure.
- `apps/mobile/app/(app)/_layout.tsx` — the tab-bar height/padding adjustment (Galaxy S23 QA fix) is a separate concern from the past-round logger. CONTRIBUTING.md: "One concern per PR." The change is small, but the rule is explicit.

**Nits**
- `apps/mobile/components/round/PastRoundMap.tsx` — `placed` is in the `useEffect` dependency array while `setPlaced` is called inside the same effect. The `seededForRef.current` guard breaks the cycle today, but it's fragile — a future edit to the seeding key logic could silently reintroduce an infinite render loop. Worth a short inline comment on the ref guard explaining the invariant it must maintain.
- `apps/mobile/app/(app)/round/[id]/index.tsx` — `mode="live"` is now hardcoded, and the comment acknowledges that `LiveRoundSession`'s `mode`/`isPastMode` plumbing is dead. CONTRIBUTING.md: "ship the minimum." Dead props and their branching inside `LiveRoundSession` should be removed rather than deferred, unless there's a concrete reason for a follow-up PR.

---
Looked good: targets `dev` (not `main`), no new migration without an RLS policy, no new `@oga/core` exports without tests, no mobile dependency changes, `nativewind` untouched, no secrets in diff.

---

cner-smith:
Addressed in `490f84f`:
- `handleGreenYes` now returns early (with an alert) when the placed shot has no id — no more advancing-and-silently-dropping the putt mark on a failed placement.
- Seeding uses a type-predicate `.filter()` instead of the `as number` casts (compiler narrows `start_lat/lng`).
- `syncScore` surfaces its `hole_scores` update error (was swallowed).
- `persistHoleScore` writes `total_score` before reflecting it in state, so a failed write can't leave the header contradicting the DB.

On the tab-bar **one-concern** note: acknowledged, but keeping it in this PR — the regression was introduced by this branch's own `49a283b` (the past-round logger commit added the Android clearance bumps), so the fix belongs with it rather than a separate branch.

Issue #222, opened 2026-05-07
Security: tighten course_tees INSERT policy


cner-smith:
course_tees INSERT is open to any authenticated user. No created_by column.

Fix: add created_by column + tighten policy, or restrict to service_role like UPDATE.

Found in security audit, May 2026.

cner-smith:
## Validation 2026-05-22

**Verdict: real bug — confirmed.** Verified against `dev`.

### Where the hole lives

- `supabase/migrations/0008_course_tees.sql:39-41` — INSERT policy is still wide-open:
  ```sql
  create policy "Authenticated users can insert course tees"
    on public.course_tees for insert
    with check (auth.role() = 'authenticated');
  ```
- Migration `0014_course_security.sql` tightened **only the UPDATE** policy to `service_role` (lines 56-60). INSERT was never touched, and the table still has no `created_by` column.

### Threat shape

Any signed-in user can `INSERT` arbitrary `course_tees` rows. Each row contributes `course_rating` + `slope_rating` that feeds `rounds.score_differential` and therefore handicap math. A malicious or buggy client can spam rows or manufacture extreme rating/slope values that other users' rounds end up referencing (the FK is just `course_id` + a unique tee_color — no ownership check on read). The same threat shape as the UPDATE attack 0014 fixed; the INSERT side was the other half of the door.

### Reasoning for fix path

Two viable approaches, in order of safety:

1. **Lock INSERT to `service_role`** — mirrors the UPDATE policy. Crawler-only writes. Pro: trivially safe, matches 0014's posture. Con: blocks user-contributed tee data, which #252 explicitly wants to enable.
2. **Add `created_by uuid references auth.users`** and an `auth.uid() = created_by` check on INSERT, with UPDATE/DELETE restricted to creator + service_role. Pro: unblocks user contributions. Con: schema change + backfill, needs a write-rate-limit story to avoid spam.

Recommendation for a follow-up agent: ship #1 now (one migration file, four lines), and treat the user-contribution path as part of #251 / #252's design work. The two issues already plan a `hole_contributions` indirection — `course_tees` user writes should pass through the same review gate rather than landing directly.

### Suggested migration

```sql
-- supabase/migrations/0022_course_tees_insert_lockdown.sql
drop policy if exists "Authenticated users can insert course tees" on public.course_tees;

create policy "Service role can insert course tees"
  on public.course_tees for insert
  with check (auth.role() = 'service_role');
```

### Cross-references

[...]

PR #518, opened 2026-06-06
fix(db): tighten course_tees INSERT policy + add created_by (#222)

cner-smith:
## Problem
`course_tees` INSERT was open to any authenticated user with no attribution (0008). Security audit (#222).

## Fix
Add `created_by uuid` (FK→profiles, default `auth.uid()`) and restrict INSERT/UPDATE to **owner or service_role** (UPDATE was already service-role-only since 0014).

**No client change needed:** the web (`useCourses`) and mobile (`round/new`) upserts only run immediately after creating a *new* course (no pre-existing rows to conflict with, explicitly best-effort) and omit `created_by`, so the column default attributes them to the inserting user. The crawler (service_role, `auth.uid()` null) keeps full access.

⚠️ Needs `supabase db push` to **dev + prod** to apply (CI doesn't run migrations).

Refs #222.

### DIFF
--- a/supabase/migrations/0037_course_tees_created_by.sql
+++ b/supabase/migrations/0037_course_tees_created_by.sql
@@ -23,8 +23,17 @@ create policy "Users insert own course tees"
   with check (auth.role() = 'service_role' or auth.uid() = created_by);
 
 -- UPDATE: owner or service_role. Replaces the service_role-only policy from
--- 0014 so a user can still correct a tee they created.
+-- 0014 so a user can still correct a tee they created. The with check is
+-- spelled out (rather than left to Postgres's implicit reuse of using) so the
+-- post-update constraint is auditable at a glance and can't be silently
+-- relaxed by a later edit to using alone.
 drop policy if exists "Service role can update course tees" on public.course_tees;
 create policy "Users update own course tees"
   on public.course_tees for update
-  using (auth.role() = 'service_role' or auth.uid() = created_by);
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(db): tighten course_tees INSERT policy + add created_by (#222)

**Concerns**

- `supabase/migrations/0037_course_tees_created_by.sql:26-29` — The UPDATE policy uses `using` but omits `with check`. PostgreSQL implicitly reuses the `using` expression as the `with check` when none is specified, which is correct here — an authenticated user cannot change `created_by` to another user's UUID because the new row would fail `auth.uid() = created_by`. But the INSERT policy uses an explicit `with check`, and relying on the implicit default for UPDATE makes the security constraint harder to audit at a glance. An explicit `with check (auth.role() = 'service_role' or auth.uid() = created_by)` would be consistent and self-documenting.

- `supabase/migrations/0037_course_tees_created_by.sql` — No DELETE policy is added or replaced. With RLS enabled and no permissive DELETE policy, all deletes by non-service-role clients are denied by default — which may well be intentional for shared course data. Unclear from the diff whether a DELETE policy exists in 0008 or 0014. If owner-delete is deliberately excluded, a short comment to that effect would prevent a future reviewer from reading this as an oversight and adding one inadvertently.

---
Looked good: targets `dev`, migration correctly numbered 0037 after 0036, INSERT `with check` properly rejects spoofed `created_by` values, `on delete set null` is consistent with nullable `created_by` for crawler-inserted rows, scope is a single focused concern.

---

Issue #115, opened 2026-05-02
Verify Mapbox token is URL-restricted in dashboard


cner-smith:
apps/web/src/lib/mapbox.ts:4-7 — VITE_MAPBOX_TOKEN shipped in bundle (unavoidable for client-side maps). If not URL-restricted in Mapbox dashboard, quota theft is trivial.

Action: verify token in Mapbox dashboard is restricted to opengolfapp domains only. Document in DECISIONS.md.

cner-smith:
## Backfill — ticket profile 2026-05-22

**Three-system shape**

| Layer | Owns |
|---|---|
| `@oga/core` | none |
| `apps/web` | none |
| `apps/mobile` | none |
| `supabase/` | none — Mapbox dashboard task |

**Acceptance criteria**

- [ ] Mapbox public token (used by `apps/web`) is URL-restricted to `oga.golf` + `localhost:5173` only
- [ ] Restriction tested: token used from another origin returns 401/403
- [ ] Public token clearly separated from `MAPBOX_DOWNLOADS_TOKEN` (used in EAS env per #265)
- [ ] Dashboard configuration screenshotted + saved in ops notes
- [ ] If currently unrestricted: token rotated + restricted

**Out of scope**
- Migrating away from Mapbox (separate strategic question)
- Per-platform token isolation beyond public-vs-downloads
- Rate-limit alerts (separate ops ticket)

**Dependencies**
- Depends on: none
- Blocks: nothing critical (this is operational hygiene)
- Related: #265 (downloads token), Mapbox free-tier abuse risk

**Effort:** XS (~15 min in Mapbox dashboard)

**Verification**
- Open a private/incognito browser tab from a non-oga.golf URL; confirm Mapbox tiles fail to load with the restricted token
- Confirm web app continues working on oga.golf + localhost

**Risk + rollback**
- Risk: restriction too tight blocks legitimate dev / preview URLs
- Rollback: edit URL allowlist in Mapbox dashboard

---
_Generated by [Claude Code](https://claude.ai/code)_

cner-smith:
### Resolved — verified + hardened

**Audit of all three Mapbox tokens:**
- **Web** (`VITE_MAPBOX_TOKEN`, public `pk.`, ships in the JS bundle) — **URL-restricted to the web domains** ✅. This is the token that's publicly scrapeable from page source, so it's the one that matters most. This was #115's literal ask, and it's satisfied.
[...]

PR #526, opened 2026-06-09
docs(env): document web/mobile Mapbox token split (#115)

cner-smith:
## What
Documents the now-split Mapbox tokens in both `.env.example` files.

- **Web** (`VITE_MAPBOX_TOKEN`): public token, URL-restricted to the web domains (ships in the JS bundle).
- **Mobile** (`EXPO_PUBLIC_MAPBOX_TOKEN`): its **own** public token, **not** URL-restricted — native apps send no `Referer`, so a URL restriction would blank the map.
- **Secret** `sk.` download token stays an EAS secret (`MAPBOX_DOWNLOADS_TOKEN`), never in env files.

## Why
Web and mobile previously shared one URL-restricted token. A native app's no-referer requests against a URL-restricted token are the ambiguous case (risk of blank maps), and a single token muddies rotation/usage attribution. Tokens have since been split in EAS + local `.env` files; this just records the intent so the two aren't re-merged or the mobile one URL-restricted by mistake.

Docs only — no code, no behavior change.

Refs #115.

### DIFF
--- a/apps/mobile/.env.example
+++ b/apps/mobile/.env.example
@@ -2,5 +2,9 @@
 # 127.0.0.1 for emulator. Hosted project URL for non-local builds.
 EXPO_PUBLIC_SUPABASE_URL=
 EXPO_PUBLIC_SUPABASE_ANON_KEY=
+# Mobile's OWN public Mapbox token — separate from web's, and NOT URL-restricted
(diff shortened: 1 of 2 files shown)


cner-smith (commented):

---

PR #527, opened 2026-06-09
fix(db): targeted course dedup migration (#470)

cner-smith:
## What

Migration `0038_dedup_courses.sql` — a **targeted dedup** of true-duplicate `courses` rows. Addresses #470.

## Reframe vs the original issue

#470 was filed as "wipe and re-import." Triage on prod (2026-06-08) showed that's the wrong tool:

- The crawler upserts on `external_id` (migration 0019), so **`external_id` dup groups = 0** — a re-crawl would reproduce the exact same rows, not fix anything.
- The actual rot is **~238 rows (1.5%)** that are genuine duplicates: same normalized name *and* same ~1km location, mostly pre-`external_id` / synthetic rows.

So this is a surgical dedup, not a destructive re-import. Not launch-blocking.

## How it works

- **Cluster key:** `lower(btrim(name)) + round(lat,2) + round(lng,2)` (~1.1km). Conservative — under-merges rather than fusing two distinct courses that share a generic name in different towns. Re-runnable, so a second pass can mop up rounding-boundary cases.
- **Survivor per cluster** (in priority order): has `external_id` → most holes → has city+state → lowest id.
- **User rounds preserved:** any round on a loser is repointed to the survivor *before* deletion.
- **FK-safe:** `holes` / `course_tees` cascade on the loser delete; a loser whose tee is still cited by a round is guarded and left for manual follow-up (so the migration never aborts mid-run). On prod `rounds_on_losers = 0`, so the repoint is a no-op there — it exists to satisfy #470's "existing rounds still resolve to a valid course" criterion on dev / future data.

## Before applying

The file ships with a **manual preflight** (commented at the bottom): loser/cluster counts, the round-safety check, and a data-loss flag for any cluster whose survivor has fewer holes than a loser. Run it on dev first.

## Plan

Merge to `dev`, then apply to the dev DB via MCP `apply_migration` and run the preflight. Prod apply (numbered CLI `db push`) happens later, separately, after dev validation.

Refs #470 (closing keyword goes on the eventual dev→main release PR).

### DIFF
--- a/supabase/migrations/0038_dedup_courses.sql
+++ b/supabase/migrations/0038_dedup_courses.sql
@@ -1,0 +1,119 @@
+-- 0038_dedup_courses.sql  (#470)
+--
+-- Targeted dedup of TRUE duplicate course rows — NOT a wipe-and-reimport.
+-- Triage (prod, 2026-06-08) showed the crawler's upsert-on-external_id (0019)
+-- already prevents re-import duplication (external_id dup groups = 0), so a
+-- re-crawl would reproduce the same rows. The actual rot is ~238 rows (1.5%)
+-- that are genuine duplicates: same normalized name AND same ~1km location.
+--
+-- CLUSTER KEY: lower(btrim(name)) + round(lat,2) + round(lng,2).
+--   ~0.01deg ≈ 1.1km. Conservative on purpose — it will NOT merge two distinct
+--   courses that happen to share a generic name in different towns (those have
+--   different rounded coords). Known under-merge cases it won't catch:
+--     • the same course straddling a rounding boundary (35.004 vs 35.006)
+--     • spelling variants ("St. Andrews" vs "St Andrews")
+--   Under-merging is the safe direction (leaves a dup rather than fusing two
+--   real courses). Re-runnable, so a second pass can mop up boundary cases.
+--
+-- SURVIVOR per cluster (row_number rn=1), in priority order:
+--   1. has external_id   — keeps the crawler-canonical row so future crawls
+--                          converge instead of re-creating the dup
+--   2. most holes        — keeps the richest geometry
+--   3. has city + state  — most complete for search/display
+--   4. lowest id         — deterministic final tiebreak
+--   ⚠ POLICY DECISION: (1) before (2) means a cluster whose external_id'd row
+--     has FEWER holes than a loser will keep the sparser geometry. Run the
+--     "data-loss preflight" below before applying; if any cluster flags, decide
+--     per-cluster or flip the ordering (holes before external_id).
+--
+-- FK handling (delete rules verified on prod 2026-06-08):
+--   holes.course_id        → courses  CASCADE   (loser's holes auto-deleted)
+--   course_tees.course_id  → courses  CASCADE   (loser's tees auto-deleted)
+--   rounds.course_id       → courses  NO ACTION (blocks delete → repoint first)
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(db): targeted course dedup migration (#470)

**Concerns**

- `supabase/migrations/0038_dedup_courses.sql:57` — The correlated subquery `(select count(*) from public.holes h where h.course_id = c.id)` inside the window `ORDER BY` is a perf landmine. Because `ranked` has no side effects, PG 12+ inlines it by default — meaning every reference to `ranked` re-evaluates the full query, including this subquery. `clustered` references `ranked` twice (once as the source, once in `where k in (select k from ranked ...)`), so the hole-count subquery fires at least twice per course row on the full courses table. Even once is O(N_courses) index seeks into `holes`. The fix is a pre-aggregated CTE:
  ```sql
  with hole_counts as (
    select course_id, count(*) as n from public.holes group by course_id
  ),
  ranked as (
    select c.id, ...,
      row_number() over (
        ...
        order by (c.external_id is not null) desc,
                 coalesce(hc.n, 0) desc,
                 ...
      ) as rn
    from public.courses c
    left join hole_counts hc on hc.course_id = c.id
    where c.lat is not null and c.lng is not null
  )
  ```
  Adding `MATERIALIZED` to the `ranked` CTE would also suppress the double evaluation, but pre-aggregating is cleaner.

- `supabase/migrations/0038_dedup_courses.sql:74–77` — The `UPDATE` repoints `rounds.course_id` unconditionally for every round on any loser, including losers that the guard (lines 82–90) will leave undeleted. After the migration, those rounds have `course_id = survivor` but `course_tee_id = loser_tee`. No FK is violated — the loser still exists — but any query joining `rounds → course_tees` and filtering `course_tees.course_id = rounds.course_id` will silently see a mismatch. On prod this is a no-op (`rounds_on_losers = 0`), but on dev it leaves data that is FK-valid but semantically wrong. Consider scoping the UPDATE to only losers that will actually be deleted, mirroring the delete guard:
  ```sql
  update public.rounds r
  set course_id = m.survivor_id
  from _dedup_map m
  where r.course_id = m.loser_id
    and not exists (
      select 1
      from public.course_tees ct
      join public.rounds r2 on r2.course_tee_id = ct.id
      where ct.course_id = m.loser_id
    );
  ```

**Nits**

- `supabase/migrations/0038_dedup_courses.sql:97` (preflight query a) — `rounds_on_losers = 0` guarantees no violation of `rounds.course_tee_id → course_tees NO ACTION` (the guard handles it), but it doesn't directly verify `hole_scores.hole_id → holes NO ACTION`. If a loser's holes have `hole_scores` rows with no corresponding round on that loser (orphaned data, or a round where `course_tee_id IS NULL` bypasses the guard), the cascade delete aborts. `rounds_on_losers = 0` implies it under clean referential integrity, but adding a direct count to preflight (a) removes the assumption:
  ```sql
  (select count(*)
   from hole_scores hs
   join holes h on h.id = hs.hole_id
   where h.course_id in (select id from cl where rn > 1)
  ) hole_scores_on_loser_holes
  ```

---
Looked good: migration numbered correctly (0038 follows 0037), targets `dev`, no new tables so no RLS concern, survivor-priority logic and under-merge rationale are clearly documented, delete guard correctly prevents a mid-run abort for tee-cited losers, preflight (b) for the holes-sparser-than-loser case is a smart catch.

---

PR #529, opened 2026-06-09
feat(web): landing v2 — editorial sections, live-mode hero, realish seed

cner-smith:
## What

Landing-page v2 — an editorial redesign of the public marketing page (`/`), in the warm-editorial type system (Fraunces / Epilogue / Inconsolata, `caddie-*` tokens).

## Changes

- **New sections:** Manifesto, Community / open-source (3 contribution lanes), "What OGA isn't" (negation), and a "$0. Forever." pricing moment. 4-cell stat bar (+ MIT). Hero platform row (iPhone · Android · Web). **Learn section kept.**
- **Spreads:** the old 3-icon Features grid → editorial spreads — strokes-gained (animated diverging bars) and shot-patterns (a live demo dispersion cone), copy + figure alternating, fully responsive.
- **Hero `AppPreview` rebuilt to match the current mobile live mode** (`apps/mobile/HoleMapOverlays`): on-map `To Hole` / `Exp` HUD pills, the drag/mark hint, the left slope·pattern·pin toolbar — replacing the old header/footer bars. Points at a real, recognizable coastal hole pulled from the courses DB (no course name on the label).
- **Brand wordmark:** nav + footer now render the lowercase **`oga.`** mark (upright forest period), not all-caps OGA.
- **`seed-demo.ts`:** realistic per-club carry distances + biased per-club dispersion cones + 8 shot-rounds, so `/patterns` renders a real cone — and the e2e fixture gets meaningfully better coverage.

## Verification

- Typecheck clean (`pnpm typecheck`).
- Reviewed by frontend + UI agents; their findings folded in: corrected an inaccurate hardcoded course count, removed a duplicated hero/spread phone, fixed a dispersion-chart label collision + clarity, tightened a11y labels.
- Rendered + screenshotted live at every step (hero, spreads, full page).

Fully live — no baked screenshots; the dispersion + SG figures are hardcoded-demo React/SVG so the public page needs no auth or live data.

PR #531, opened 2026-06-09
chore(seed): realistic per-club dispersion for demo data

cner-smith:
## What

Make the demo / e2e user's seeded shots produce **realistic shot-pattern cones**, split out of the landing-v2 PR (#529) per "one concern per PR" — the landing page uses a hardcoded demo figure, not seed data, so this stands alone (it feeds the real authed `/patterns` page + the e2e fixture).

## Changes (`scripts/seed-demo.ts`)

- **Realistic club selection** — walk each hole hitting the club for the distance-to-pin and advance by its typical carry, so approaches land in iron/wedge ranges instead of every shot picking the same club.
- **Per-club biased dispersion** — driver wide + slice bias, wedges tight; sampled with a Box–Muller Gaussian (dense center, sparse tails) oriented to the shot's start→aim bearing, so `computeDispersion` fits a believable 68/95 cone.
- **8 shot-rounds (was 5)** so each common club clears the 5-shot floor the dispersion fit needs.

## Result

`/patterns` renders a real cone for the common clubs (verified on dev: 7i went from <5 shots to 23, with a believable short-bias/centered pattern), and the e2e fixture gets meaningfully better shot coverage. Box–Muller uses `1 - Math.random()` to guard `log(0)`.

### DIFF
--- a/scripts/seed-demo.ts
+++ b/scripts/seed-demo.ts
@@ -79,6 +79,98 @@ function pickClubForDistance(yards: number): string {
   return 'putter'
 }
 
+// Typical carry per club (yards). Used to walk a hole realistically: hit the
+// club for the distance-to-pin, advance by its carry, repeat — so approaches
+// land in iron/wedge ranges instead of every shot picking the same club.
+const TYPICAL_CARRY: Record<string, number> = {
+  driver: 255,
+  '3w': 225,
+  '5w': 210,
+  '4i': 190,
+  '5i': 178,
+  '6i': 165,
+  '7i': 152,
+  '8i': 140,
+  '9i': 128,
+  pw: 115,
+  gw: 95,
+  sw: 75,
+}
+
+// Per-club dispersion in yards, in the shot's own frame: biasLong (negative =
+// tends short, the amateur norm), biasLat (positive = push right), and the
+// 1σ spread on each axis. Short clubs are tight; driver is wide with a slice
+// bias — so the fitted 68/95 cones look like a real player's, not a blob.
+interface Disp {
+  biasLong: number
+  biasLat: number
+  sdLong: number
+  sdLat: number
+}
+const CLUB_DISPERSION: Record<string, Disp> = {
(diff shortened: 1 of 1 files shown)


cner-smith (commented):
## Review — chore(seed): realistic per-club dispersion for demo data

**Concerns**
- `scripts/seed-demo.ts` (non-last shot path, `distToPin = Math.max(8, ...)`) — `pickClubForDistance` falls through to `return 'putter'` as its default case (visible at the top of the diff at the function's closing line). If `distToPin` hits the 8-yard floor on any non-last shot, the insert ends up with `club: 'putter'`, `lie_type: 'fairway'`/`'rough'`, and `putt_result: null` — a record that can't exist in real play. Any query or stat that does `WHERE club = 'putter'` expecting only green/putt rows would be corrupted by this seed data. The old code used `Math.max(20, ...)` which raised the floor above the putter range; the new `Math.max(8, ...)` re-opens it. Raising the floor back to 20, or adding `if (club === 'putter' && !isLast) { club = 'sw' }` before the TYPICAL_CARRY lookup, would close this off.

**Nits**
- `scripts/seed-demo.ts` (`result` derivation) — The ±9 yd lateral and −11 yd longitudinal thresholds are club-agnostic. For wedges where `sdLat` is 4–5 yards, a ±9-yard miss is ~2.25σ (< 2.5% of shots), so `push_right`/`pull_left` effectively never appear for short clubs, and result is almost always `solid` or a flat 18% random `thin` regardless of the actual miss direction. The inline comment "Result follows the actual miss so result-based stats stay consistent" is accurate for driver and long irons but overstates the fidelity for short clubs. Not a blocker for seed data — just worth knowing if e2e tests ever assert result-type distributions by club.

---
Looked good: bearing-oriented geometry in `dispersedEnd` is correct (unit-vector decomposition, offset applied from aim in the shot's own frame); `1 - Math.random()` guard in Box–Muller is correct; `lie_slope` preserved on every shot; PR targets `dev`; single file changed, one concern.

---

Issue #309, opened 2026-05-13
iOS: produce App Store assets (icon, screenshots, optional preview video)


cner-smith:
**Severity:** blocker for submission (App Store Connect won't accept without)
**Audit source:** certification-planner agent, 2026-05-13

## Required assets

| Asset | Spec | Required? |
|---|---|---|
| App icon | 1024×1024 PNG, RGB only, no alpha, no rounded corners (Apple rounds it) | yes |
| Screenshots — 6.9\" (iPhone 16 Pro Max) | 1320×2868 | yes (required since iOS 18) |
| Screenshots — 6.7\" (iPhone 14/15 Pro Max) | 1290×2796 | yes |
| Screenshots — 6.5\" (older Plus models) | 1242×2688 | no longer required if 6.9\" present, but historically still expected |
| iPad screenshots — 12.9\" iPad Pro | 2048×2732 | only if iPad submission (currently \`supportsTablet: true\` so yes if we publish for iPad) |
| App preview video | 15–30s, .mov/.mp4 | optional but materially boosts conversion |
| iPhone 8 Plus 5.5\" | 1242×2208 | no longer required (deprecated 2024) — skip |

## Checklist

- [ ] Refine 1024×1024 app icon master (current asset is at \`apps/mobile/assets/icon.png\` — verify size and that it has no transparency)
- [ ] Take 5–10 screenshots in 6.9\" size: hero (live round map), strokes-gained chart, shot patterns, scorecard, profile
- [ ] Take 5–10 screenshots in 6.7\" size (same shots, scaled)
- [ ] Optional: produce 6.5\" set if we want maximum reach
- [ ] Decide on iPad submission. Currently \`supportsTablet: true\` — if we keep this, need iPad screenshots; if we don't actively want iPad support, set to \`false\` and skip iPad assets
- [ ] Optional: 15–30s app preview video showing the live-round flow

## Screenshot strategy (suggestions)

Top of mind, viewing from across the room:
1. **Hero shot** — live round map with shot breadcrumb and aim line, big headline overlay (\"Track every shot\")
2. **Strokes gained dashboard** — the SG breakdown by category
3. **Shot patterns** — dispersion ellipse on driving range / per-club analysis
4. **Practice plan** (when implemented) — AI-generated drill list
5. **Scorecard** — completed round summary

[...]

PR #532, opened 2026-06-09
chore(mobile): lock to iPhone-only for v1 (supportsTablet: false)

cner-smith:
Locks the v1 dual-store launch to **iPhone-only** by setting `ios.supportsTablet: false` in `app.config.ts`.

## Why
Drops the App Store's 12.9" iPad screenshot requirement (2048×2732) and iPad-layout QA from the v1 critical path. The app still installs on iPad in iPhone-compatibility mode. iPad support can be added in a later release.

## Scope
- One field flip + an explanatory comment. No code paths affected (`supportsTablet` is consumed only at native prebuild to set the iOS device-family capability).
- Part of the broader store-listing prep tracked in `docs/internal/store-submission.md`; closes the iPad portion of #309.

## Test plan
- Typecheck only — config-only change. iPad device-family behavior validates on the #376 device build.

Refs #309

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -43,7 +43,9 @@ const config: ExpoConfig = {
     // appVersionSource: 'remote' in eas.json, no seed = first build errors
     // because there's no remote value to increment.
     buildNumber: '1',
-    supportsTablet: true,
+    // iPhone-only for v1: drops the App Store's 12.9" iPad screenshot
+    // requirement and iPad-layout QA. Revisit when an iPad layout is on the roadmap.
+    supportsTablet: false,
     infoPlist: {
       // Foreground-only: OGA calls requestForegroundPermissionsAsync during a
       // round and never requests background/Always location. The expo-location


cner-smith (commented):
## Review — chore(mobile): iPhone-only for v1 — supportsTablet: false

**Nits**
- PR title — CONTRIBUTING.md requires "Title in imperative present tense" (example: "Add lie-slope filter to patterns"). This reads as a noun phrase. Something like "Lock app to iPhone-only for v1" satisfies the convention.
- `apps/mobile/app.config.ts:49` — The middle clause, "the app still installs on iPad in iPhone-compatibility mode," describes what the platform does rather than explaining why `supportsTablet` is `false` for v1. CONTRIBUTING.md: "Don't narrate what the code does." The surrounding clauses (screenshot-requirement drop, revisit note) already carry the why; this sentence can be dropped.

---
Looked good: config-only change, one concern, targets `dev`, no logic added outside `@oga/core`, no dependency or migration touched.

---

PR #533, opened 2026-06-09
fix(mobile): partial-course holes can't log shots (#525)

### LINKED ISSUE #525: Mobile: partially-mapped courses strand synthetic-padded holes with no hole_scores (can't log shots)
## Symptom (predicted from code; not yet reproduced on a real course)

On a course that has **some but not all 18** rows in the `holes` table, the holes that get client-side synthetic padding have **no `hole_scores` row**, so the player cannot log any shot on them — `persistShot` silently no-ops. The two-dot tee box / camera / distances also never persist for those holes.

## Mechanism

Round creation only materializes holes when the course has **zero** rows — it's all-or-nothing:

`apps/mobile/app/(app)/round/new.tsx:198-237`
```ts
let holes = existingHoles ?? []
if (holes.length === 0) {
  // insert_synthetic_hole RPC × 18 …
}
// hole_scores inserted only for `holes` (the existing real rows)
const holeScoreRows = holes.map((h) => ({ round_id, hole_id: h.id, score: 0 }))
```

So for a course with, say, 14 mapped holes: `holes.length === 0` is **false** → materialization is skipped → `hole_scores` is created for only those 14 holes.

`useHoleData` then pads the missing holes with synthetic ids (no DB row, no score):

`apps/mobile/components/round/hole/useHoleData.ts:65-87` → `id: synthetic-${round.id}-hole-${n}`, and `currentHoleScore = holeScores.find(hs => hs.hole_id === currentHole.id)` is `null` for a padded hole.

`buildPayload` bails without a score, so the shot never saves:

`apps/mobile/components/round/hole/useShotActions.ts:126`
```ts
if (!user || !currentHoleScore || !ball) return null
```

This matters because the course crawler averages **~14 holes/course** (126k holes across 8,885 courses), so partial coverage is common, not an edge case.

## Repro plan (not yet run)

1. Find/seed a course with a partial `holes` set (e.g. holes 1–14 only).
2. Start a live round there.
3. Navigate to hole 15 (a padded/synthetic hole) and try to log a shot → expected: nothing saves.

## Likely fix direction

Make materialization fill the **gaps**, not gate on `length === 0`: when `existingHoles.length < 18`, call `insert_synthetic_hole` for the missing numbers and insert `hole_scores` for the full 18. (Mirrors web's per-hole `ensureRealHole`.) Confirm against the `(course_id, number)` unique handling.

## Scope note

Pre-existing — **not** introduced by the live-round tee rework in #511 / PR #519. Surfaced during that PR's review while verifying `writeTee`'s synthetic-hole behavior (which is itself unreachable for the same reason: synthetic holes can't be logged on).

cner-smith:
Closes #525.

## Problem
Round creation (`apps/mobile/app/(app)/round/new.tsx`) only materialized `holes` + `hole_scores` when a course had **zero** `holes` rows (`if (holes.length === 0)`). A **partially-mapped** course — the crawler averages ~14 of 18 holes from OSM — got `hole_scores` for only its real holes. The synthetic display-padded holes had no `hole_scores` row, so `persistShot` silently no-opped (`useShotActions.ts` `if (!currentHoleScore) return null`) and the player **could not log shots** on the unmapped holes. Pre-existing; surfaced during #519 review.

## Data that shaped the fix
Queried prod (15,297 courses):
- **`course_tees.par` is ~0% populated** for mapped courses (10,694 of ~10,695 have no tees row) — the web's hole-count heuristic is unusable here.
- No course exceeds 18 holes; 1,588 courses have interior gaps; 682 don't start at hole 1.
- The only usable signal is the hole **numbers** themselves.

## Fix
- New pure `@oga/core` helper **`inferHoleCount(numbers): 9 | 18`** — empty → 18; `max(numbers) ≤ 9` → 9; else 18. Keyed on `max` (gap-immune), with tests.
- **`new.tsx`** — materialize whichever hole numbers in `1..expectedCount` are missing via the idempotent `insert_synthetic_hole` RPC (zero-case folds in), merge existing+created by number, create `hole_scores` for the full set.
- **`useHoleData.ts`** — replace hardcoded `expectedHoleCount = 18` with `inferHoleCount(...)`; expose `holeCount` so display padding matches creation.
- **`useShotActions.ts` + `MapBottomChrome.tsx`** — nav bounds, "Finish round" label, and Next-chevron disabled state now use `holeCount` instead of a hardcoded 18, so a 9-hole round navigates within bounds and finishes at 9.

Per the user requirement, genuine 9-hole courses are saved as 9, not coerced to 18.

## Verification
- `pnpm test` — 529 core tests pass (+8 new for `inferHoleCount`) ✅
- On-device validation pending (next iOS/Android QA build).

## Known limitations (deliberate)
- **No backfill** — this is a creation-time fix; rounds already created on a partial course keep their incomplete `hole_scores`. Small population; self-corrects on new rounds.
- **Ambiguous contiguous-9 courses** (2,045 in prod) are bucketed as 9-hole. A true front-9-of-18 mapping caps at 9 — but the inverse (coercing a real 9-hole to 18 phantom holes) is worse, so bias-to-9 is intentional.
- **`ShareableScorecardCard.tsx`** still renders 18 cells on the post-round **share image** (cosmetic, non-logging surface) — left as a follow-up.

### DIFF
--- a/apps/mobile/app/(app)/round/new.tsx
+++ b/apps/mobile/app/(app)/round/new.tsx
@@ -13,6 +13,7 @@ import * as Location from 'expo-location'
 import {
   formatLocation,
   getOpenGolfApiCourse,
+  inferHoleCount,
   searchOpenGolfApi,
   todayLocalDate,
   type OpenGolfApiSearchResult,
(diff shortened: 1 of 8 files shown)


cner-smith (commented):
## Review — fix(mobile): partial-course holes can't log shots (#525)

**Nits**
- `apps/mobile/components/round/hole/useHoleData.ts` (the `holeCount` `useMemo` comment) — the first three lines earn their place: they explain why `course_tees.par` can't be used and why the bias-to-9 is intentional, neither of which is obvious from the name or types. The last sentence ("Drives both the synthetic display padding below and the nav/finish bounds in useShotActions (consumed via the returned holeCount)") narrates callers; CONTRIBUTING.md says to avoid that. Drop it.

---
Looked good: `inferHoleCount` placed in `@oga/core` with comprehensive tests (8 cases covering empty, contiguous-9, back-nine-only, interior gaps), navigation bounds consistently updated across all three call sites (`useShotActions`, `ContextualActions`, `HoleNavPill`), `>= holeCount` on the nav-chevron guard is safer than the old `=== 18`, single concern held throughout, and the `ShareableScorecardCard` cosmetic gap is explicitly documented as a follow-up.

---

PR #536, opened 2026-06-09
feat: label handicap provenance — calculated vs entered (#521)

### LINKED ISSUE #521: [mobile+web] Surface whether handicap is a calculated value vs. entered
**From on-device QA (2026-06-08).**

Players can't tell when their displayed handicap is **derived** (calculated from round differentials via `@oga/core/handicap.ts`) vs. a **manually entered / onboarding** value. Surface the source.

- Show a small indicator/label next to the handicap (e.g. 'Calculated from your last N rounds' vs. 'Set during onboarding').
- Likely surfaces: AppBar handicap chip, profile/settings, anywhere the index is shown.
- Tie to the existing handicap pipeline (`calculateHandicapIndex` / `profiles.handicap_index`); needs a way to know provenance (may require a flag or inferring from round count).

cner-smith:
Closes #521.

## Problem
The displayed `handicap_index` could be either a value the player typed at onboarding/settings or one derived from their rounds — with no way to tell which. (Mobile had no signal at all; web had an "OFFICIAL" badge but on the wrong threshold.)

## Approach
The number of rounds with a non-null `score_differential` is the signal. `calculateHandicapIndex` needs **≥3** differentials to produce a value (below that, the stored index is still the entered one). New `@oga/core` `handicapProvenance(count)` centralizes that threshold so web + mobile agree.

- **`@oga/core`**: `handicapProvenance(differentialsCount) → 'calculated' | 'provisional'`, `HANDICAP_PROVENANCE_LABEL` (`{ calculated: 'WHS index', provisional: 'Provisional' }`), and `MIN_ROUNDS_FOR_HANDICAP = 3` (now also used inside `calculateHandicapIndex`, replacing a magic `< 3`). +5 tests, incl. a parity test pinning the provenance threshold to the calc minimum.
- **Web** (`useHandicapMeta` + `Sidebar`): the hook used `official: count > 0` — so a 1–2-round user got an "OFFICIAL" badge while their stored index was still entered. **Fixed to the core `≥3` helper** and relabelled to "WHS index" / "Provisional" (badge now shows in both states; tooltip explains each).
- **Mobile** (`profile.tsx`): adds the same badge under the index, fed by a `score_differential` count query (manual fetch — mobile has no react-query).

## Known / deliberate
- **Mobile never computes differentials** (only web's finalize flow writes `score_differential`), so a mobile-only player always reads "Provisional". That's accurate to current behavior; **mobile handicap recompute is a separate gap** worth its own ticket, not folded in here.
- Approach A (infer from count) over a new `handicap_source` column — chosen to avoid a migration; the only case it mislabels is a web user who manually re-enters *after* playing ≥3 rated rounds (narrow, low-harm).

## Verification
- `pnpm test` — 526 core tests pass (+5) ✅
- On-device + browser confirmation rides the next QA pass.

`merged-to-dev` on merge; closes on the `dev→main` release.

### DIFF
--- a/apps/mobile/app/(app)/profile.tsx
+++ b/apps/mobile/app/(app)/profile.tsx
@@ -145,7 +145,11 @@ export default function ProfileTab() {
       .select('id', { count: 'exact', head: true })
       .eq('user_id', user.id)
       .not('score_differential', 'is', null)
-      .then(({ count }) => {
+      .then(({ count, error }) => {
+        if (error) {
+          // eslint-disable-next-line no-console
+          console.warn('[profile/differentials-count]', error.message)
+        }
         if (active) setDifferentialsCount(count ?? 0)
       })
     return () => {
@@ -196,6 +200,9 @@ export default function ProfileTab() {
     )
   }
 
+  const provenance = handicapProvenance(differentialsCount)
+  const provenanceCalculated = provenance === 'calculated'
+
   return (
     <View style={{ flex: 1, backgroundColor: '#F2EEE5' }}>
       <AppBar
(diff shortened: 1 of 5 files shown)


cner-smith (commented):
## Review — feat: label handicap provenance — calculated vs entered (#521)

**Concerns**
- `apps/mobile/app/(app)/profile.tsx:148` — The new query chains `.then(({ count }) => ...)` without destructuring `error`. When Supabase returns an error (network drop, RLS rejection, auth expiry), `count` is `null` and `?? 0` gives `0` — so the component silently shows "Provisional" with no signal that anything went wrong. The fallback behavior is conservative rather than incorrect, but the error is fully swallowed. At minimum, a `console.warn` on the `error` path would give future debugging something to work with.

**Nits**
- `apps/mobile/app/(app)/profile.tsx:230` — The IIFE inside JSX (`(() => { const provenance = ...; return <View>...</View> })()`) works, but `provenance` (and the derived `calculated` flag) could just be top-of-body `const`s derived from `differentialsCount`, placed alongside the other state declarations. That flattens the render output and follows the pattern used everywhere else in the component.
- `apps/web/src/components/layout/Sidebar.tsx:207` — The `differentialsCount === 1 ? '' : 's'` ternary inside the `'calculated'` tooltip arm is unreachable: `provenance === 'calculated'` is only true when `differentialsCount >= MIN_ROUNDS_FOR_HANDICAP` (3), so the singular branch can never fire here. Worth dropping the ternary and hardcoding `'s'`, or a brief note for the next reader.

---
Looked good: `handicapProvenance`, `HANDICAP_PROVENANCE_LABEL`, and `MIN_ROUNDS_FOR_HANDICAP` all land in `@oga/core` with tests; the parity test pins the provenance threshold directly to `calculateHandicapIndex`'s behavior; the web bug (1-round "OFFICIAL" badge) is fixed; PR targets `dev`; single concern throughout.

---

PR #537, opened 2026-06-09
feat: call out the standout SG category on stats pages (#522)

### LINKED ISSUE #522: [mobile+web] Stats page: call out standout stats (costing you / strength)
**From on-device QA (2026-06-08).**

On the Stats page, proactively flag standout numbers instead of leaving the player to interpret:

- 'This is costing you ~X strokes — your biggest leak.' (worst SG category / metric)
- 'This is a strength — keep doing it.' (best SG category / metric)

The Practice plan already does this prose framing from the same SG data (`computeDetailedStats` / SG averages); reuse that logic to annotate the stats surfaces. Keep it skill-calibrated (compare to the player's handicap bracket baseline, not scratch).

cner-smith:
Closes #522.

## Problem
The stats pages showed four strokes-gained category averages but never said which was the **leak** or the **strength** — the player had to eyeball four numbers.

## Fix
New `@oga/core` `sgStandouts(sg)` picks the lowest (biggest leak) and highest (biggest strength) category from the per-round SG averages — null-safe, sign left to the caller, weakest===strongest when only one category has data. Pure + tested (+5).

A one-line prose callout renders on both surfaces, calibrated to the same handicap-bracket SG the tiles already use:
- **Web** `StrokesGainedSection`: italic note above the SG tiles ("Your biggest leak is **Approach** — -0.8 a round. **Off tee** is a strength, +0.4 a round.").
- **Mobile** `stats.tsx`: same prose in a left-border note above the averages, using a label map kept identical to web's `SG_SERIES`.

Copy adapts to sign: "biggest leak" when the weakest category is negative, "softest area" when it's still positive; the strength clause only shows when it's a distinct, positive category.

## Verification
- `pnpm test` — 531 core tests (+5 `sgStandouts`) ✅

`merged-to-dev` on merge; closes on the `dev→main` release.

### DIFF
--- a/apps/mobile/app/(app)/stats.tsx
+++ b/apps/mobile/app/(app)/stats.tsx
@@ -194,45 +194,7 @@ export default function Stats() {
           </View>
         ) : (
           <>
-            {stats &&
-              (() => {
-                const { weakest, strongest } = sgStandouts(stats.sg)
-                if (!weakest) return null
-                const showStrength =
-                  strongest != null &&
-                  strongest.key !== weakest.key &&
-                  strongest.value > 0
-                const leak =
-                  weakest.value < 0
-                    ? `Your biggest leak is ${SG_STANDOUT_LABEL[weakest.key]} — ${formatSG(weakest.value)} a round.`
-                    : `Your softest area is ${SG_STANDOUT_LABEL[weakest.key]} (${formatSG(weakest.value)} a round).`
-                const strength = showStrength
-                  ? ` ${SG_STANDOUT_LABEL[strongest.key]} is a strength, ${formatSG(strongest.value)} a round.`
-                  : ''
-                return (
-                  <View
-                    style={{
-                      borderLeftWidth: 2,
-                      borderColor: '#1F3D2C',
-                      paddingLeft: 14,
-                      marginBottom: 22,
-                    }}
-                  >
-                    <Text
-                      style={[TYPE.serif, {
-                        color: '#1C211C',
-                        fontSize: 17,
-                        fontStyle: 'italic',
-                        fontWeight: '500',
-                        lineHeight: 24,
-                      }]}
-                    >
-                      {leak}
-                      {strength}
-                    </Text>
-                  </View>
-                )
-              })()}
+            {stats && <StandoutCallout sg={stats.sg} />}
 
             <Section kicker={`Avg — last ${rounds.length} rounds`}>
               <View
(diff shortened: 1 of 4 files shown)


cner-smith:
## Review — feat: call out the standout SG category on stats pages (#522)

**Concerns**
- `packages/core/src/stats.ts:143–156` — The block comment above `sgStandouts` narrates behaviour already documented by the `SGStandouts` interface JSDoc (`weakest: Lowest-scoring…`, `strongest: Highest-scoring…`) and readable from the implementation. CONTRIBUTING.md: "Don't narrate what the code does." The only line that carries non-obvious information is the single-category invariant (`weakest === strongest when exactly one category has data`); everything else, including the `#522` reference, belongs in the PR description rather than source. Trim to that one line or drop the block entirely.

- `apps/mobile/app/(app)/stats.tsx:197–230` — The callout is computed inside an IIFE invoked inline in JSX. The web component does this correctly: `const standouts = sgStandouts(data.sg)` above the `return`, then `{standouts.weakest && (…)}` in JSX. The IIFE hides a `return null` inside the render tree, allocates a new closure on every render, and makes the control flow harder to follow. Bring the mobile in line with the web pattern — assign above the `return`, guard in JSX.

**Nits**
- `apps/mobile/app/(app)/stats.tsx:24–30` — `SG_STANDOUT_LABEL` is a separate hardcoded record; the web derives the same strings from `SG_SERIES`. The PR description says they're "kept identical" but that's a manual invariant with no enforcement. If a label in `SG_SERIES` ever changes this copy won't follow. Risk is low while `SGAverages` keys are stable, but worth leaving a note so the next SG change doesn't miss it.

- `apps/web/src/pages/stats/sections/StrokesGainedSection.tsx:76` — `sgLabel` is extracted as a local helper with 2 call sites. CONTRIBUTING.md's [3-caller rule](https://github.com/cner-smith/opengolfapp/blob/61b04be19d7af6f8a9f4f4c9dfe1da737eca29d0/CONTRIBUTING.md#3-caller-rule-for-extraction): "Don't extract a helper… unless it has three or more callers, or the inline version is genuinely unreadable." The inline form (`SG_SERIES.find(s => s.key === key)?.label ?? key`) is wordy but not opaque. If a third caller doesn't appear, consider inlining.

---
Looked good: `sgStandouts` correctly placed in `@oga/core`, five boundary tests covering all-null / all-positive / single-category / mixed, sign left to the caller, PR targets `dev`, single concern well-held.

---

PR #538, opened 2026-06-09
feat(mobile): handicap differential + index recompute (rated tees)

cner-smith:
Builds on #536 (#521). Surfaced there: **mobile never recomputes a handicap** — it records no played tee and never computes a `score_differential`, so a mobile player's `handicap_index` stays the entered value forever ("Provisional"). All WHS calculation was web-only. Data confirms it's dormant app-wide: 0 of 11 prod rounds have a differential, and only 1 of 15,297 courses has a rated tee.

## Root cause
A WHS differential needs the played tee's **course rating + slope**, which the crawler (OpenGolfAPI/OSM) doesn't carry. Web has a `TeeSelector` to record one; mobile had nothing.

## Fix
**Capture** (`RoundTeeSelector`, new) — on the round detail screen, pick an existing `course_tees` row or add one (colour / rating / slope / yards), persisting `course_tee_id` + `tee_color` on the round. Mirrors web's `TeeSelector`; the 0037 RLS allows the insert (`created_by` defaults to `auth.uid()`).

**Compute** (`lib/completeRound.ts`) — both finalize paths (live end-round + past Save-SG) funnel through `completeRound()`, so the recompute lives there once, mirroring web's `useCompleteRound`:
1. Resolve the round's tee (by id, else colour).
2. If it has rating + slope, compute the WHS differential from the ESC-adjusted gross; stamp `score_differential` + `course_tee_id` on the round.
3. Recompute the index from the most recent 20 differentials; write the profile.

Reuses `@oga/core` math (`adjustedScore` / `calculateDifferential` / `calculateHandicapIndex`) — no new core logic. Differential stays `null` when no rated tee is recorded (the common case), so behavior is unchanged for courses without tee data; the index only becomes a calculated "WHS index" once ≥3 rated rounds exist.

## Verification
- On-device validation needed (tee add/select → Save SG → differential written → profile index flips from "Provisional" to "WHS index" at the 3rd rated round). Rides the next QA build.

## Note
Pairs with #536's provenance label. With this merged, a mobile-primary player can finally reach a calculated index — closing the gap that PR flagged.

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -30,6 +30,7 @@ import type { Database } from '@oga/supabase'
 import { supabase } from '../../../../lib/supabase'
 import { completeRound } from '../../../../lib/completeRound'
 import { ShareableScorecardCard } from '../../../../components/round/ShareableScorecardCard'
+import { RoundTeeSelector } from '../../../../components/round/RoundTeeSelector'
 import { PastHoleShotsSheet } from '../../../../components/round/PastHoleShotsSheet'
 import { PastRoundMap } from '../../../../components/round/PastRoundMap'
 import { ScoreCell, TabSwitcher } from '../../../../components/round/PastScorecardParts'
@@ -816,6 +817,20 @@ export default function RoundIndex() {
             )
           })}
         </View>
+
+        {user && (
+          <RoundTeeSelector
+            courseId={round.course_id}
+            roundId={round.id}
+            userId={user.id}
+            currentTeeId={round.course_tee_id}
+            onChange={(tee) =>
+              setRound((r) =>
+                r ? { ...r, course_tee_id: tee.id, tee_color: tee.color } : r,
+              )
+            }
+          />
+        )}
 
         <Pressable
           accessibilityRole="button"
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — feat(mobile): handicap differential + index recompute (rated tees)

**Concerns**
- `apps/mobile/components/round/RoundTeeSelector.tsx:52-59` — `getCourseTees(...).then(...)` has no rejection handler. If the network call fails, `setLoading(false)` is never called and the component stays on "Loading tees…" with no error message and no way to recover. Also generates an unhandled-promise-rejection warning in the Expo runtime. Fix: chain `.catch(() => { if (active) { setTees([]); setLoading(false) } })`, or convert to async/await with try/catch in the effect body.

**Nits**
- `apps/mobile/components/round/RoundTeeSelector.tsx:271` — `width: width as \`${number}%\`` asserts a `string` into a template-literal type rather than expressing the correct type in the first place. Typing the prop as `ViewStyle['width']` (from `react-native`) removes the need for the assertion and lets RN's style system enforce valid values at the callsite.

---
Looked good: WHS math entirely delegated to existing `@oga/core` exports — no logic reimplemented in `apps/mobile/lib`; no new `@oga/core` exports so no new tests required; `nativewind` unchanged at `4.1.23`; PR targets `dev`; scope held tightly to one concern.

---

PR #539, opened 2026-06-09
chore(scripts): course rating/slope backfill from GolfCourseAPI

cner-smith:
Populates `course_tees.course_rating` / `slope_rating` — the data OSM and our crawler can't supply, and the reason the WHS handicap engine is dormant (0 rated courses in prod). Pairs with the mobile-handicap recompute (#538) and the provenance label (#536): once courses are rated, finalized rounds actually produce a calculated index.

## What it does
`pnpm backfill:ratings` — matches our `courses` to GolfCourseAPI by **normalized name + location**, pulls the rated tees, and upserts them into `course_tees` (men's + women's merged; men's wins a `tee_color` collision since the table has no gender column).

## Safety (this writes handicap-grade data, so it's deliberately careful)
- **Dry-run by default** — writes nothing without `--write`; logs every proposed match to a gitignored JSONL audit file to review first.
- **Coordinates are authoritative** — a match needs the two courses within `--max-km` (default 25); same-state is only a fallback when coords are missing. *A dry-run caught two different "Bayou Bend" courses 284 km apart in LA that a state-only gate would have mismatched* — fixed before this PR.
- **Resumable** — skips courses that already have a rated tee, so you can run in rate-limit-sized batches (`--limit`).
- **Key never committed** — read from `GOLFCOURSEAPI_KEY` env only; verified absent from all tracked files.

Flags: `--write --limit N --delay-ms N --threshold 0–1 --max-km N --has-state --out path.jsonl`.

## Validation (dry-run vs prod courses)
Real CR/slope come through correctly, e.g. *River Hills CC, Corpus Christi → black 69.8/127* (exact city, 0 km). ~58% match rate on a US-state sample; unmatched courses fall back to the app's player-entered tee (PR #538).

## Run plan (for when the paid tier is active)
1. `pnpm backfill:ratings --has-state --limit 200` (dry-run) → review `backfill-ratings-matches.jsonl`.
2. Tune `--threshold` / `--max-km` if needed.
3. `pnpm backfill:ratings --write --has-state --limit <daily cap>` in batches until done.

Note: GolfCourseAPI is a commercial API; confirm its terms permit storing the values in our DB before the full `--write` pass. Scripts aren't in the CI typecheck (matches existing crawler convention); validated by running under tsx.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,6 @@ PHASE_2.md
 PHASES_3_TO_8.md
 docs/internal/
 docs/superpowers/
+
+# course-rating backfill audit log
+backfill-ratings-*.jsonl
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "seed:my-data": "tsx scripts/seed-my-data.ts",
     "import:osm": "tsx scripts/import-osm-course.ts",
     "crawl:courses": "tsx scripts/crawl-courses.ts",
+    "backfill:ratings": "tsx scripts/backfill-ratings.ts",
     "email:vendor-core": "mkdir -p supabase/functions/_shared && cp packages/core/src/round-summary.ts supabase/functions/_shared/round-summary.ts",
     "vendor:practice-plan": "node scripts/vendor-practice-plan.mjs"
   },
--- a/scripts/backfill-ratings.ts
+++ b/scripts/backfill-ratings.ts
@@ -234,7 +234,16 @@ function ratedTeeRows(courseId: string, api: ApiCourse): TeeRow[] {
   const byColor = new Map<string, TeeRow>()
   const add = (tees: TeeBox[] | null | undefined, isMale: boolean) => {
     for (const t of tees ?? []) {
-      if (t.course_rating == null || t.slope_rating == null || t.slope_rating <= 0) continue
+      // Both must be present AND positive — a 0.0 rating/slope from the API
+      // is invalid for any real course and would write a handicap-breaking
+      // zero.
+      if (
+        t.course_rating == null ||
+        t.course_rating <= 0 ||
+        t.slope_rating == null ||
+        t.slope_rating <= 0
+      )
+        continue
       const name = (t.tee_name ?? '').trim()
       if (!name) continue
       const color = name.toLowerCase()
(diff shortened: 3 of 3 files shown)


cner-smith (commented):
## Review — chore(scripts): course rating/slope backfill from GolfCourseAPI

**Concerns**

- `scripts/backfill-ratings.ts:~346` — `appendFileSync` writes the audit record before the `if (WRITE)` block, so a failed upsert (console-logged at ~356) leaves a record in the JSONL that is indistinguishable from a successful proposal. In the intended dry-run-first workflow the log is reviewed before `--write`, so it doesn't break anything, but after a `--write` run with partial failures the file can mislead an operator auditing what actually landed. Marking failed records with a `"upsert_error": true` field would close that gap.

- `scripts/backfill-ratings.ts:~346` (same block) — GolfCourseAPI is a commercial API and the PR description correctly calls out "confirm its terms permit storing the values in our DB before the full `--write` pass." Worth being explicit: OGA's MIT license and free-forever pillar mean any ToS clause that restricts redistribution or requires attribution of stored data would need resolution before production use. This is less a code issue and more a gate that belongs clearly in the run plan before step 3.

**Nits**

- `scripts/backfill-ratings.ts:~238` — `slope_rating <= 0` is guarded but `course_rating` only has a null check. A CR of `0.0` from the API (invalid for any real course) would pass through and write a handicap-breaking zero rating. Adding `|| t.course_rating <= 0` alongside the slope guard makes the validation symmetric.

- `scripts/backfill-ratings.ts:~348,~356` — `✓` and `✗` in console output. CONTRIBUTING.md: "No emojis in source code or commit messages unless we're laying down user-facing copy." ASCII alternatives (`[ok]` / `[FAIL]`) satisfy the rule.

- `bestMatch` selects by name similarity score only; when two candidates tie on score, the first API result wins regardless of km distance. In practice floating-point equality of similarity scores is rare, and the dry-run audit is the real safety net, but preferring the lower-km candidate on a tie would be strictly safer for handicap data.

---
Looked good: API key exclusively from env with an early-exit guard, coordinate-over-state matching with a clear comment explaining the Bayou Bend catch, dry-run-by-default design, audit JSONL gitignored, `--write` gate, target branch is `dev`.

---

Issue #528, opened 2026-06-09
Mobile: My Bag — "Add to bag" button clipped at bottom (missing safe-area inset)


cner-smith:
**Observed (mobile):** On the My Bag screen, the "Add to bag" / "Add club" button is clipped a little at the bottom of the screen.

**Repro:** Profile → My bag, on a device with a gesture nav / home indicator.

**Likely cause:** `apps/mobile/app/(app)/bag.tsx` hardcodes bottom padding instead of using `useSafeAreaInsets().bottom`, so the bottom edge sits under the home indicator. Two candidate spots:
- Footer button block — `DraggableFlatList contentContainerStyle={{ padding: 18, paddingBottom: 40 }}` wrapping the **"Add club →"** Pressable (~line 264/292).
- The "Add club" modal — a `justifyContent: 'flex-end'` bottom-sheet whose `ScrollView contentContainerStyle={{ padding: 18, paddingBottom: 28 }}` ends in the **"Add to bag →"** save button (~line 350/498). A flex-end sheet with a fixed 28px bottom pad is the usual home-indicator clip.

**Proposed fix:** add `const insets = useSafeAreaInsets()` and use `paddingBottom: insets.bottom + <n>` on both the list footer and the modal sheet, per the CLAUDE.md safe-area rule (never hardcode inset padding; `AppBar.tsx` `insets.top + 14` is the reference).

Label: `bug` (observed user-visible symptom). Low risk; cosmetic-but-real.

cner-smith:
Shipped in the 0.9.0 release (PR #552, via #541) — now live on `main`/prod. Closing as done.

PR #541, opened 2026-06-10
fix(mobile): bound My Bag list height so footer is reachable (#528)

cner-smith:
## Problem (on-device, iPhone)

The My Bag screen still cuts off at the bottom — the **Reset to default bag** option isn't reachable. The #528 inset fix (`5351d9e`) added `paddingBottom: insets.bottom + 24` but that only pads the scroll *content*; it didn't address the real cause.

## Root cause

The `DraggableFlatList` had no bounded height (no `containerStyle`). As the last child of a `flex:1` root next to the `AppBar`, it sizes to its content, so when the bag is full the footer (Add club / Reset to default bag) is pushed past the viewport with no scroll. `react-native-draggable-flatlist` requires a bounded-height container.

## Fix

One line: `containerStyle={{ flex: 1 }}` on the list, so it fills the remaining space and scrolls to the footer.

## Test
- `npm run typecheck` clean.
- On-device QA pending (the symptom was found on iPhone).

### DIFF
--- a/apps/mobile/app/(app)/bag.tsx
+++ b/apps/mobile/app/(app)/bag.tsx
@@ -263,6 +263,11 @@ export default function BagScreen() {
           data={bag}
           keyExtractor={(c) => c.id}
           onDragEnd={({ data }) => onDragEnd(data)}
+          // The list is the last child of a flex:1 root next to the AppBar.
+          // Without a bounded height it sizes to content and pushes the
+          // footer (Add club / Reset to default bag) off-screen with no
+          // scroll — react-native-draggable-flatlist needs flex:1 here (#528).
+          containerStyle={{ flex: 1 }}
           contentContainerStyle={{ padding: 18, paddingBottom: insets.bottom + 24 }}
           ListHeaderComponent={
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(mobile): bound My Bag list height so footer is reachable (#528)

**Concerns**
- The PR description notes "On-device QA pending." For a layout fix whose failure mode is purely visual and can't be caught by `typecheck` or static review, landing without at least one successful device run risks trading one layout break for another. Worth holding until that's confirmed.

**Nits**
- `apps/mobile/app/(app)/bag.tsx:266–269` — The comment narrates the pre-fix behavior across four lines ("it sizes to content and pushes the footer off-screen with no scroll"), which CONTRIBUTING.md flags: "Don't narrate what the code does." The non-obvious part is the library requirement; one line covers it: `// react-native-draggable-flatlist needs a bounded container height to scroll (#528)`.

---
Looked good: targets `dev`, minimal one-line fix, no dependency changes, no logic in the wrong package, no type suppressions.

---

PR #543, opened 2026-06-10
feat(mobile): tee selection at round creation (#542)

### LINKED ISSUE #542: [mobile] Live-round flow has no tee selection — only the scorecard does
## Gap (on-device, iPhone QA)

When tracking a **live** round on mobile there's no way to pick the tee you're playing. Web surfaces a `TeeSelector` at round creation (`NewRoundPage`); mobile only mounts `RoundTeeSelector` in the scorecard view (`round/[id]/index.tsx`, the past-round companion). The live flow (`LiveRoundSession`) and the course-pick screen (`round/new.tsx`) never show it.

This matters now that `course_tees` ratings/slopes are backfilled (#539) — the tee played is the one input WHS differentials need (#538), and a live-round player currently can't supply it without detouring to the scorecard.

## Fix

Mirror web: surface tee selection **at round creation**. Mobile's create flow is one-tap (pick course → round starts), so add a lightweight setup step after course selection with the tee picker (optional — start without it to keep pace of play), threading `course_tee_id` + `tee_color` into the round insert.

Extract the presentational picker out of `RoundTeeSelector` so the scorecard and the new setup step share it.

cner-smith:
## Problem (on-device, iPhone QA)

When tracking a **live** round on mobile there's no way to pick the tee you're playing. Web surfaces a tee picker at round creation (`NewRoundPage`); mobile only mounted `RoundTeeSelector` in the scorecard view (the past-round companion). The live flow and the course-pick screen never showed it.

This matters now that `course_tees` ratings/slopes are backfilled (#539) — the tee played is the one input WHS differentials need (#538), and a live-round player couldn't supply it without detouring to the scorecard.

## Approach — mirror web, tee at creation

Mobile's create flow is one-tap (pick course → round starts), so there was no place for a picker. Added a lightweight **setup step** after the course resolves and before the round is created:

- **`TeePicker`** — extracted the presentational tee list + add-tee form out of `RoundTeeSelector` (selection-only, no round write). The scorecard wrapper now consumes it too, so both surfaces share one implementation instead of duplicating ~150 lines.
- **`RoundSetupStep`** in `round/new.tsx` — shows the course name + the picker, then Back / Start. **Tee is optional** so it never blocks pace of play; `course_tee_id` + `tee_color` only get written when a tee is picked, and the scorecard can still set it later. All four course-selection paths (already-imported, OpenGolfAPI existing/new, manual add) route through it.

## Test
- `npm run typecheck` clean.
- The scorecard `RoundTeeSelector` is unchanged in behavior (now a thin wrapper over `TeePicker`).
- On-device QA pending.

Closes #542

### DIFF
--- a/apps/mobile/app/(app)/round/new.tsx
+++ b/apps/mobile/app/(app)/round/new.tsx
@@ -30,6 +30,7 @@ import type { Database } from '@oga/supabase'
 import { supabase } from '../../../lib/supabase'
 import { useAuth } from '../../../hooks/useAuth'
 import { FONT, TYPE } from '../../../lib/typography'
+import { TeePicker } from '../../../components/round/RoundTeeSelector'
 
 type CourseRow = Database['public']['Tables']['courses']['Row']
 type HoleInsert = Database['public']['Tables']['holes']['Insert']
@@ -66,6 +67,13 @@ export default function NewRound() {
   const [busy, setBusy] = useState(false)
   const [error, setError] = useState<string | null>(null)
   const [showManualForm, setShowManualForm] = useState(false)
+  // Once a course is resolved we step to a lightweight setup view (tee pick)
+  // before creating the round — mirrors web's NewRoundPage, where tee is an
+  // up-front field. Null = still on the course-search screen.
+  const [pendingCourse, setPendingCourse] = useState<{
+    id: string
+    name: string
+  } | null>(null)
   const [gps, setGps] = useState<GpsState>({ status: 'idle' })
   const searchAbort = useRef<AbortController | null>(null)
   const insets = useSafeAreaInsets()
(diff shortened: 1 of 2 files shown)


cner-smith (commented):
## Review — feat(mobile): tee selection at round creation (#542)

**Concerns**
- `apps/mobile/app/(app)/round/new.tsx:391` — `error` state is set in `startWith`'s catch block but only rendered in the main `NewRound` return path. When `pendingCourse !== null` the component returns `<RoundSetupStep>` instead, so any failure at the "Start round" tap (network error, DB constraint, etc.) silently calls `setError(...)`, un-grays the button, and shows the player nothing. They have no signal that the round wasn't created. Fix: pass `error` as a prop to `RoundSetupStep` and render it there, or clear `pendingCourse` in the catch branch so the error panel in the main view becomes visible.

**Nits**
- `apps/mobile/components/round/RoundTeeSelector.tsx:27` — `TeePicker` has 2 callers (`RoundSetupStep` in `new.tsx` and the `RoundTeeSelector` wrapper in the same file). CONTRIBUTING.md's 3-caller rule does carve out an exception for cases where "the inline version is genuinely unreadable" — ~150 lines of JSX duplicated in two places plausibly clears that bar — but the count is short of 3, and if a third caller appears it'll be an easy justification after the fact.

---
Looked good: PR targets `dev`, not `main`; scope is one concern; tee is correctly optional and never blocks pace of play; `addBusy` is cleanly separated from the parent-supplied `busy` prop in `TeePicker` so an in-flight add doesn't race with the caller's own busy flag.

---

PR #544, opened 2026-06-10
fix(mobile): on-device Android QA batch — practice, stats, splash, course search

cner-smith:
On-device Android QA fixes (one batch, all small).

## Practice tab white-screened on a 'something went wrong / retry' loop
Root cause: `seed-demo.ts` wrote a stale `practice_plans` shape (`focus_areas` used `insight`, `drills` was a flat array) — the UI reads `area.reason`, so `normalizeCategoryProse(undefined).replace()` threw → ErrorBoundary loop. Only seeded accounts (demo/e2e) were affected; real generated plans use `reason`.
- Rewrite the seed to the live `@oga/core` shape (`{category, reason}`, `{sessions:[{blocks}]}`, + `coach_note`).
- Defensive guard (mobile + web): `normalizeCategoryProse()` tolerates null/undefined; session-blocks spread guards a missing array. A bad plan degrades, never white-screens.
- (Live demo dev+prod and e2e plans already regenerated/cleared out-of-band.)

## Practice lede contradicted the UI
"A column, not a checklist." sat above a literal checklist → **"Your game, read closely."** (keeps the editorial voice DESIGN.md asks for).

## Stats club labels wrapping
`DRIVER → 'DRIVE / R'`, `Ball Above/Below` — `StatRow` label now `numberOfLines={1}` + `adjustsFontSizeToFit` (minScale 0.8).

## Post-login splash showed old caps 'OGA'
`welcome.tsx` hardcoded "OGA"; brand wordmark is lowercase **"oga."** (matches the o. icon + landing). Font (Fraunces) was already correct.

## 'Pick a course' search text invisible in dark mode
The course-search `TextInput` set no text color → Android used the system-theme color (invisible on cream in dark mode). Added explicit ink `color` + `placeholderTextColor`. (login/signup get color via inline style / NativeWind class — verified, left as-is.)

## Test
mobile + web typecheck clean. On-device QA pending (next build).

### DIFF
--- a/apps/mobile/app/(app)/practice.tsx
+++ b/apps/mobile/app/(app)/practice.tsx
@@ -46,7 +46,10 @@ function asFocusAreas(value: unknown): StoredFocusArea[] {
 
 // UI safety net: rewrite any leaked raw snake_case category enums in displayed
 // prose (coach_note / focus reasons). `approach`/`putting` are already readable.
-function normalizeCategoryProse(text: string): string {
+// Tolerates null/undefined — a malformed plan (e.g. a focus area missing its
+// `reason`) must degrade to empty text, never crash the whole tab.
+function normalizeCategoryProse(text: string | null | undefined): string {
+  if (!text) return ''
   return text
     .replace(/\boff_tee\b/gi, 'off the tee')
     .replace(/\baround_green\b/gi, 'around the green')
@@ -330,7 +333,7 @@ function SessionBlock({
   completed: Set<string>
   onToggle: (blockId: string) => void
 }) {
-  const blocks = [...session.blocks].sort((a, b) => a.order - b.order)
+  const blocks = [...(session.blocks ?? [])].sort((a, b) => a.order - b.order)
   const doneCount = blocks.filter((b) => completed.has(b.id)).length
   return (
     <View style={{ borderTopWidth: 1, borderColor: LINE, paddingTop: 18, marginBottom: 28 }}>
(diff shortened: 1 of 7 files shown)


cner-smith:
## Review — fix(practice+stats): harden plan render, fix seed shape, stats label wrap

**Concerns**
- `apps/mobile/app/(app)/stats.tsx:601` — The `StatRow` label-shrink fix is a separate concern from the practice crash. CONTRIBUTING.md states "One concern per PR" and calls out that changes should be splittable for independent revert. The practice crash (seed shape + defensive guards) and the stats label wrapping are unrelated — different root causes, different files, different user-visible features. Both are small, but the rule exists precisely so a bad stats change doesn't drag in a practice fix when reverted. Consider splitting `stats.tsx` out to its own `fix/` branch.

- `apps/mobile/app/(app)/practice.tsx:49` / `apps/web/src/pages/practice/PracticePlanPage.tsx:71` — `normalizeCategoryProse` is now byte-for-byte identical in both files. CONTRIBUTING.md lists "format helpers" as belonging in `packages/core/src/`. This duplication predates the PR and is the right call to leave alone here, but a follow-up `chore/` PR to lift it to `@oga/core` would mean the next null-guard (or new category enum) only needs to land in one place.

---
Looked good: target branch is `dev`, no secrets or RLS-less migrations in the diff, both defensive guards (`?? []` and the null/undefined check on `text`) degrade cleanly rather than crashing, and the seed shape now matches the live `@oga/core` `StoredFocusArea`/`StoredSession` types.

---

PR #549, opened 2026-06-10
chore(scripts): seed demo account on real courses

cner-smith:
## What

Rewrites `scripts/seed-demo.ts` so the demo account's rounds sit on **real, fully-loaded courses** already in the DB (18 holes with tee/pin coordinates + rated tees) instead of three arbitrary courses — which were frequently hole-less synthetic courses that couldn't even be opened.

## Why

On-device App Review QA found seeded rounds attached to hole-less courses ("can't be viewed"), and the data was too uniform to demo well (every round the same weak-approach / strong-putting story).

## How

- **`selectDemoCourses`** prefers a curated set of verified-complete real courses (Whistling Straits, Harbour Town, Erin Hills), and otherwise **dynamically** picks fully-loaded courses from whatever DB the script targets — so `pnpm seed:e2e` against dev still works (it finds dev's own complete courses).
- Every round writes `hole_scores` for **all 18 holes** and `total_score` is the **exact sum**, so the scorecard always renders and totals match (the old script let those drift).
- Varied SG archetypes round-to-round; real per-tee colours; shots anchored to each course's **real** hole coordinates so the Patterns tab shows true dispersion.
- **Never** creates or mutates courses/holes — only the demo user's own rows.

## Verification

Ran against **prod** (App-Review account) and **dev** (e2e account): all 15 rounds render with 18 hole_scores, `total_score` equals the scorecard sum on every round, real rated tees used, varied SG (−3.4 … +1.3), and a realistic club distribution for the Patterns cones.

Scripts aren't in the workspace `tsc -b` / vitest paths, so CI is unaffected; validated by running the seed end-to-end.

### DIFF
--- a/scripts/seed-demo.ts
+++ b/scripts/seed-demo.ts
@@ -1,13 +1,23 @@
 /**
- * Seed a demo user with realistic round / shot / practice-plan data.
+ * Seed a demo user with realistic round / shot / practice-plan data on
+ * REAL courses.
  *
- * Idempotent: re-running wipes the demo user's prior rounds + plans
- * before re-inserting, so it's safe to run after migrations or schema
- * changes.
+ * Idempotent: re-running wipes the demo user's prior rounds / plans / clubs
+ * before re-inserting. It never creates or mutates courses or holes — it
+ * attaches the demo rounds to real, fully-loaded courses already in the DB
+ * (18 holes with tee/pin coordinates + rated tees), so every round renders a
+ * scorecard AND a map, the Patterns tab shows real per-course dispersion, and
+ * the handicap/tee features have genuine rating/slope to work with.
  *
- * Requires SUPABASE_SERVICE_ROLE_KEY (admin) — auth.admin.createUser
- * and bypassing RLS for the inserts both need it. Reads from .env at
- * the repo root.
+ * Course selection (selectDemoCourses):
+ *   1. Prefer the curated PREFERRED_COURSE_IDS when present (prod) — famous
+ *      courses verified to have full hole geometry + rated tees.
+ *   2. Otherwise fill dynamically from whatever fully-loaded courses the target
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — chore(scripts): seed demo account on real courses

**Concerns**
- `scripts/seed-demo.ts` (`loadCourse`): all three Supabase calls — `courses`, `holes`, and `course_tees` — destructure only `data`, silently discarding the `error` channel. A transient DB or permissions failure on the `holes` query makes the course appear to have no holes; `loadCourse` returns `null` and the curated course is quietly dropped. The script then falls through the dynamic path and may throw the vague "No course with 18 holes found" error with no indication of why. `courseIdsWithHoles` in the same file correctly does `if (error) throw error` — applying the same pattern inside `loadCourse` would surface real failures as real errors instead of silently masking them.

**Nits**
- `scripts/seed-demo.ts` (`pickTeeColor`): tee cycling uses the global round index `i`. Because each curated course appears every 3rd round (`i` steps by 3 across visits), when a course has exactly 1 or 3 rated tees, `i % teeColors.length` resolves to the same value on every visit — all rounds on that course use the same tee. Using a per-course counter rather than the global `i` would make the rotation actually rotate.

---
Looked good: `total_score` is now the exact sum of the 18 `hole_scores` (the key fix stated in the PR), courses and holes are never mutated, three-tier fallback keeps `pnpm seed:e2e` alive on sparse DBs, single concern, targets `dev`.

---

cner-smith:
Both addressed in 215f9bb:
- `loadCourse` now throws on the courses/holes/course_tees query errors instead of discarding them, so a transient failure can't masquerade as 'no holes' and silently drop a curated course.
- Tee colour cycles on a per-course visit counter (`Math.floor(i / courses.length)`) instead of the global round index, so each course's tees actually rotate.
Re-ran `pnpm seed:e2e` against dev — clean.

PR #567, opened 2026-06-11
fix(round-map): mirror mobile tee box on web (two dots from first shot)

### LINKED ISSUE #564: Round map: tee/first-shot marker renders as a labeled box vs dots for later shots
The first-shot (tee) marker is a labeled box while subsequent shots are dots (`markerFactories.ts`). Confirm whether this is intended or should be unified. Live-QA #2. DECISION NEEDED.

cner-smith:
Live-QA #564. Mirrors the mobile redesign: the draggable `TEE` badge + manual tee-placement flow are replaced by **two understated dots** flanking the tee shot, derived from the first shot (origin = first shot start, falling back to the stored tee; perpendicular to the line of play, ±4yd — reusing `@oga/core` `bearingDegrees`/`destinationYards`, identical geometry to mobile).

Removed: the 'Place tee box' button + the `placementMode==='tee'` tap branch (pin placement untouched).

**Note:** `teeOverride` stays in the round hooks (now never set) so the LENGTH-based synthetic-holes detection is byte-for-byte unchanged — cleaning up that vestigial state is a deliberate separate follow-up (it's entangled with the synthetic-holes landmine). Web typechecks clean. **Needs browser QA on the dev deploy** before it rides a release. Closes #564

### DIFF
--- a/apps/web/src/components/round/RoundMap.tsx
+++ b/apps/web/src/components/round/RoundMap.tsx
@@ -226,7 +226,6 @@ export function RoundMap({
     handicap,
     onMovePoint,
     onMovePin,
-    onMoveTee,
     onSetAim,
     onMoveExistingShot,
     onMoveExistingShotAim,
--- a/apps/web/src/components/round/map/RoundMapInstructionStrip.tsx
+++ b/apps/web/src/components/round/map/RoundMapInstructionStrip.tsx
@@ -116,26 +116,8 @@ export function RoundMapInstructionStrip({
     )
   }
   const placeButtons =
-    (needsTee && onStartPlaceTee) || (needsPin && onStartPlacePin) ? (
+    needsPin && onStartPlacePin ? (
       <>
-        {needsTee && onStartPlaceTee && (
-          <button
-            type="button"
-            onClick={onStartPlaceTee}
-            style={{
-              border: '1px solid #5C6356',
-              borderRadius: 2,
-              padding: '6px 10px',
-              fontSize: 12,
-              background: 'transparent',
-              color: '#5C6356',
-              fontWeight: 600,
-              letterSpacing: '0.02em',
-            }}
-          >
-            Place tee box
-          </button>
-        )}
         {needsPin && onStartPlacePin && (
           <button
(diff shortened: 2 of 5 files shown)


cner-smith (commented):
## Review — fix(round-map): mirror mobile tee box on web (two dots from first shot)

**Concerns**
- `apps/web/src/components/round/map/useMapLayers.ts` (deps array of `renderLayers`) — `onMoveTee` is still destructured from the input and listed in the `useCallback` dependency array, but it is never referenced inside `renderLayers` after this change. Every time the parent component creates a new `onMoveTee` identity (even a stable-shaped one that wasn't memoized because it was considered "in use"), the entire marker pass runs: all existing Mapbox markers are removed and re-added. That teardown-and-rebuild can produce a visible flicker on the map. The dead dep should be dropped from the array and the parameter removed from `UseMapLayersInput` — or, if `onMoveTee` is intentionally kept alive for the vestigial `teeOverride` follow-up the PR description mentions, a brief inline comment explaining that would forestall the same question next review.

**Nits**
- `apps/web/src/components/round/map/markerFactories.ts` (new `makeTeeDotMarker` function) — this has one call site. CONTRIBUTING.md: "Single-caller 'components' stay co-located in the parent file." The function is seven style lines; it would inline cleanly in the `for...of` loop in `useMapLayers.ts`. `markerFactories.ts` is already the natural home for all marker DOM constructors, so this may be a deliberate pattern extension rather than premature extraction — flagging for awareness rather than insisting on a change.
- `apps/web/src/components/round/map/markerFactories.ts` (block comment above `makeTeeDotMarker`) — the comment describes what the function does ("Two of these flank the tee shot…Non-interactive"), which the function name, shape, and `pointer-events:none` already convey. CONTRIBUTING.md: "Comments only when the *why* is non-obvious from the name and types alone."

---
Looked good: `bearingDegrees`/`destinationYards` correctly sourced from `@oga/core` with no logic duplicated in `apps/web`; no new `@oga/core` exports so no test gap; scope is one concern; PR targets `dev`.

---

PR #570, opened 2026-06-11
chore(db): prune 24 legacy pre-v2 drills (source NULL)

### LINKED ISSUE #560: Drill library shows 149 drills — verify against intended corpus (~125)
Migrations 0034 (~85) + 0035 (~40) insert ~125 drills (v2 corpus), but the drill library header shows '149 DRILLS'. ~24 extra likely leftover pre-v2 drills never removed. Query prod `drills`, confirm the right number, and clean up any stale/duplicate rows. Live-QA #8.

cner-smith:
Live-QA #560 — "is 149 drills the right number?". **No, 125 is.**

Verified against both DBs: **prod = 149, dev = 125, 0 dups either side.** The 24 extra on prod all carry `source IS NULL` + `verified = false` — the original pre-v2 corpus (0030) that the v2 corpus (0034/0035) added alongside but never removed. (Names: Gate drill, Wedge ladder, Lag putting clock, 9-shot grid, etc. — all the simple un-attributed ones.)

`0039` deletes `drills WHERE source IS NULL AND verified = false` → removes exactly those 24 on prod, **no-op on dev** (0 such rows), and makes a fresh full migrate converge to 125.

⚠️ **NOT applied to prod by me** (deleting prod rows while the user's asleep) — apply via the prod CLI channel (`supabase db push`) when you're ready. Practice-plan drill ids are jsonb (no FK) and the UI null-guards a missing drill (#544), so cascade risk is low. Closes #560

### DIFF
--- a/supabase/migrations/0039_prune_legacy_drills.sql
+++ b/supabase/migrations/0039_prune_legacy_drills.sql
@@ -1,15 +1,16 @@
 -- Prune the legacy pre-v2 drills that predate the curated corpus (0034 v2
--- data + 0035 v1 corpus). 0030 seeded an original ~40-drill set and added the
--- `source` column; the v2 corpus was inserted alongside it with a non-null
--- `source`, so the originals were never removed. On prod that left the drill
--- library at 149 instead of the intended 125.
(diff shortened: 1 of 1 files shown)


cner-smith (commented):

---

PR #571, opened 2026-06-11
feat(practice): sort, paginate, and grid the drill library

### LINKED ISSUE #561: Drill library: sorting, grid layout, and pagination
The full drill list is a long single-column scroll. Add: sort (alphabetical asc/desc), a grid layout option, and pagination (~25/page with flip-through) so it's not one long scroll. Live-QA.

cner-smith:
Live-QA #561. The drill library was one long single-column scroll. Adds:
- **A→Z / Z→A sort** toggle
- **pagination** at 24/page with Prev · 'n / m' · Next (no endless scroll)
- a **responsive grid** (auto-fill, 320px min) instead of a single column

Page resets on filter/sort change and clamps if the count shrinks. Existing filters (part-of-game, practice mode) and the expand-for-detail card are unchanged. Typechecks clean. Visual QA on the dev deploy recommended. Closes #561

### DIFF
--- a/apps/web/src/pages/practice/DrillLibraryPage.tsx
+++ b/apps/web/src/pages/practice/DrillLibraryPage.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState } from 'react'
+import { useMemo, useState } from 'react'
 import { Link } from 'react-router-dom'
 import type { BlockType, PlanCategory } from '@oga/core'
 import type { Database } from '@oga/supabase'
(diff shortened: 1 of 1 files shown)


cner-smith (commented):
## Review — feat(practice): sort, paginate, and grid the drill library

**Concerns**

- `DrillLibraryPage.tsx:40–44` — The `useEffect` that resets `page` to `0` causes a two-render cycle on every filter or sort change. `clampedPage` guards against an out-of-bounds page, but it does *not* reset a within-bounds stale one: if the user is on page 2 and changes a filter whose result set still spans 3+ pages, the first render briefly shows page 2 of the newly-filtered results before the effect fires and snaps back to page 0. The straightforward fix is to call `setPage(0)` directly inside the three change handlers (category, mode, sort), which eliminates the extra render and the flash entirely.

- `DrillLibraryPage.tsx` (overall scope) — This PR changes three distinct things: a CSS grid restyle of `DrillCard` (full card border replacing the bottom-only separator, `maxWidth: 720` removed), a sort-toggle feature, and a pagination feature. CONTRIBUTING.md is explicit: "UI restyle, bug fix, and new feature do not belong in the same PR." The grid/card-border change is a restyle; sort and paginate are new features. They all address #561, but a shared parent issue doesn't consolidate them into one concern. I recognise the grid and pagination are visually coupled — you wouldn't paginate a single-column list the same way — so this may be a judgement call, but flagging it for consistency with the project rule.

**Nits**

- `DrillLibraryPage.tsx:47–49` — `pageCount` and `pageItems` are derived inline in the render body while every other piece of derived state in this component is `useMemo`-wrapped. The cost is negligible (one `Math.ceil` and one `Array.slice` over an already-memoized array), but the inconsistency is mildly surprising to read. Either wrap them in `useMemo` for consistency, or leave them inline — either is defensible, just pick one.

---
Looked good: targets `dev` not `main`, no `@oga/core` exports added or tested, no RLS surface touched, no secrets, `Pager` is co-located in the parent file satisfying the single-caller rule, sort direction is correctly wired through `clampedPage` so a shrinking result set never throws a bounds error.

---

PR #579, opened 2026-06-11
fix(stats): exclude putts from club accuracy

### LINKED ISSUE #574: Stats: putter appears in Club Accuracy with a yardage dispersion
## Symptom

The mobile **Stats → Club Accuracy** section lists **"PUTTER · 364 shots · 3.5 yd"** under MOST ACCURATE. Putting isn't a full-swing club-accuracy stat, and a lateral dispersion in **yards** is meaningless for putts.

## Mechanism (verified)

`packages/core/src/stats.ts:678-712` (`clubAccuracy`) buckets every shot by `s.club`, gated only on coordinate presence (`stats.ts:681-692`). There is **no** exclusion of putts — no `s.club === 'putter'`, no `s.lie_type === 'green'`, no `category === 'putting'` filter. `'putter'` is a first-class bag club, and green-logged putts carry start/aim/end coords, so they pass straight through. `lateralMissYards` (`stats.ts:45-65`) returns yards → the "3.5 yd" putter dispersion.

## Proposed fix

Skip putts at the top of the per-shot loop (`stats.ts:682`):

```ts
if (s.club === 'putter' || s.lie_type === 'green') continue
```

`@oga/core` export → needs a unit test in `packages/core/src/__tests__/`.

cner-smith:
Fixes the putter showing up under Stats → Club Accuracy with a yards dispersion. `clubAccuracy` now skips `club === 'putter'` / `lie_type === 'green'`. Adds a `@oga/core` test (540 pass).

Closes #574

### DIFF
--- a/packages/core/src/__tests__/stats.test.ts
+++ b/packages/core/src/__tests__/stats.test.ts
@@ -483,3 +483,42 @@ describe('sgStandouts', () => {
     expect(s.strongest).toEqual({ key: 'putting', value: 0.9 })
   })
 })
+
+describe('clubAccuracy — putter exclusion', () => {
+  // A putter with full start/aim/end coords would otherwise bucket into club
+  // accuracy with a meaningless yards dispersion. Build 3 such putts + 3 real
+  // 7i shots (the n >= 3 floor) and assert only the iron survives.
+  function coordShot(overrides: Partial<ShotRow>): ShotRow {
+    return makeShot({
+      start_lat: 35.0,
+      start_lng: -97.5,
+      aim_lat: 35.001,
+      aim_lng: -97.5,
+      end_lat: 35.001,
+      end_lng: -97.4999,
+      ...overrides,
+    })
+  }
+  const round = makeRound({
+    hole_scores: [
+      makeHoleScore({
+        holes: makeHole({ id: 'h1', number: 1, par: 4 }),
+        shots: [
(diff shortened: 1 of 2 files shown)


cner-smith (commented):
## Review — fix(stats): exclude putts from club accuracy

**Concerns**
- `packages/core/src/stats.ts:686` — The guard is `s.club === 'putter' || s.lie_type === 'green'`. The `lie_type === 'green'` arm would also exclude a non-putter club hit from the green — a wedge chip from the putting surface, for example. The PR description and the inline comment only mention putters/putts, so it's unclear whether this wider exclusion is intentional. If it is, the comment should say so explicitly. If only putter exclusion is needed, the `|| s.lie_type === 'green'` clause should be dropped.

- `packages/core/src/__tests__/stats.test.ts:519` — All three excluded shots set both `club: 'putter'` and `lie_type: 'green'` together, so neither branch of the OR is exercised in isolation. As written, removing `s.club === 'putter'` from the condition and keeping only `s.lie_type === 'green'` would leave the test green. A case with `{ club: 'putter', lie_type: 'fringe' }` (putter from off-green) and a case with `{ club: 'sw', lie_type: 'green' }` (wedge from green) would pin down which clause does the work and make the intended behavior explicit.

---
Looked good: logic is in `@oga/core` not in an app lib, a test is added for the new exclusion, PR targets `dev`, single concern.

---

PR #581, opened 2026-06-11
fix(practice-plan): frame focus areas by SG sign, not always 'weakness'

### LINKED ISSUE #573: Practice plan labels neutral/positive SG categories as weaknesses
## Symptom

On the `appreview` demo account (handicap 12.4), the generated practice plan headline reads **"Attack approach _and off the tee_ weaknesses…"** and lists Off-the-tee as focus area #2: _"Off the tee ranks second at 0.03, needing blocked setup and path drills…"_

But that player's SG by category is: **Off-the-tee 0.00 / +0.03 (neutral)**, Approach **-1.30** (the only real leak), Around-green +0.10, Putting +1.16. The plan's own prose even admits _"off the tee is nearly neutral at 0.03"_ while still building a focus block for it and calling it a weakness.

## Mechanism (verified)

The focus selection is a pure ascending SG sort with **no sign/baseline gate**:

- `supabase/functions/_shared/practice-plan/digest.ts:66-75` — builds `ranked_weaknesses` by sorting categories ascending on raw SG average; returns all of them, "worst" = lowest even when lowest is positive.
- `supabase/functions/_shared/practice-plan/prompt.ts` rules 4 (line 186) and 8 (lines 196-197) — instruct Claude to include a non-warmup block for **each of the top-2 "weaknesses."**
- `supabase/functions/_shared/practice-plan/validation.ts:83-85` (D5) — **hard-rejects** any plan that doesn't cover `weaknesses.slice(0,2)`, so a neutral/positive category is structurally guaranteed a focus block whenever it's 2nd-lowest.

So SG values are already bracket-relative, but nothing checks whether the 2nd focus is actually below par.

## Decision / proposed fix

Keep selecting the two lowest categories (don't shrink the list), but make the digest + prompt **sign-aware**:

- A category with SG `< 0` is framed as a **leak** ("your biggest drag on scoring").
- A neutral/positive category is framed as **already strong — "you're doing this well, let's make it great"** (a sharpen/maintenance block, not a "weakness").

This fixes the mislabeling without the complexity of a shrinking focus list.

## Future (separate, larger)

Eventually we may want true sub-stat targeting — digging past category SG into specific distances / lies / miss patterns to pick what to work on — rather than just the two lowest categories. Tracked as a follow-up if/when we pursue it.

cner-smith:
## What

The practice plan picks the two lowest SG categories as focus areas, but the prompt told Claude to frame **both** as "weaknesses" regardless of sign — so a neutral/positive category (off-the-tee at +0.03 on the demo account) was headlined as a weakness to attack.

Per the decision on #573: keep the two-focus selection (still surfaces and practices the next-lowest category, even if positive), but make the framing **sign-aware** in the system prompt:
- negative average → genuine leak to attack;
- zero-or-positive → framed as an existing strength to sharpen ("already solid — let's make it great"), never a weakness.

Prompt-side only — the digest already carries signed per-category `averages`. Both the `@oga/core` copy and the vendored Edge copy are updated and verified in sync; 539 core tests pass.

## Activation (post-merge)

⚠️ This only takes effect once **`generate-practice-plan` is redeployed to prod** (`supabase functions deploy generate-practice-plan`). Prompt-only change (no schema/import changes), but it needs a live smoke: generate a plan for an account with one negative + one positive focus category and confirm the positive one reads as a strength-to-sharpen, not a weakness.

Closes #573


cner-smith (commented):
## Review — fix(practice-plan): frame focus areas by SG sign, not always 'weakness'

**Concerns**
- `packages/core/src/practice-plan/prompt.ts:211` (rule 11) — the boundary condition for zero. Rule (11) groups `0.00` with positive values and tells the model to frame it as "already solid … let's make it great". A `0.00` SG average means exactly average — neither losing nor gaining strokes against the field — not a genuine strength. For a player whose lowest category sits at exactly `+0.00`, that framing is a small misrepresentation. Consider tightening the boundary: strictly-positive (`> 0`) gets the "strength to sharpen" frame; zero gets a neutral frame ("right at average — let's push it positive") so the language matches what the number actually means.

- `supabase/functions/_shared/practice-plan/prompt.ts` — this file is a manually-maintained vendor copy of the `@oga/core` module. The PR updates both correctly, but there's no enforcement: a future contributor editing only one copy will silently drift them. A `scripts/` sanity check (e.g., `diff` the two files in CI) or even a comment at the top of each file naming the other would reduce the maintenance risk.

---
Looked good: prompt-only scope, correct target branch (`dev`), both copies updated identically, single concern per PR, no type violations, no security issues, no domain pillar violations.

---

PR #582, opened 2026-06-11
fix(putt): derive break direction from aim, drop the Break (line) chips

### LINKED ISSUE #576: Putt diagram: derive break direction from aim instead of manual chips
## Symptom

On the mobile live-round putt screen, a 21 ft putt renders a **curved** path captioned **"12 in right"** while the top-right label reads **"STRAIGHT."** The label and the drawn break contradict each other.

## Mechanism (verified)

The two are driven by **independent inputs**:

- **Corner label** ← `breakDirection` prop, from the **Break (slope/line) chips** (`GreenDiagram.tsx:232-236`). `combinedBreakDirection` (`packages/core/src/types.ts:208-218`) returns null when both axes are unset, and `PuttingSheet.tsx:289-294` applies `?? 'straight'` — so if the user never taps a Break chip, the corner reads **STRAIGHT regardless of aim.**
- **Curve + "12 in right"** ← `aimOffsetInches` prop, from **dragging the aim handle** (`GreenDiagram.tsx:177-194`, `formatAim` at `:314-319`; parent wiring `PuttingSheet.tsx:288/295`). The curve is actually rendering **aim bias**, not break.

So the diagram bends to your aim while the label describes a separate, unset break field.

## Decision

Drop the manual **straight / left-to-right / right-to-left** break-direction choice entirely and **derive break direction from the aim offset's sign** (the data already lives in `aimOffsetInches`). One input instead of two; the corner label, the curve, and the caption all come from the same source and can no longer disagree. (Keep the vertical slope/uphill-downhill input as-is — this is only about the horizontal direction.)

Reconciliation seam: the corner-label `<Text>` at `GreenDiagram.tsx:232-236` + removing the horizontal Break chip in `PuttingSheet.tsx`.

cner-smith:
Resolves the green-diagram contradiction where the corner label ("Straight") could disagree with the drawn curve + caption ("12 in right") because they came from two independent inputs.

The manual **Break (line)** chips are removed; `breakDirectionHorizontal` is now derived from the aim handle's offset sign — aim right of the hole → right-to-left break, aim left → left-to-right, with a ±2in deadband matching `formatAim`'s. Label, curve, and caption now share one source. Vertical slope (uphill/downhill) is unchanged and still carries the combined label when the aim is straight.

Mobile typecheck green. Mobile-only, no e2e — **needs on-device QA**: drag the aim each way and confirm the corner label matches the curve, and that a straight aim with an uphill/downhill slope still labels correctly.

Closes #576

### DIFF
--- a/apps/mobile/components/round/PuttingSheet.tsx
+++ b/apps/mobile/components/round/PuttingSheet.tsx
@@ -74,11 +74,18 @@ const BREAK_SLOPE_OPTIONS: { value: BreakDirectionVertical; label: string }[] = 
   { value: 'flat', label: 'Flat' },
   { value: 'downhill', label: 'Downhill' },
 ]
-const BREAK_LINE_OPTIONS: { value: BreakDirectionHorizontal; label: string }[] = [
-  { value: 'left_to_right', label: 'L → R' },
-  { value: 'straight', label: 'Straight' },
-  { value: 'right_to_left', label: 'R → L' },
-]
+// The horizontal break is derived from the aim line rather than chosen: the
+// player aims to the side the putt breaks FROM, so a right-of-hole aim plays a
+// right-to-left break and vice-versa. Mirrors GreenDiagram's formatAim 2-inch
+// deadband so a dead-straight aim leaves the axis unset, letting the vertical
+// slope carry the legacy combined label.
+function horizontalBreakFromAim(
+  offsetInches: number,
+): BreakDirectionHorizontal | undefined {
+  if (offsetInches > 2) return 'right_to_left'
+  if (offsetInches < -2) return 'left_to_right'
+  return undefined
+}
 
 const SLOPE_INTENSITY_LABELS = ['Flat', 'Slight', 'Moderate', 'Strong', 'Severe']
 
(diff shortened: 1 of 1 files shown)


cner-smith (commented):
## Review — fix(putt): derive break direction from aim, drop the Break (line) chips

**Concerns**
- `apps/mobile/components/round/PuttingSheet.tsx:84` — The deadband comparison (`offsetInches > 2` / `< -2`) doesn't round first, but `GreenDiagram.tsx`'s `formatAim` does (`Math.abs(Math.round(offsetInches)) <= 2`). For a fractional aim like `2.3 in`, `formatAim` returns "Straight" while `horizontalBreakFromAim` returns `'right_to_left'` — the caption and corner label would contradict each other, which is exactly the problem this PR is fixing. In practice this is safe because `GreenDiagram`'s `onEnd` stores `Math.round(...)`, so live putts always carry integer offsets. But the two functions now encode the same threshold with different comparison logic, which is fragile if either is ever touched independently. The minimum fix is a shared `const AIM_DEADBAND_INCHES = 2` imported from one place; a stricter read of CONTRIBUTING.md's "format helpers live in `@oga/core`" would lift both functions there, but that's probably out of scope for this PR.

**Nits**
- `apps/mobile/components/round/PuttingSheet.tsx:80-87` — `horizontalBreakFromAim` has one caller. CONTRIBUTING.md's 3-caller rule: "Don't extract a helper… unless it has three or more callers, or the inline version is genuinely unreadable." The inline ternary (`n > 2 ? 'right_to_left' : n < -2 ? 'left_to_right' : undefined`) is readable; the explanatory comment could move to the `onAimChange` callback. Take or leave — the named function does make the golf convention legible.

---
Looked good: PR targets `dev`, single concern, correct break-direction convention (aim right → `right_to_left`), `undefined` deadband handled cleanly by the `?? 'straight'` fallback downstream, dead code (`BREAK_LINE_OPTIONS`, `setBreakLine`) fully removed with no dangling references.

---

PR #583, opened 2026-06-11
fix(patterns): drop putter from the Shot Patterns club picker

cner-smith:
Web parallel of #574 (caught by cross-agent review of #579). The Shot Patterns club picker exposed a **Putter** chip rendering a meaningless aim-relative yards dispersion. Filtered `putter` out of the picker; default club stays `7i`.

Refs #574

### DIFF
--- a/apps/web/src/pages/patterns/ShotPatternsPage.tsx
+++ b/apps/web/src/pages/patterns/ShotPatternsPage.tsx
@@ -145,7 +145,9 @@ export function ShotPatternsPage() {
 
       <Section kicker="Club">
         <div className="flex flex-wrap" style={{ gap: 6 }}>
-          {CLUBS.map((c) => (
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(patterns): drop putter from the Shot Patterns club picker

**Concerns**
- `apps/web/src/pages/patterns/ShotPatternsPage.tsx:148` — The diff removes `putter` from the picker UI but doesn't guard against `selectedClub === 'putter'` arriving from a persisted source (URL param, localStorage, or similar). Because putter was selectable before this fix, a returning user with a cached selection could land in a state where no chip appears active while putter data is still queried and rendered. May be a non-issue if there's no persistence layer for this picker — unclear from the diff alone.

---
Looked good: targets `dev`, single-concern fix, JSX comment explains the why per CONTRIBUTING.md's "non-obvious why" standard, no package-boundary or security issues.

---

PR #586, opened 2026-06-11
fix(live-round): clarify the place-ball CTA marks the ball at your GPS (#483)

### LINKED ISSUE #483: [mobile] Live-round: 'Acquire GPS' button seems to do nothing — should mark the ball at the GPS dot
First real on-course test (2026-06-01). The **"Acquire GPS" button in live-round mode appears to do nothing**, and it's not obvious that its purpose is to drop the ball marker at the player's current GPS position.

## Desired flow
Tapping the GPS button should **mark the ball where the GPS dot is** — snap/place the ball marker to the current fix — with clear visible feedback that it happened.

## Investigate / fix
- Confirm what the button's handler actually does today — it may only recenter the camera (conflated with the recenter control) rather than placing the ball.
- Make the action explicit: "drop ball at my location," with an obvious affordance/label.
- Visible confirmation on tap (marker moves; optional toast).

Note: GPS itself worked fine — it lagged while walking and caught up when stationary, which is expected (see #331, motion-aware accuracy).

cner-smith:
## Problem (your on-device feedback)
In the live flow there's no clear 'mark ball at my GPS' affordance. Root cause confirmed in code: live mode auto-tracks the ball to the player's GPS during PLACE_BALL (`useHoleState.ts:281` `setBall(smoothed)`), so `ball` is always set and `MapBottomChrome` always shows the generic **'Mark ball here →'** — the **'Mark ball at my GPS →'** branch (needs `!ball`) was effectively dead.

## Fix (A + C)
- **A — GPS-explicit label.** New derived `ballFromGps` (live + GPS + ball present + not yet dragged), threaded from `LiveRoundSession` via a reactive `ballMoved` flag (set on manual ball drag; reset on each fresh PLACE_BALL entry and hole change). CTA reads **'Mark ball at my GPS →'** while GPS-tracked, reverting to **'Mark ball here →'** once you drag the marker.
- **C — hint** under the CTA: *'The ball follows your GPS — drag to adjust.'*

## Verification
Mobile typecheck clean. **Mobile-only, no e2e — needs on-device QA**: start a live hole, confirm the CTA reads 'Mark ball at my GPS →' and the hint shows; drag the ball and confirm it flips to 'Mark ball here →'; re-place / next hole and confirm it resets to the GPS wording.

Closes #483

### DIFF
--- a/apps/mobile/components/round/LiveRoundSession.tsx
+++ b/apps/mobile/components/round/LiveRoundSession.tsx
@@ -88,6 +88,10 @@ export default function LiveRoundSession({
   // historical-shot scatter overlay; the render lands in T4. Off by
   // default — it's a summoned planning aid, not always-on clutter.
   const [dotsVisible, setDotsVisible] = useState(false)
+  // True once the player drags the ball off its GPS-tracked position this
+  // PLACE_BALL cycle — flips the place-ball CTA from "Mark ball at my GPS"
+  // to the generic "Mark ball here". Reset on each new placement / hole.
+  const [ballMoved, setBallMoved] = useState(false)
   // Aim overlay shape + size (T3). Tee → arc band, Appr → circle ring; the
   // rail index sizes each, kept per-mode so switching modes preserves the
   // other's pick. Default Tee, widest rail.
@@ -108,6 +112,7 @@ export default function LiveRoundSession({
     setLoggerOpen(false)
     setPinPlacementOpen(false)
     setLoggerInitial({})
+    setBallMoved(false)
     // Clear only hole-scoped dialogs (onGreen / aim). Session-scoped
     // confirms (delete / leave / end / exit) stay open across hole
     // navigation — a confirmDelete dialog mid-navigation should not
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — fix(live-round): clarify the place-ball CTA marks the ball at your GPS (#483)

**Concerns**
- `apps/mobile/components/round/LiveRoundSession.tsx:143–148` — The `useEffect` that resets `ballMoved` to `false` on PLACE_BALL entry fires after the commit phase, not before paint. On the first render following a state transition into PLACE_BALL (e.g., after a shot is logged), `ballMoved` is still `true`, so `MapBottomChrome` renders "Mark ball here →" for one frame before the effect triggers a re-render with the GPS label. The PR description says the GPS wording should appear on "fresh PLACE_BALL entry" — that intent is undercut by the asynchronous reset. `useLayoutEffect` runs synchronously before the screen is committed and would eliminate the flicker without any other change needed.

**Nits**
- `apps/mobile/components/round/MapBottomChrome.tsx:~108–120` — The hint `<View>` and its inner `<Text>` both carry inline style object literals (`style={{ backgroundColor: CHROME_BG, ... }}` and `style={[TYPE.body, { ... }]}`). During a live round, `MapBottomChrome` re-renders on every incoming GPS fix; inline objects allocate fresh references each render. These are static values — hoisting them to a `StyleSheet.create` entry or a module-level `const` is the React Native convention and avoids the allocations on the hot path.

---
Looked good: scope held to one concern; `ballFromGps` is optional and defaults to `undefined`, so non-live and past-mode callers are unaffected; `resetPerHoleState` correctly handles the hole-change reset path via a direct `setState` call rather than an effect; targets `dev`.

---

PR #587, opened 2026-06-11
fix(live-round): free the aim handle near the green (#473)

### LINKED ISSUE #473: Live round: can't drag the aim marker when the distance pill overlaps it (near green)
On the mobile live-round aim screen, when the aim crosshair sits under the carry distance pill (happens around the green, where the ball->aim leg is short), you cannot drag the aim marker — the label steals the touch.

Root cause (confirmed in source): the distance labels (AimDistancePill, and the remaining-distance label in HoleMap) are @rnmapbox/maps MarkerViews. MarkerView's content wrapper hardcodes onStartShouldSetResponder => true plus onTouchEnd stopPropagation (node_modules/@rnmapbox/maps/src/components/MarkerView.tsx lines 133-138, v10.1.33), so it captures the touch before it reaches the draggable aim PointAnnotation beneath. pointerEvents='none' on the child View is a no-op against this (attempted and reverted).

Fix options to evaluate on-device (don't guess — each is a build):
- Convert the labels from MarkerView to PointAnnotation, and verify an overlapping non-draggable annotation does not itself block the handle drag.
- Suppress/hide the carry pill when the ball->aim leg is short (near green).
- Reposition the pill clear of the handle's grab zone (anchor offset).
- Render the labels as RN absolute overlays with pointerEvents='none' (needs geo->screen projection + tracking).

Files: apps/mobile/components/round/markers/AimDistancePill.tsx, apps/mobile/components/round/HoleMap.tsx (remaining-distance MarkerView).

cner-smith:
## Problem
Near the green (short ball→aim leg) you can't drag the aim marker — the carry distance pill steals the touch. Root cause (per the issue): the distance labels are @rnmapbox `MarkerView`s whose content wrapper hardcodes the touch responder, so a label overlapping the draggable aim `PointAnnotation` captures the gesture. `pointerEvents:'none'` is a no-op against it (tried + reverted before).

## Fix (the lowest-risk option from the issue)
Hide each distance label once its leg is under `MIN_LABEL_LEG_YARDS` (20yd) — at that point its midpoint is inside the handle's grab zone. Applies to both the carry pill (ball→aim) and the remaining label (aim→pin). The persistent top **'to hole'** readout still shows distance near the green, so nothing critical is lost.

## Caveat
The overlap is a screen-space problem; the yard threshold is a zoom-dependent proxy. **20yd is a starting value that needs on-device tuning** — the issue explicitly flagged each option as 'a build'. Mobile-only, no e2e.

Closes #473

### DIFF
--- a/apps/mobile/components/round/HoleMap.tsx
+++ b/apps/mobile/components/round/HoleMap.tsx
@@ -155,6 +155,13 @@ const AIM_EXTENSION_YARDS = 250
 // Android drag jank. Gating to ~25 Hz cuts that storm while the readouts
 // still track the finger; onDragEnd always applies the final position.
 const AIM_DRAG_THROTTLE_MS = 40
+// Distance labels are @rnmapbox MarkerViews that capture touches (their content
+// wrapper hardcodes the touch responder), so a label sitting at its leg's
+// midpoint steals the drag from the aim handle when that leg is short — the
+// near-green "can't drag the aim marker" bug (#473). Hide a label once its leg
+// is under this many yards; its midpoint is then within the handle's grab zone.
+// Tune on-device — the screen overlap is zoom-dependent, this is a geo proxy.
+const MIN_LABEL_LEG_YARDS = 20
 
 function extractCoord(feature: unknown): LatLng | null {
   const geom = (feature as { geometry?: { coordinates?: unknown } } | null)?.geometry
@@ -797,7 +804,9 @@ export function HoleMap({
             </Mapbox.PointAnnotation>
           )}
 
-          {aimMidpoint && aimDistanceYards !== null && (
+          {aimMidpoint &&
+            aimDistanceYards !== null &&
+            aimDistanceYards >= MIN_LABEL_LEG_YARDS && (
             <AimDistancePill
               midpoint={aimMidpoint}
               display={toDisplay(aimDistanceYards)}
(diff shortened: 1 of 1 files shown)


cner-smith (commented):
## Review — fix(live-round): free the aim handle near the green (#473)

**Concerns**
- `apps/mobile/components/round/HoleMap.tsx:161` — The PR description explicitly calls `MIN_LABEL_LEG_YARDS = 20` "a starting value that needs on-device tuning," yet the PR is set to `Closes #473`. Closing the tracking issue while the threshold is self-described as unvalidated leaves the zoom-level tuning work with no tracking item — if 20yd proves too tight or too loose on device, there's nowhere to record that. Consider keeping #473 open with a status note (or opening a follow-up `fix/` issue) before this merges, so the validation work doesn't get lost.

**Nits**
- `apps/mobile/components/round/HoleMap.tsx:163` — The last sentence of the block comment ("Tune on-device — the screen overlap is zoom-dependent, this is a geo proxy.") is already covered by the PR description and the caveat section there; trimming it would tighten the comment without losing the non-obvious *why*. Minor — the comment earns its length explaining the `@rnmapbox` touch-responder constraint.

---
Looked good: single-concern fix targeting `dev`, no package boundary issues, no security concerns, no domain pillar violations, both labels guarded symmetrically, and the persistent "to hole" readout preserves critical distance info when both pills go under threshold.

---

PR #588, opened 2026-06-12
fix(backfill-ratings): pace under rate limit + resume cursor + quota breaker

cner-smith:
The course slope/rating backfill (`scripts/backfill-ratings.ts`, `pnpm backfill:ratings`) matches our OSM-sourced courses to GolfCourseAPI and upserts per-tee `course_rating`/`slope_rating` into `course_tees` — the data the WHS handicap engine needs.

## Problem
A live prod run spent most of its wall-clock asleep. GolfCourseAPI throttles around 4 req/s and sends **no `Retry-After` / rate-limit headers**, so the `429` handler fell back to a blind **60s backoff**. Measured 429 recovery is actually **sub-second**, so that flat sleep was ~60x too long and dominated runtime (32 backoffs ≈ 32 min of pure sleeping in one run).

## Changes
- **Pacing:** measured that **350ms+ spacing → 0/12 throttled** while 250ms (the old default) trips constantly. Backoff dropped `60_000ms → 3_000ms`; runs use `--delay-ms 500` to stay comfortably under the limit, effectively eliminating 429s.
- **`--after-id <uuid>` resume cursor:** `candidateCourses` gains `.gt(\047id\047, AFTER_ID)` so a partial run can resume past where it stopped instead of re-scanning from id 0 (the rated-course skip set already prevents re-querying matched courses; this skips the known-no-match low-id region too).
- **Daily-quota circuit breaker:** `searchApi` now returns `null` on throttle-exhaustion (vs `[]` for a genuine empty result); the main loop counts consecutive throttles and stops cleanly after 8 in a row — "re-run to resume" — instead of doing 3 futile 9s retries on every remaining course once the daily cap is hit.

## Verification
- Typechecks clean.
- Used live: prod rated-course coverage went **5,211 → 8,060** (+2,849 courses, +14,792 tees) across paced runs with **0 throttles** on the fresh region; the breaker then stopped the mop-up cleanly when the daily 10k cap was reached.

No schema or app changes — script-only.

### DIFF
--- a/scripts/backfill-ratings.ts
+++ b/scripts/backfill-ratings.ts
@@ -19,6 +19,7 @@
 //   GOLFCOURSEAPI_KEY=... pnpm tsx scripts/backfill-ratings.ts --write --limit 5000
 //   flags: --write  --limit N  --delay-ms N  --threshold 0.0-1  --max-km N
 //          --has-state (only courses with a state set — likelier US/rated)
+//          --after-id UUID (resume: only courses with id > UUID)
 //          --out path.jsonl
 //
 // Env: SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY (via scripts/crawl/client),
@@ -50,6 +51,7 @@ const DELAY_MS = parseInt(opt('delay-ms', '250'), 10)
 const THRESHOLD = parseFloat(opt('threshold', '0.78'))
 const MAX_KM = parseFloat(opt('max-km', '25'))
 const HAS_STATE = flag('has-state')
+const AFTER_ID = opt('after-id', '')
 const OUT = opt('out', 'backfill-ratings-matches.jsonl')
 
 // ---- GolfCourseAPI shapes (subset of its OpenAPI TeeBox / Course) ----------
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(backfill-ratings): pace under rate limit + resume cursor + quota breaker

**Concerns**
- `scripts/backfill-ratings.ts:339-345` — The circuit-breaker stop message says "Re-run to resume" but doesn't include the course ID at which the run stopped. Unmatched courses are not written to the JSONL audit log, so the operator has no easy way to derive an `--after-id` value from the script's output — they'd have to grep console lines manually. Since `--after-id` and the circuit breaker were both added in this PR to work together, the stop message could close the loop: `Re-run with --after-id ${course.id} to resume.` The `course` variable is in scope at the `break`.

**Nits**
- `scripts/backfill-ratings.ts:337-338` — `sleep(DELAY_MS)` runs unconditionally before the `null` check. When `searchApi` returns `null`, all three retry attempts already slept 3 s each (9 s total), so `DELAY_MS` (500 ms by default) is charged on top of a retry-exhausted call rather than on a successful one. Functionally harmless — consistent pacing regardless of outcome — but it's slightly surprising that a quota-busted call incurs an extra delay. If the intent is pacing only between successful API calls, moving the sleep inside the `candidates !== null` branch would make that clearer; if pacing-always is intentional (reasonable), a short comment would explain why.

---
Looked good: `null` / `[]` distinction for quota vs genuine empty result is a clean API contract; circuit-breaker logic is correct; `--after-id` pagination via `.gt('id', AFTER_ID)` on a consistently offset-based query is sound; scope held to one concern.

---

PR #589, opened 2026-06-12
fix(mobile): device-QA batch — past-round routing, live tee, home tile

cner-smith:
On-device QA fixes (mobile).

## Past-round routing (the big one)
Opening a round from the list could drop you on the bare **live map with no tabs** instead of the read-only Scorecard⇄Map view. Root cause was a routing heuristic + a finalize bug, not a missing screen:
- `[id]/index.tsx` decided "live vs past" purely on `total_score == null`. A finalized round whose score is null then redirected into the live state machine.
- `completeRound.ts` wrote `total_score: totalScore || null`, so a round finalized **without entered hole-scores became `total_score = null`** — feeding the bad redirect.

Fixes:
- Redirect into the live flow **only when `completed_at == null AND total_score == null`** (and not the past-logger creation race). `completed_at` is the canonical finalized flag, so a finalized round always opens the tabbed scorecard. The `total_score` guard keeps seeded past rounds (scored, no `completed_at`) out of the live map too.
- `completeRound` stores the real total (incl. `0`) instead of coercing to null.

## Live tee anchor (GPS-first)
~82% of courses now carry an OSM `holes.tee_lat`, so the live tee box stopped following where you actually teed off and pinned to the OSM coordinate. Now in live mode the tee anchors to your **first shot's start**, with the OSM tee as the pre-shot fallback. Shared `holes.tee_lat` is never overwritten (`writeTee` still only persists synthetic holes); `useHoleData` is live-round-only so past rounds are unaffected.

## Home "By the numbers"
Replaced the static **Categories (4)** tile (always 4 — useless) with **Best round** (lowest real score; the `0` sentinel excluded).

## Verification
- Mobile typecheck clean.
- Data check (dev/prod): all real rounds are scored; the redirect change keeps them on the tabbed view and stops finalized-but-null-score rounds from hitting the live map.

Mobile-only; no schema changes. Wants on-device confirmation in a fresh build (past round opens to tabbed scorecard; live tee follows first shot).

### DIFF
--- a/apps/mobile/app/(app)/index.tsx
+++ b/apps/mobile/app/(app)/index.tsx
@@ -145,9 +145,16 @@ export default function Home() {
     const avgScore = scoreRounds.length > 0
       ? scoreRounds.reduce((s, r) => s + (r.total_score ?? 0), 0) / scoreRounds.length
       : null
+    // total_score === 0 is the past-round-logger sentinel for "no score
+    // entered" (map-created rounds), so exclude it from the best-round min —
+    // otherwise a single unscored round always reads as a best of 0.
+    const realScores = rounds
+      .map((r) => r.total_score)
+      .filter((s): s is number => s != null && s > 0)
+    const bestScore = realScores.length > 0 ? Math.min(...realScores) : null
     const totalSG = avgs.reduce((s, a) => s + a.value, 0)
     const sorted = [...avgs].sort((a, b) => b.value - a.value)
-    return { avgScore, totalSG, weakest: sorted[sorted.length - 1]!, strongest: sorted[0]! }
+    return { avgScore, bestScore, totalSG, weakest: sorted[sorted.length - 1]!, strongest: sorted[0]! }
   }, [rounds])
 
   const eyebrow =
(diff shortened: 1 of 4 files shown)


cner-smith (commented):
## Review — fix(mobile): device-QA batch — past-round routing, live tee, home tile

**Blockers**
- Multiple files — three distinct concerns (routing fix, live-tee anchor, home tile) ship in one PR. CONTRIBUTING.md is explicit: "One concern per PR. UI restyle, bug fix, and new feature do not belong in the same PR." The routing fix (`round/[id]/index.tsx` + `completeRound.ts`) is the load-bearing change here; the tee anchor (`useHoleData.ts`) and the home tile (`app/(app)/index.tsx`) could each go out independently. Bundling them makes a targeted revert impossible if one regresses on device.

**Concerns**
- `apps/mobile/app/(app)/round/[id]/index.tsx` — `handleSaveSG` contains a guard that now describes behaviour this PR fixes away:
  ```ts
  // completeRound writes `total_score: totalScore || null`, so an
  // all-zero past round comes back null …
  if (saved.total_score == null) {
    await updateRound(supabase, round.id, { total_score: 0 }, user.id)
    …
  }
  ```
  With `completeRound.ts` now writing `result.totals.totalScore` directly (no `|| null` coercion), `saved.total_score` after a `completeRound` call will never be `null` for a zero-score round. The fallback `updateRound` write is dead, and the comment now misrepresents the current contract. Both should be removed alongside the `completeRound` fix — leaving them ships a lie in the codebase and an extra unnecessary write on the code path if somehow triggered.

- `apps/mobile/lib/completeRound.ts` — `total_putts: result.totals.totalPutts || null` on the line immediately below the fixed field applies the same `|| null` coercion that was just corrected for `total_score`. A round where no putts were logged would null the putts count incorrectly. Worth a follow-up fix in its own PR.

---
Looked good: routing predicate (`completed_at == null && total_score == null`) is sound; `completeRound` `|| null` removal is correct; tee anchor to `previousShots[0]` is clean; no schema changes, no security issues, PR targets `dev`.


---

PR #590, opened 2026-06-12
fix(mobile): completed_at is the round lifecycle flag (review follow-up to #589)

cner-smith:
Addresses the code-review on #589 (which I merged before reading — this closes the loop).

## What the review caught
- `handleSaveSG` had a `if (saved.total_score == null) { updateRound({total_score: 0}) }` fallback whose comment described the old `|| null` contract. With routing now on `completed_at`, that branch is dead and the comment was a lie.

## What I found verifying the rest
- The review also suggested fixing `total_putts: totalPutts || null` the same way `total_score` was. **Rejected** — `shortGameStats` (`packages/core/src/stats.ts:562`) null-filters `total_putts`, so storing `0` would drag putt averages. `scoringStats` (`stats.ts:366`) null-filters `total_score` the same way — meaning my merged `total_score`-stores-`0` change risked the identical corruption for a round finalized with no scores entered.

## Fix — make `completed_at` the lifecycle flag, `total_score` just the score
- `handleSaveSG`: remove the dead fallback + stale comment.
- `completeRound`: restore `total_score: result.totals.totalScore || null` (null = no score → correctly excluded from null-filtered score/putt stats).
- `useActiveRound`: gate the Resume banner on `completed_at IS NULL` **and** `total_score IS NULL`, so a scoreless finalized round leaves the banner and seeded scored rounds never show as active.
- Routing (#589) already keys on `completed_at`, so a finalized round opens the read-only summary regardless of score.

`total_putts` left as-is (correct). Mobile typecheck clean. No schema changes.

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -434,15 +434,8 @@ export default function RoundIndex() {
         supabase.from('hole_scores').select('*').eq('round_id', round.id),
       ])
       if (rRes.data) {
-        let saved = rRes.data as RoundRow & {
+        const saved = rRes.data as RoundRow & {
           courses?: { name: string | null } | null
-        }
-        // completeRound writes `total_score: totalScore || null`, so an
-        // all-zero past round comes back null — which would re-strand it on
-        // the live map on next open (#514). Preserve the non-null sentinel.
-        if (saved.total_score == null) {
-          await updateRound(supabase, round.id, { total_score: 0 }, user.id)
-          saved = { ...saved, total_score: 0 }
         }
         setRound(saved)
       }
--- a/apps/mobile/hooks/useActiveRound.ts
+++ b/apps/mobile/hooks/useActiveRound.ts
@@ -9,8 +9,11 @@ export interface ActiveRound {
   currentHole: number
 }
 
-// Active = total_score IS NULL AND played_at within the last day, so a
-// round abandoned a week ago doesn't haunt the home screen forever.
+// Active = not finalized (completed_at IS NULL) AND no score yet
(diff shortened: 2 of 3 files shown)


cner-smith:
## Review — fix(mobile): completed_at is the round lifecycle flag (review follow-up to #589)

**Concerns**

- `apps/mobile/hooks/useActiveRound.ts:38-39` — The double-null filter (`completed_at IS NULL AND total_score IS NULL`) handles seeded rounds correctly today, but surfaces a latent data inconsistency: rounds with `total_score IS NOT NULL` and `completed_at IS NULL` (seeded/legacy) are excluded from the resume banner here, and they presumably also fail the `completed_at IS NOT NULL` gate in #589's routing — so they route neither to the summary nor to the banner. Those rounds may be silently unreachable from the home screen. This PR doesn't introduce the problem, but the new lifecycle model makes it more visible. A follow-up backfill migration setting `completed_at` on every row where `total_score IS NOT NULL` would close the gap and let the `total_score IS NULL` guard be dropped from this query entirely.

**Nits**

- `apps/mobile/hooks/useActiveRound.ts:9-13` — The first two lines of the updated comment restate what the query already says ("Active = not finalized (completed_at IS NULL) AND no score yet (total_score IS NULL)"). CONTRIBUTING.md: *"Comments only when the why is non-obvious from the name and types alone. Don't narrate what the code does."* The last sentence about seeded rounds is the non-obvious why and worth keeping; the SQL restatement could be cut.

---
Looked good: dead `total_score: 0` fallback cleanly removed with its now-false comment, `let` → `const` follows naturally, restoring `|| null` correctly re-aligns `total_score` semantics with the null-filters in `scoringStats` and `shortGameStats`, PR is a single focused concern, targets `dev`.

---

PR #595, opened 2026-06-13
feat(mobile): FIR/GIR check marks on the past-round scorecard (#591)

### LINKED ISSUE #591: Mobile past-round scorecard: per-hole FIR/GIR check marks (web parity)
The mobile past-round Scorecard tab shows HOLE / PAR / SCORE / PUTTS / SHOTS. Web (`ScorecardView`) also shows per-hole **Fairway** and **GIR** columns. `hole_scores` already stores `fairway_hit` and `gir` (both `boolean | null`), so the data exists.

**Ask:** add Fairway (FIR) + GIR check-mark columns to the mobile scorecard table, mirroring web. Null = blank, true = ✓.

From on-device QA 2026-06-13.

cner-smith:
On-device QA (#591): mobile scorecard lacked the per-hole Fairway + GIR marks web shows.

- Added FIR + GIR columns to the scorecard table.
- Both derived per hole from the holes shots via @oga/core inferHoleStats (pure): check = hit, dot = miss, blank = n/a (e.g. fairway on a par 3). No dependency on whether hole_scores.fairway_hit/gir is persisted.
- Trimmed the Shots column to make room.

Needs device QA: it is now a 7-column row (Hole/Par/Score/Putts/FIR/GIR/Shots); please confirm it does not crowd/truncate on a narrow phone (we can drop Par or restyle if so). Mobile typecheck clean. Closes #591.

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -228,6 +228,18 @@ export default function RoundIndex() {
     for (const s of shots) m.set(s.hole_score_id, (m.get(s.hole_score_id) ?? 0) + 1)
     return m
   }, [shots])
+  // Shots grouped by hole_score_id, built once — the scorecard's FIR/GIR
+  // fallback reads this O(1) per row instead of filtering the full `shots`
+  // array per hole (was O(shots×holes) in the render body).
+  const shotsByHoleScoreId = useMemo(() => {
+    const m = new Map<string, ShotRow[]>()
+    for (const s of shots) {
+      const arr = m.get(s.hole_score_id)
+      if (arr) arr.push(s)
+      else m.set(s.hole_score_id, [s])
+    }
+    return m
+  }, [shots])
 
   // Refetch shots on screen focus. The end-round write order is:
   // total_score + hole_scores first, then pending shot inserts trickle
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — feat(mobile): FIR/GIR check marks on the past-round scorecard (#591)

**Concerns**

- `apps/mobile/app/(app)/round/[id]/index.tsx:775–779` — `shots.filter(s => s.hole_score_id === hs.id)` runs inside `sortedHoles.map(...)` in the render body, so every render walks the full `shots` array once per hole — O(shots × holes). For an 18-hole round with ~70 shots that's ~1 260 iterations per render, and this component re-renders on every scorecard edit. The file already uses the correct pattern for shot counts: a `useMemo`-backed `Map<string, number>` built once over `shots`. The same approach applies here — build a `Map<string, ShotRow[]>` alongside `holeScoreShotCount` and do an O(1) lookup in the map:

  ```ts
  const shotsByHoleScoreId = useMemo(() => {
    const m = new Map<string, ShotRow[]>()
    for (const s of shots) {
      const arr = m.get(s.hole_score_id) ?? []
      arr.push(s)
      m.set(s.hole_score_id, arr)
    }
    return m
  }, [shots])
  ```

  Then the render call becomes `inferHoleStats(hs ? (shotsByHoleScoreId.get(hs.id) ?? []) : [], h.par)`.

**Nits**

- `apps/mobile/app/(app)/round/[id]/index.tsx:773` — `// Derive FIR/GIR from the hole's shots (par-3 fairway = null/blank).` — the first clause narrates what the next two lines already say. Per CONTRIBUTING.md: "Comments only when the *why* is non-obvious from the name and types alone." The par-3 invariant is the non-obvious part; the comment could be just `// par-3 fairway → null (blank cell)`.

- `apps/mobile/app/(app)/round/[id]/index.tsx:824` — `key={i}` is safe here since the `[fairway, gir]` array is fixed-length and never reorders, but it's a pattern that lint and code-review tooling flag on sight. Stable string keys (`'fir'` / `'gir'`) cost nothing and are unambiguous.

---
Looked good: `inferHoleStats` stays in `@oga/core` (no logic added to `apps/mobile/lib/`), no new core exports without tests, single-concern change, targets `dev`.

---

PR #596, opened 2026-06-13
feat(mobile): past-round map → review stepper + edit-this-shot (#593)

### LINKED ISSUE #593: Mobile past-round map: shot stepper + edit-this-shot (stop reusing the live append flow)
Opening the **Map** tab on a finished round runs the **live** placement flow: "drag the ball to refine, then tap mark ball here," BALL/AIM/PIN modes, "+ SHOT", "N shots placed". For a round that already has shots, this just drops another marker / appends.

**Desired (read-first editing):**
- Lock onto the existing shots for the hole and let you **step shot-to-shot** (1 → 2 → …).
- Each shot has an **"Edit this shot"** affordance.
- **Add a new shot only off the last shot.**
- Stop showing live-only chrome/copy on a finished round — the past-round map should feel distinct from live mode.

Bigger UX redesign (relates to the past-vs-live separation theme and #482). Spec before building.

From on-device QA 2026-06-13.

cner-smith:
Rebuilds the **Map tab on a finished round** into a **review + edit** experience, while preserving the live placement flow for a round that's still being logged. Closes #593.

## Lifecycle-gated behavior (`completed = round.completed_at != null`)
The Map tab is reached in **both** lifecycle states (the `index.tsx` redirect), so the new UI is gated on whether the round is finalized:

- **Still logging (`completed_at == null`):** unchanged — full placement chrome, `BALL / AIM / PIN` modes (**pin-setting included**), `+ Shot`. You're entering an old round; you still need to drop pins and shots freely.
- **Finalized (`completed_at != null`):** the new **review stepper** — `‹ Shot N of M ›`, drag-to-reposition the selected marker, **"Edit this shot"** → the shared `PastHoleShotsSheet` editor, **"+ Add shot"** gated to the last shot only. No pin chrome (a finalized round already has its pins).

Both modes share the seed / persist / score-sync plumbing in one component.

## Implementation notes
- **`PastRoundMap`** gains a `completed` prop; the bottom panel + the post-placement handler tails branch on it.
- **`HoleMap`** gets an additive `tapToPlaceBall?: boolean` (**default `true`** → live round + logging untouched). Finalized-review passes `false` while not adding, so the selected marker is **drag-only** (stray taps don't move it).
- **`PastHoleShotsSheet`** gets an additive `initialShotId?: string | null` → "Edit this shot" opens straight into that shot's editor (vs the list). One shared editor, no duplication.

## Verification
- `npm run typecheck` (mobile) green — covers `LiveRoundSession` (the other `HoleMap` consumer) too.
- Not yet on-device. Suggested QA:
  - **Logging round:** confirm BALL/AIM/PIN + "+ Shot" still work and you can set a pin.
  - **Finalized round:** step a multi-shot hole, drag a marker, edit a shot's details, add off the last shot, near-green add → putt prompt.
  - Confirm the live round Map tab is unaffected.

Branched off `dev`. Not merging — review + on-device pass first.

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -902,6 +902,7 @@ export default function RoundIndex() {
         <PastRoundMap
           roundId={round.id}
           userId={user.id}
+          completed={round.completed_at != null}
           holes={sortedHoles}
           holeScores={holeScores}
           shots={shots}
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -113,6 +113,10 @@ export default function RoundIndex() {
   // back into the live HoleScreen route on tap (mode=past would drop
   // the player into the place-ball state machine on a finalized round).
   const [shotsForHole, setShotsForHole] = useState<HoleRow | null>(null)
+  // When the past-round map's "Edit this shot" opens the shots sheet, this
+  // jumps it straight into that shot's editor (vs the shot list). Cleared on
+  // close so the scorecard drill-down still opens to the list.
+  const [editShotId, setEditShotId] = useState<string | null>(null)
   const shareCardRef = useRef<View>(null)
   const { unit } = useUnits()
   const { user } = useAuth()
@@ -901,9 +905,17 @@ export default function RoundIndex() {
           holes={sortedHoles}
           holeScores={holeScores}
           shots={shots}
+          unit={unit}
           courseCenter={courseCenter}
           holeNumber={mapHole}
           onHoleChange={setMapHole}
+          onEditShot={(shotId) => {
+            const hole = sortedHoles.find((h) => h.number === mapHole)
+            if (hole) {
+              setEditShotId(shotId)
+              setShotsForHole(hole)
+            }
+          }}
           onShotUpserted={(s) =>
             setShots((prev) => {
               const i = prev.findIndex((x) => x.id === s.id)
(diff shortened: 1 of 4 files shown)


cner-smith:
## Review — feat(mobile): past-round map → review stepper + edit-this-shot (#593)

**Concerns**
- `apps/mobile/components/round/PastRoundMap.tsx` (`SHOT_RESULT_SHORT` / `summarizeShot`) — the inline comment justifies keeping these in the app component by citing the 3-caller rule ("two callers, so not worth a shared export"). That rule governs *extraction* decisions; it doesn't override the package-boundary rule. CONTRIBUTING.md is separate on this: "Web and mobile apps wire UI; they do not own logic. Don't add helpers to … `apps/mobile/lib` that other surfaces would want to reuse — lift them to `@oga/core`." `summarizeShot` is a format helper; the web surface may eventually want shot summaries too. The justification in the comment should reference the package-boundary rule, not the 3-caller rule, and explain why this *specific* function is too display-specific for core.
- `apps/mobile/components/round/PastRoundMap.tsx` (`summarizeShot`, `isPutt` heuristic) — `const isPutt = row.lie_type === 'green' || row.club === 'putter'` is defined locally. I can't tell from the diff whether `@oga/core` already exports a canonical `isPutt`- or `isOnGreen`-style helper; if it does, this silently diverges from it. Worth a quick check before this lands.

**Nits**
- `apps/mobile/components/round/PastHoleShotsSheet.tsx` (new `useEffect`) — `// eslint-disable-next-line react-hooks/exhaustive-deps` has its explanation in the preceding block comment rather than inline. CONTRIBUTING.md: "No `// eslint-disable-next-line` without a comment explaining why." The reasoning is solid; it just needs to move inline, e.g. `// eslint-disable-next-line react-hooks/exhaustive-deps -- sortedShots intentionally omitted: re-saving shot details must not re-open the editor`.

---
Looked good: targets `dev`, scope held to one concern, `tapToPlaceBall` defaults `true` so the live round is byte-for-byte unchanged, dead `ModeChip` + aim-debounce timer cleaned up completely.

---

PR #598, opened 2026-06-13
polish(mobile): render intended emphasis on Android (fontWeight → bodyBold)

cner-smith:
Fixes the "Android isn't rendering all the fonts correctly in the right places" QA note (device QA 2026-06-12).

## Root cause
On Android, RN **ignores inline `fontWeight` when a custom `fontFamily` is set** — the family name picks the file. `TYPE.body` = Epilogue-Regular (400), and Epilogue bundles **no 500/600 weight**. So `<Text style={[TYPE.body, { fontWeight: '500'|'600' }]}>` (used for button labels, titles, links) rendered **plain regular on Android** but heavier on iOS. That's both the cross-platform inconsistency *and* why emphasized text "doesn't stand out" on Android.

## Fix
The 11 sites that intend emphasis → `TYPE.bodyBold` (Epilogue-Bold 700, the bundled emphasis face), dropping the now-dead inline `fontWeight`. The single muted `See all rounds →` caption stays regular (its 500 was a no-op anyway). **Weight/style only — no color/size/spacing touched.**

Sites: login/signup/onboarding CTAs, ConfirmDialog & delete-account Cancel, My Bag row + club-name title, practice article links, stats bar % labels, map nav chevrons.

## Scope
Deliberately narrow: only the **visible-divergence class** (`TYPE.body` + emphasis weight). The ~200 other inline `fontWeight` uses render correctly on Android already (family name wins) — sweeping them would be churn with no visible change, so they're left alone.

## On "making small text pop"
Per DESIGN.md, muted text (`#8A8B7E`) is intentionally muted — pop comes from serif headlines, big serif numbers, and sparing accent, not from recoloring tokens. This PR's real-emphasis-on-Android *is* the in-system pop win; I did not do a speculative global contrast bump. If you want specific labels promoted, point me at them.

## Verification
Mobile typecheck green. Not on-device — visual; QA: buttons/titles read bold on Android, unchanged on iOS.

Branched off `dev`. Not merging — review first.

### DIFF
--- a/apps/mobile/app/(app)/bag.tsx
+++ b/apps/mobile/app/(app)/bag.tsx
@@ -574,7 +574,7 @@ function ClubRow({
         <Text style={[TYPE.body, { color: '#8A8B7E', fontSize: 16 }]}>⠿</Text>
       </Pressable>
       <View style={{ flex: 1 }}>
-        <Text style={[TYPE.body, { color: '#1C211C', fontSize: 16, fontWeight: '500' }]}>
+        <Text style={[TYPE.bodyBold, { color: '#1C211C', fontSize: 16 }]}>
           {club.name}
         </Text>
         <Text style={{ ...KICKER, marginTop: 2 }}>
--- a/apps/mobile/app/(app)/practice.tsx
+++ b/apps/mobile/app/(app)/practice.tsx
@@ -308,7 +308,7 @@ function FocusAreas({ areas }: { areas: StoredFocusArea[] }) {
               asChild
             >
               <Pressable hitSlop={6} style={{ marginTop: 6 }}>
-                <Text style={[TYPE.body, { color: ACCENT, fontSize: 13, fontWeight: '600' }]}>
+                <Text style={[TYPE.bodyBold, { color: ACCENT, fontSize: 13 }]}>
                   {a.article.title} →
                 </Text>
               </Pressable>
--- a/apps/mobile/app/(app)/profile.tsx
+++ b/apps/mobile/app/(app)/profile.tsx
@@ -389,7 +389,7 @@ export default function ProfileTab() {
         >
           <View>
             <Text style={{ ...KICKER, marginBottom: 2 }}>Equipment</Text>
-            <Text style={[TYPE.body, { color: '#1C211C', fontSize: 16, fontWeight: '500' }]}>
+            <Text style={[TYPE.bodyBold, { color: '#1C211C', fontSize: 16 }]}>
               My Bag
             </Text>
           </View>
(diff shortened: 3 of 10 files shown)


cner-smith:
## Review — polish(mobile): render intended emphasis on Android (fontWeight → bodyBold)

**Concerns**
- `apps/mobile/app/(auth)/login.tsx:136`, `signup.tsx:179`, `onboarding.tsx:250,271`, `bag.tsx:577`, `practice.tsx:315`, `profile.tsx:396`, `components/round/MapBottomChrome.tsx:321` — The PR description says "unchanged on iOS" but this looks like an untested assumption. In RN, iOS applies synthetic weight when `fontWeight` is set alongside a custom `fontFamily`, so the pre-PR render at `fontWeight: '500'` or `'600'` was a synthesized intermediate — heavier than Epilogue-Regular but lighter than Epilogue-Bold. After this change, all 11 sites render actual 700 on iOS, which is noticeably heavier than before. The Android fix is logically correct; just flagging that iOS is not "unchanged" and the PR hasn't been on-device verified.

- `apps/mobile/components/ui/ConfirmDialog.tsx:111`, `apps/mobile/app/(app)/profile.tsx:663` — The Cancel labels in the destructive confirmation flows go from a synthesized medium-ish weight to full Epilogue-Bold 700 on both platforms. Bold Cancel competing with the primary destructive action is an unusual pattern. The PR's stated goal is "render the *intended* emphasis" — worth confirming that 700 was actually the intent for secondary Cancel actions, since the original `fontWeight: '500'` reads more like de-emphasis that just never worked on Android.

- `apps/mobile/app/(app)/stats.tsx:661` — `fontVariant: ['tabular-nums']` is kept alongside `TYPE.bodyBold`. The tabular-nums feature keeps digit widths consistent in the scoring bar's percentage labels, preventing layout jitter as values change. Whether it survives the family switch depends on whether the static Epilogue-Bold.ttf includes the `tnum` OpenType table. The regular face presumably has it; the bold face may not. Can't verify from the diff — worth a quick spot-check with the font file or a visual scan at varied percentages on device.

---
Looked good: scope held to one concern, targets `dev`, no package boundary violations, `TYPE.bodyBold` is a well-defined existing token in `typography.ts`, the muted "See all rounds →" caption correctly stays at `TYPE.body`

---

cner-smith:
Review addressed in 5ecbc61:

- **Bold Cancel → reverted.** `ConfirmDialog` + delete-account `Cancel` are back to `TYPE.body` (regular). You're right — bold Cancel competes with the destructive action; the original `'500'` was intended de-emphasis that never rendered on Android. The primary-action promotions stand.
- **"Unchanged on iOS" — correction acknowledged.** You're correct: iOS synthesizes weight, so the pre-PR `500/600` sites rendered an intermediate, and they now render true 700 — *heavier* on iOS, not unchanged. The Android fix is still right; the description overstated iOS. Noting it here rather than rewriting the body.
- **tabular-nums on Epilogue-Bold (`stats.tsx`) — deferred to device QA.** Kept `bodyBold` (the % labels should read clearly), but whether the bold `.ttf` carries the `tnum` table can't be confirmed from the diff. Flagging for the on-device pass; if digits jitter at varied percentages, that one site reverts to `TYPE.body`.

PR #600, opened 2026-06-13
polish(mobile): subtle staggered entrance animations across the main screens

cner-smith:
Adds the "extra pop" from device-QA (2026-06-12), kept on-brand and now applied across the app.

## What
A shared **`<Entrance index={n}>`** wrapper (`components/ui/Entrance.tsx`) applies a gentle Reanimated `FadeInDown` (380ms, staggered `index*80ms`) to each screen's major content blocks, so they settle in as they mount / as data arrives instead of popping.

Applied to **6 screens**: home, stats, patterns, practice, rounds, drills — wrapping each screen's top-level `Section`s / logical groups (home: 5, stats: 6, patterns: 5, practice: 5, drills: 3, rounds: 1 list container).

## Restraint (per DESIGN.md)
One direction, short duration, no bounce/loop — a settle-in, not a show. Timing is centralized in `Entrance.tsx` (one place to tune).

## Correctness notes
- **Measurement safety:** blocks whose child uses `onLayout` for scroll positioning are left **unwrapped** (an entrance translate would corrupt the measured y) — only the home `LearnPreview` had this; it's untouched.
- **Lists** are wrapped as a single container (no per-item entrance — janky on long lists). No `FlatList` is involved.
- **Conditionals:** the `<Entrance>` sits *inside* each `{cond && …}` guard so an empty animated view never mounts.
- **Reduce-motion:** Reanimated 3.17 auto-skips entering animations under the OS Reduce-Motion setting (accessible by default).

## Cross-check
Reviewed by an independent agent pass: **clean — no blockers/high/med**. Verified additive-only (no logic/style/data changes), balanced JSX, no measurement/layout/list/flex regressions, typecheck exit 0. Lone Low note: reduce-motion is implicit (library default) rather than explicit — optional follow-up, not a bug.

## Verification
Mobile typecheck green. Not on-device — QA: open each tab, confirm blocks settle in gently with no flicker/jank on Android + iOS; confirm the home "yardage book" scroll-to hint still lands correctly.

Branched off `dev`. Not merging — review + on-device pass first.

### DIFF
--- a/apps/mobile/app/(app)/drills.tsx
+++ b/apps/mobile/app/(app)/drills.tsx
@@ -5,6 +5,7 @@ import { getDrills } from '@oga/supabase'
 import type { Database } from '@oga/supabase'
 import type { BlockType, PlanCategory } from '@oga/core'
 import { AppBar } from '../../components/ui/AppBar'
+import { Entrance } from '../../components/ui/Entrance'
 import {
   BLOCK_TYPE_LABEL,
   CATEGORY_LABEL,
@@ -73,6 +74,7 @@ export default function Drills() {
     <View style={{ flex: 1, backgroundColor: CREAM }}>
       <AppBar eyebrow="Practice" title="Drill library" />
       <ScrollView contentContainerStyle={{ padding: 18, paddingBottom: 48 }}>
+        <Entrance index={0}>
         <Pressable hitSlop={6} onPress={() => router.back()} style={{ marginBottom: 16 }}>
           <Text style={{ ...KICKER, color: INK_MUTE }}>← Practice plan</Text>
         </Pressable>
@@ -86,7 +88,9 @@ export default function Drills() {
           Every drill the plan generator can draw from. Each one explains the why,
           the how, and the rep target — no gimmicks.
         </Text>
+        </Entrance>
 
+        <Entrance index={1}>
         <FilterRow
           label="Part of the game"
           active={category}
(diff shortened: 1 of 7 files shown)


cner-smith (commented):
## Review — polish(mobile): subtle staggered entrance animation on the home screen

**Concerns**
- (no specific line) — The PR description explicitly marks on-device testing as outstanding ("Not on-device"). Animation PRs with multi-block Reanimated sequences warrant at least one Android device pass before merge. JS-thread animations are more prone to frame drops on lower-end Android than iOS, and a 5-block stagger with 380ms durations is worth watching under real conditions before it lands.

**Nits**
- `apps/mobile/app/(app)/index.tsx:251` (approx) — The `{pending > 0 && ...}` sync-pending banner sits between the CTAs block (delay 160) and the SGTrendChart block (delay 240) with no `entering` prop. On first load with a pending sync, it pops in while its neighbors are still fading. May be intentional — ephemeral status UI doesn't need to settle in with the content — but the PR description doesn't call it out the way the `LearnPreview` `onLayout` exception is explained. Worth a note either way.
- Branch prefix `polish/` — CONTRIBUTING.md documents `feature/`, `fix/`, and `chore/` as the naming convention. Low stakes but worth following for consistency.

---
Looked good: single-file scope, no `package.json` changes (Reanimated pre-existing via HoleMap), `onLayout` view correctly left outside any translating wrapper (preserving scroll-to measurement), targets `dev`, single concern throughout.

---

cner-smith:
Review nits noted:
- **pending-sync banner has no `entering`** — intentional: it's ephemeral status UI between the CTA (delay 160) and chart (delay 240) blocks; status shouldn't settle in with content. Leaving as-is.
- **`polish/` branch prefix** — fair, `feature/fix/chore` is the convention; can't cleanly rename a live PR, will use a sanctioned prefix next time.
- **On-device jank pass** — agreed, that's the gating QA item before merge.

Issue #599, opened 2026-06-13
Perf: cache screen data to kill refetch-on-revisit (screen-to-screen loading)


cner-smith:
From device QA 2026-06-12: "any other performance gains to minimize the loading from screen to screen? there's still a little bit of that."

## Verified findings (an audit flagged some phantom wins — corrected here)
- ❌ **"Parallelize Home profile+rounds"** — NOT a real win. The two `getProfile`/`getRecentRounds` `.then()` calls in `app/(app)/index.tsx:91,100` already fire concurrently (neither awaits the other). `Promise.all` would change nothing.
- ❌ **"Fold round-detail shots into the initial fetch"** — can't. Shots are queried by `hole_score_id IN (...)` (`round/[id]/index.tsx:244`), so they depend on `holeScores` arriving first. The `useFocusEffect` placement is also load-bearing for #246 (end-round sync self-heal). Leave it.

## The real cause + real fix (M)
There is **no caching anywhere** — every screen refetches from scratch on each visit (verified: no React Query/SWR/context cache; only `usePracticePlan` reuses within-screen). So navigating away and back always shows a spinner. The user-visible "screen-to-screen loading" is this.

Proposed, in priority order:
1. **Lightweight in-memory cache for rounds list + round detail** (module-level `Map` or a small context). Show cached data instantly on revisit, refetch in background. **Must** invalidate on every mutation (score edit, shot add/edit/remove, delete, finalize) — stale-data correctness is the whole risk, esp. the editable scorecard. M.
2. **Stats L5/L10/L20 client-side filter** — fetch the max window once, filter in memory instead of refetching per toggle (`stats.tsx:60`). S–M.
3. **Pass list-row data as route params → instant header/summary** on round detail while the full fetch runs (avoids blank flash). M.
4. **Loading skeletons** — perceived-speed win, BUT a shimmer skeleton may fight DESIGN.md's paper aesthetic; design first. M.

Recommend #1 + #2 as the meaningful wins; do them deliberately (not as a launch-week rush) given the invalidation surface.

PR #602, opened 2026-06-13
perf(mobile): render-side + query-payload optimizations (safe pass)

cner-smith:
A pre-build performance pass. Three independent read-only agent audits (render, data-fetch, startup/native), each finding cross-checked against current code before any edit. **Only Low-risk, behavior-preserving wins** here — the two Med-risk items (gating off-screen share-card mounts; lazy-loading the Home victory chart) were deliberately left out, and the broader cache layer is tracked separately (#599).

## Render
- **`useUnits` memoization** *(highest leverage).* `toDisplay`/`toDisplayFt` were plain function decls → fresh identity every render, which silently defeated downstream memos. Concretely it made `BreadcrumbLayers`' `segmentFeatures` (`deps: [segments, toDisplay]`) **re-serialize the breadcrumb GeoJSON across the Mapbox bridge on every render**, fighting the existing aim-drag throttle. Now `useCallback`/`useMemo` keyed on `unit` (formatters are pure over it).
- **`round/[id]` scorecard** — the `PastHoleShotsSheet` shots prop did `shots.filter()` in the render body; now an O(1) lookup on the `shotsByHoleScoreId` Map that already exists there.
- **`bag.tsx`** — per-row `bag.filter().length` (O(n²), ran every drag frame) → one precomputed `club_type→count` Map.

## Startup
- **Dropped `Inconsolata-Regular.ttf`** — zero consumers (`mono` → `Inconsolata-Medium`). 105 KB + one fewer font to parse before the native splash dismisses (`useFonts` blocks on all of them).

## Query payload
- **Narrowed the biggest shot queries**: `getShotsForUser` (caps at **1000 rows**) → `DISPERSION_SHOT_COLUMNS` (11 cols; web+mobile dispersion read only coords+club), `getShotsByClub` (200) → `PATTERN_SHOT_COLUMNS` (19; union of what web `useShotPatterns` + mobile `patterns` read). Added `getRoundsList` (minimal projection) for the **500-row** all-rounds list, which never reads the per-category SG columns.
- Shared constants untouched; each consumer was verified column-by-column, and the `as const` string literals make **TypeScript enforce no over-drop** (a dropped column a consumer reads → compile error).

## Deferred (noted, not done)
- `getCourseTees` is fetched twice on a past-round open (parent share-card tee + child `TeePicker`) — de-duping needs a small `TeePicker` refactor (it owns its own fetch/add/refresh), more than a one-liner. Worth a follow-up.
- Off-screen share-card mount gating + Home chart lazy-load — Med risk (behavior/cold-start), left for a deliberate pass.

## Verification
- `pnpm typecheck` (web + `@oga/core` + `@oga/supabase`) green — proves the narrowed columns still satisfy **web** consumers.
- Mobile `tsc` green. Mobile's `@oga/supabase` is a live symlink, so the EAS build sees the new export.
- No behavior change intended; the map breadcrumb + dispersion math read the same fields. Worth a quick on-device sanity (map breadcrumb labels, patterns screen, all-rounds list) on the build.

Branched off `dev` (chore/ prefix). Not merging — review first.


cner-smith (commented):
## Review — perf(mobile): render-side + query-payload optimizations (safe pass)

**Nits**
- `packages/supabase/src/queries/shots.ts` (block comment above `PATTERN_SHOT_COLUMNS`) — the parenthetical "(getShotsForUser caps at 1000 rows; getShotsByClub at 200)" restates what the function signatures already show (`limit = 1000`, `limit = 200`). CONTRIBUTING.md: "Don't narrate what the code does." The `as const` type-enforcement note in the same block is the non-obvious *why* and is worth keeping; the row-cap numbers are the redundant part.

---
Looked good: `useCallback`/`useMemo` dep arrays in `useUnits` are correct; `as const` on the new column constants gives TypeScript compile-time enforcement of no over-drop; `getRecentRounds` is still live (`index.tsx` home screen uses it for the 20-row SG breakdown, so the narrower `getRoundsList` is the right additive approach); `Inconsolata-Regular.ttf` removal is clean (file deleted, `useFonts` entry removed, no orphaned references); PR targets `dev`, single concern.

---

PR #603, opened 2026-06-13
fix(mobile): unbreak Android entrance animations + SG-trend Y-axis labels

cner-smith:
Two fixes from on-device QA of the dev build — Android broke badly, iPhone was fine.

## 1. Android entrance animations collapsed the layout (regression from #600)
On Android the `<Entrance>` blocks rendered **overlapping/stacked**, then **vanished to blank**; iOS was perfect.

**Root cause (researched, sourced):** Reanimated's declarative `entering`/`FadeInDown` *layout-animation* subsystem is unreliable on Android (old arch). Late- or conditionally-mounted ScrollView children race the native initial-value snapshot → blocks paint at their final position (overlap) or get dropped (blank); `.delay()` widens the window; `translateY` makes the overlap visible; Android view-flattening drops the driven node. iOS doesn't flatten and snapshots synchronously. Refs: reanimated [#8445](https://github.com/software-mansion/react-native-reanimated/issues/8445), [#5492](https://github.com/software-mansion/react-native-reanimated/issues/5492), [#3992](https://github.com/software-mansion/react-native-reanimated/issues/3992), [#5715](https://github.com/software-mansion/react-native-reanimated/issues/5715).

**Fix — keeps the animation on both platforms** (per request, not disabled on Android): rewrote `Entrance` to drive the same staggered fade+rise with a **manual mount animation** — `useSharedValue` + `withDelay(withTiming)` via `useAnimatedStyle`, plus `collapsable={false}`. This is the *core* shared-value engine `HoleMap` already uses reliably — it sidesteps the broken layout-animation subsystem entirely. Same `Entrance` API, the 6 screens are untouched, and it respects `useReducedMotion`.

## 2. SG-trend Y-axis labels didn't match the lines (pre-existing, Stats + Home)
The trend lines use `monotoneX` (no overshoot) — the steep edges are *real* outlier rounds. But the charts hardcoded `tickValues` (`[-1.5..1.5]` / `[-0.5,0,0.5]`) with **no `domain`**, so Victory auto-scaled the domain to the outliers while the labels stayed fixed and crammed into a useless central band.

**Fix:** new pure helper `@oga/core` **`symmetricNiceTicks(values)`** → a symmetric, nice-rounded domain + tick set scaled to the data peak (tested, 5 cases). Both charts now set `domain` + `tickValues` from it, so labels always match the plotted range. (You chose "labels match full range.")

## Verification
- `pnpm typecheck` (web + core + supabase) ✅ · `pnpm --filter @oga/core test` → **548 passed** (5 new) ✅ · mobile `tsc` ✅
- The Android animation fix is inherently on-device-verified by the next build — that's the point.

Branched off `dev`. Not merging without your okay — but this is what gates a clean rebuild.


cner-smith (commented):
## Review — fix(mobile): unbreak Android entrance animations + SG-trend Y-axis labels

**Concerns**
- `CONTRIBUTING.md` — this PR bundles two independent bug fixes: the Android `Entrance` animation regression and the SG-trend Y-axis label mismatch. CONTRIBUTING.md is explicit: "One concern per PR." The PR description acknowledges both came from the same QA session, but different root causes, different files, and different packages normally warrant separate PRs so each can be reverted cleanly. If they genuinely gate the same build and splitting is impractical right now, that reasoning is worth stating explicitly for the record.

- `packages/core/src/chart-axis.ts` — `symmetricNiceTicks` ships with two call sites in this diff. CONTRIBUTING.md's 3-caller rule says "don't extract a helper … unless it has three or more callers, or the inline version is genuinely unreadable." The algorithm is ~12 lines of non-trivial tick math — I think the readability exception holds — but it's the grey zone the rule is designed to flag. If a third caller (web charts, a stats summary tile) is already planned, noting that in the PR description would close the question.

**Nits**
- `apps/mobile/components/home/SGTrendChart.tsx:32` — `symmetricNiceTicks(data.map((d) => d.y))` runs naked in the render body. `stats.tsx` wraps the equivalent call in `useMemo`; this one should too. The function is pure and cheap, but the inconsistency will catch someone off-guard, and if a parent re-renders on scroll or focus events the map+tick calculation fires every time.

- `packages/core/src/chart-axis.ts:18` — the `?? 10 * pow` fallback is unreachable. `pow` is defined as `10^⌊log₁₀(rawStep)⌋`, so `rawStep / pow` always lands in `[1, 10)`, and `.find(c => c >= rawStep)` will always match one of `[1, 2, 2.5, 5, 10] × pow` before exhausting the array. The branch is dead code; removing it saves a future reader from wondering when it fires.

---
Looked good: `symmetricNiceTicks` is correctly homed in `@oga/core` (not `apps/mobile/lib`), the new export ships with five focused tests, `Entrance`'s public API is unchanged so all six screens are untouched, `collapsable={false}` is the right Android view-flattening guard, PR targets `dev`, no secrets, no RLS migrations touched.

---

cner-smith:
Review addressed:

**Nits — fixed in 98c1777:**
- `SGTrendChart` axis calc now wrapped in `useMemo` (parity with `stats.tsx`).
- Dropped `10` from the candidate steps in `chart-axis.ts` so the `?? 10*pow` fallback is the genuine (5,10)×pow branch, not dead code. Tests still green (548).

**Concerns — rationale:**
- **One-concern-per-PR:** acknowledged — these are two independent fixes (animation regression vs chart axis). They're bundled deliberately: both surfaced in the same on-device QA of the dev build and **gate the same urgent rebuild**, and each is small + independently tested. Happy to split into two PRs if preferred for clean revertability — say the word and I'll separate them before merge. Otherwise the coupling is intentional and temporary (this build only).
- **3-caller rule on `symmetricNiceTicks`:** it has 2 call sites today, but they live in **two separate files** (the `stats` route + the `SGTrendChart` component), so an inline helper can't be shared between them without a module regardless — and pure chart/axis math belongs in `@oga/core` per the package-boundary rule, not duplicated in two app files. The readability exception also holds (non-trivial tick math, worth the 5 tests). A 3rd caller (web SG-trend chart) is plausible later but not committed.

PR #604, opened 2026-06-14
feat(mobile): per-club distance stats (max/min/avg total) on Stats + Patterns

cner-smith:
"How far do I hit each club" — one of the most-loved golf stats. Adds per-club distance to the Stats and Shot Patterns pages.

## Data (reuses the existing pipeline)
New pure `@oga/core` **`clubDistanceStats(shots)`**: per club, **min / max / avg / count** of TOTAL distance = `haversineYards(start → end)` — the exact computation the existing driver-avg stat already does. Excludes putts + greenside (`putter` / `lie_type === 'green'`), clamps 3–450 yd to drop GPS-zero/duff noise, sorted longest-first. Takes a flat shot list so both surfaces feed it the same way. **Total distance, not carry** (we only have ball positions) — labeled as such. Tested (3 cases).

## Stats page — "Club distances" table
```
CLUB DISTANCES        Total distance, not carry · avg (min–max)
Driver     258   231–276   12
3 wood     232   214–245    7
7 iron     163   148–171    9
PW         118    96–129    6
```
Avg in serif (per DESIGN), min–max + sample count alongside. Hidden when there's no shot-coord data yet (like proximity).

## Patterns page — per-club readout
Adds **Avg distance** + **Range** to the selected club's Pattern summary — visible even below the 5-shot dispersion threshold (distance only needs start+end coords, not aim).

## Notes
- Unit-aware (yd/m) via the user's prefs.
- Driver-avg tile in Ball Striking still stands (headline); this adds the full bag.
- Additive — no schema/data/query changes; flows through the same `rounds`/`shots` already loaded.

## Verification
`pnpm typecheck` (web + core + supabase) ✅ · `pnpm --filter @oga/core test` → **551 passed** (3 new) ✅ · mobile `tsc` ✅. Not on-device yet.

Branched off `dev`. Not merging without your okay.

### DIFF
--- a/apps/mobile/app/(app)/patterns.tsx
+++ b/apps/mobile/app/(app)/patterns.tsx
@@ -7,6 +7,8 @@ import {
   CLUBS,
   LIE_SLOPES,
   LIE_TYPES,
+  YARDS_TO_METERS,
+  clubDistanceStats,
   computeDispersion,
   computeDispersionStats,
   dispersionVerdict,
@@ -121,6 +123,10 @@ export default function Patterns() {
   }, [shots, lieType, lieSlope])
 
   const stats = useMemo(() => computeDispersionStats(points), [points])
+
+  // Total distance for the selected club (across all its shots, not the
+  // lie-filtered dispersion set). null until there are tracked shots.
+  const clubDist = useMemo(() => clubDistanceStats(shots)[0] ?? null, [shots])
 
   const shareCardRef = useRef<View>(null)
   const [sharing, setSharing] = useState(false)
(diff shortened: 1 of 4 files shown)


cner-smith:
## Review — feat(mobile): per-club distance stats (max/min/avg total) on Stats + Patterns

**Concerns**
- `apps/mobile/app/(app)/patterns.tsx:127` — `clubDistanceStats(shots)[0]` is fragile. The function groups by club and returns sorted by longest average first; `[0]` is only the right answer if `shots` is already filtered to the selected club (which the comment implies, but can't be confirmed from the diff alone). If `shots` ever contains multi-club data, this silently shows the wrong club's distances — no error, no warning. A more defensive form would be:
  ```ts
  const clubDist = useMemo(
    () => clubDistanceStats(shots).find((c) => c.club === selectedClub) ?? null,
    [shots, selectedClub],
  )
  ```
  If `selectedClub` isn't already in scope at that point in the component, that's worth surfacing.

**Nits**
- `packages/core/src/stats.ts:544–551` — The block comment above `CLUB_DISTANCE_MIN_YARDS` narrates the implementation. Per CONTRIBUTING.md: "Comments only when the *why* is non-obvious from the name and types alone. Don't narrate what the code does." The only non-obvious why here is "total not carry since we only have ball positions"; that's worth one line. The rest ("sorted longest-first so a bag reads driver → wedges", "takes a flat shot list so both surfaces feed it the same way") describes the what.
- `packages/core/src/stats.ts:543–552` (`ClubDistanceStat` JSDoc) — `min`, `max`, `avg`, and `count` don't need docstrings; the names convey meaning. The unit ("yards") on the numeric fields is genuinely non-obvious and worth a single `// yards` trailing comment or a note on the interface. "reliability hint" on `count` is editorial and can go.
- `apps/mobile/app/(app)/stats.tsx:558–559` — `conv` is defined inside the `.map()` callback, so a new closure is allocated per club per render. Hoisting it above the `.map()` (it closes over only `unit`, not `c`) is a one-line fix.

---
Looked good: `clubDistanceStats` and `ClubDistanceStat` live in `packages/core/src/stats.ts` as required; new export has three test cases covering sort order, exclusions, and the distance clamp; scope held to one concern; targets `dev`.

---

PR #607, opened 2026-06-20
fix(mobile): widen scorecard FIR/GIR columns + fix splash brand font

cner-smith:
On-device QA surfaced two mobile issues; both fixed here.

## Scorecard FIR/GIR columns
The FIR (Fairway in Regulation) and GIR (Green in Regulation) columns were cramped at 28px, crowding the mono kicker headers and the check/dot indicators. Widened both to 40px (header + data rows), taking the extra 24px from the flex Hole column, which had ample room. Header and data cells stay aligned.

`apps/mobile/app/(app)/round/[id]/index.tsx`

## Login splash brand font
The `oga.` wordmark and tagline were rendering in a system fallback font on Android instead of the Fraunces italic brandmark. Cause: a redundant `fontStyle: 'italic'` declared alongside `fontFamily: FONT.serifItalic` (`Fraunces-MediumItalic`, an already-italic registered face). On Android, requesting an italic style against a family whose face is already italic can make the font manager fail resolution and fall back to a system serif/sans. Removed the redundant `fontStyle` from both styles; the italic is baked into the named face.

`apps/mobile/app/(auth)/welcome.tsx`

## Testing
- Verified on iOS preview build (on-device).
- Mobile-only RN styling changes; no `@oga/core` or web surface touched.

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -772,10 +772,10 @@ export default function RoundIndex() {
             <Text style={{ ...KICKER, width: 48, textAlign: 'right', color: '#8A8B7E' }}>
               Putts
             </Text>
-            <Text style={{ ...KICKER, width: 28, textAlign: 'center', color: '#8A8B7E' }}>
+            <Text style={{ ...KICKER, width: 40, textAlign: 'center', color: '#8A8B7E' }}>
               FIR
             </Text>
-            <Text style={{ ...KICKER, width: 28, textAlign: 'center', color: '#8A8B7E' }}>
+            <Text style={{ ...KICKER, width: 40, textAlign: 'center', color: '#8A8B7E' }}>
               GIR
             </Text>
             <Text style={{ ...KICKER, width: 76, textAlign: 'right', color: '#8A8B7E' }}>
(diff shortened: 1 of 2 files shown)


cner-smith (commented):
## Review — fix(mobile): widen scorecard FIR/GIR columns + fix splash brand font

**Concerns**
- Two separate bugs fixed in one PR. CONTRIBUTING.md §Workflow says "**One concern per PR.** UI restyle, bug fix, and new feature do not belong in the same PR." The scorecard column fix (`round/[id]/index.tsx`) and the splash font fix (`welcome.tsx`) are unrelated defects in unrelated files — the PR description even separates them into distinct sections. Both fixes are trivially small, which may be why they landed together, but the rule doesn't carve out an exception for small changes. Whether to ask for a split is the maintainer's call; flagging per the contribution guidelines.

**Nits**
- `apps/mobile/app/(auth)/welcome.tsx:117,124` — the `fontStyle: 'italic'` lines are removed with no inline note explaining why. The Android dual-italic footgun (specifying italic style on a face whose name already encodes the italic axis causes the font manager to fail resolution and fall back) is genuinely non-obvious and would likely puzzle a future reader during a routine style audit. A short comment on one of the two removed lines would preserve the diagnostic reasoning the PR description contains.

- The splash font failure is Android-specific per the PR description, but testing was iOS-only ("Verified on iOS preview build"). Worth an Android check before merge if a device or emulator is handy — the fix is sound on its face, but the failure mode (font manager fall-through) is platform-specific.

---
Looked good: all three width-bearing places updated consistently — FIR header (`index.tsx:775`), GIR header (`index.tsx:779`), and the shared data-row map (`index.tsx:851`) — so header and cell alignment stay in sync; `fontStyle` removal is correct for a pre-italic named face; PR targets `dev`; no `@oga/core` exports, no migrations, no security surface touched.

---

PR #608, opened 2026-06-21
fix(mobile): deleting an active round no longer bricks round viewing

cner-smith:
## Symptom
With a live round in progress, going to the home screen and deleting that round from the recent-rounds list (without exiting live mode) bricked the app: it showed `cannot coerce the result to a single JSON object`, and afterward **no other round could be viewed**.

## Root cause
1. The round fetch used `.single()` in three places (`round/[id]/index.tsx` detail loader + SG refetch, `hole/useHoleData.ts` live loader). `.single()` throws PGRST116 on a 0-row result — exactly what a just-deleted round returns.
2. The `[id]` detail loader's effect never reset `error` / `redirectToLive` at the start of a run, so once the deleted round set `error`, the `if (error || !round)` guard stayed tripped for **every** round opened afterward → "can't view any other rounds".

## Fix
- Round fetches switched to `.maybeSingle()` — a deleted round resolves to `null` and surfaces cleanly as "Round not found" instead of throwing.
- The detail loader resets `loading` / `error` / `redirectToLive` up front each run so a prior round's error can't leak onto the next.

## Testing
- Mechanism confirmed from source; on-device repro recommended (delete the active live round from home, then open another round → should load normally, no PGRST116).

## Notes
Mobile-only. No `@oga/core` / web / migration surface touched. Does not address tearing down the still-mounted live session of the deleted round (it now shows a graceful "Round not found / exit" screen rather than bricking) — that hardening can be a follow-up if desired.

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -137,8 +137,8 @@ export default function RoundIndex() {
     setRedirectToLive(false)
     ;(async () => {
       try {
-        // maybeSingle (not single): a deleted round returns null rather than
-        // throwing PGRST116 "cannot coerce the result to a single JSON object".
+        // A deleted round returns null rather than throwing PGRST116 "cannot
+        // coerce the result to a single JSON object".
         const { data: r, error: rErr } = await supabase
           .from('rounds')
           .select('*, courses(name, lat, lng)')
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -128,13 +128,22 @@ export default function RoundIndex() {
   useEffect(() => {
     if (!id) return
     let active = true
+    // Reset per-round state up front so a previously-viewed round's error or
+    // live-redirect can't leak onto the next one — e.g. after the active round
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — fix(mobile): deleting an active round no longer bricks round viewing

**Nits**
- `apps/mobile/app/(app)/round/[id]/index.tsx:133` and `apps/mobile/components/round/hole/useHoleData.ts:131` — Both `// maybeSingle (not single):` comments open by narrating what changed rather than why. CONTRIBUTING says "Don't narrate what the code does." The PGRST116 explanation that follows is the genuinely non-obvious part; the `(not single)` preamble can be dropped without losing anything useful.

- `apps/mobile/app/(app)/round/[id]/index.tsx:~460` — The SG refetch block checks `if (rRes.data)` but never inspects `rRes.error`, so real errors (network failure, permission denied) are silently swallowed. Pre-existing issue, not introduced here — worth a follow-up issue.

---
Looked good: state reset correctly placed before the async IIFE so a prior round's `error`/`redirectToLive` can't bleed across navigations; `maybeSingle()` cleanly lets the existing `if (rErr || !r)` guard surface "Round not found" without false-positive PGRST116 noise; fix is mobile-only with no `@oga/core` or migration surface touched; PR targets `dev`, one concern.

---

PR #609, opened 2026-06-21
fix(mobile): on-map tee box always derives from the first shot, not stored coords

cner-smith:
## Decision
Product call: the on-map tee box should **always** anchor to where the player actually teed off — their **first shot's start** — never the surveyed `holes.tee_lat/tee_lng`. Real tee boxes move daily, so the stored coordinate misrepresents where you hit from. Stored tee data stays useful only for course rating / slope / handicap (`course_tees`), which is a separate concept from the map marker.

## Symptom
On a hole with map-placed shots, the tee marker stayed pinned to the stored survey coordinate and did **not** track the player's first shot.

## Root cause (live path)
`LiveRoundSession` built the tee origin as `data.tee ?? (shotNumber === 1 ? ball : null)`. `data.tee` resolves to `previousShots[0] ?? storedTee`, and `previousShots` excludes the current ball — so while placing shot 1, `data.tee === storedTee` (non-null on OSM-mapped holes) and the `??` short-circuited, pinning the box to survey coords.

## Fix
- Reorder to `(shotNumber === 1 ? ball : null) ?? data.tee` — the live first-shot ball wins on shot 1 (reactive), the **saved** first-shot start anchors it afterward, and the stored course tee is consulted **only** as a last resort before any first shot exists (so the hole isn't marker-less on entry).
- Past/review path (`PastRoundMap`, `placed[0].start ?? tee`) was already correct — comment only, documenting the locked decision.

## Testing
- On-device repro recommended: place/adjust your first shot → tee box should track it; stored survey coords should no longer pin it.

### DIFF
--- a/apps/mobile/components/round/LiveRoundSession.tsx
+++ b/apps/mobile/components/round/LiveRoundSession.tsx
@@ -175,12 +175,10 @@ export default function LiveRoundSession({
   // teed off — the FIRST shot's start — never the surveyed holes.tee_lat (real
   // tee boxes move daily; the stored coord lies about where you hit from).
   // While placing shot 1 that's the live ball as you drag it, so the marker
-  // tracks it reactively (this is the fix: previously `data.tee ?? …` let the
-  // stored tee win on shot 1 and the box stayed pinned to survey coords while
-  // the player adjusted their drive). After shot 1, `data.tee` is the SAVED
-  // first-shot start (= previousShots[0]). The stored course tee inside
-  // `data.tee` is consulted ONLY as a last resort — before any first shot
-  // exists — so the hole isn't marker-less on entry.
+  // tracks it reactively. After shot 1, `data.tee` is the SAVED first-shot
+  // start (= previousShots[0]). The stored course tee inside `data.tee` is
+  // consulted ONLY as a last resort — before any first shot exists — so the
+  // hole isn't marker-less on entry.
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — fix(mobile): on-map tee box always derives from the first shot, not stored coords

**Concerns**
- `apps/mobile/components/round/LiveRoundSession.tsx:178–180` — The parenthetical `(this is the fix: previously \`data.tee ?? …\` let the stored tee win on shot 1 and the box stayed pinned to survey coords while the player adjusted their drive)` embeds bug-fix history in source. CONTRIBUTING.md: "Don't narrate what the code does." That history already lives in the commit message and PR body — in source it will rot and mislead the next person to touch this line. Consider keeping only the timeless "why" sentences: the real-tee-moves-daily rationale and the last-resort fallback note. The fix mechanism itself is now self-evident from the operand order.

**Nits**
- `apps/mobile/components/round/PastRoundMap.tsx:295` — "Dragging shot 1 moves `placed[0].start`, so the box tracks it reactively." narrates what `placed[0]?.start ?? tee` does rather than why that's the right source. Same CONTRIBUTING.md note; the sentence can be dropped without losing meaningful context.

---
Looked good: one-line operand swap is correct and minimal, targets `dev`, no new `@oga/core` exports or migrations, `any` and `eslint-disable` absent, single concern held throughout.

---

Issue #484, opened 2026-06-01
[mobile] Live-round logging feels heavy — reduce per-shot steps / offer a minimal tracking flow


cner-smith:
First real on-course test (2026-06-01), two testers. The per-shot live-round flow — **mark ball → set aim → confirm aim → pick club → hit → mark result** — was enough back-and-forth that one tester gave up tracking after 3 holes ("it took me out of the round, context-switching back and forth"). The other tester's summary: *"seems like it'll be a great post-round tool, but a bit overwhelming for the live round."* Both noted it's a brand-new app for them and may ease with familiarity.

This rubs against the project's own **speed-first live-round principle** (CLAUDE.md: "Max 3 taps to log a shot; lie/slope never blocking; 'Skip all — just track location' always visible; never require input before the shot is hit").

## Directions to explore (future session — not a quick fix)
- Audit actual taps-per-shot vs the 3-tap target; cut steps / confirmations (e.g. auto-confirm aim unless changed).
- A genuinely **minimal live mode**: mark ball location (+ optional aim), defer ALL metadata (club/lie/result) to a fast post-round review — leaning into the "great post-round tool" strength.
- Make "skip the rest / just track location" frictionless and prominent at every step.
- Related: #465 (first-run walkthrough → familiarity), #460 (club picker), #456 (drag-anywhere target).

Capturing the strongest product signal from the first real round.

cner-smith:
Superseded by the epic #791, which carries the concrete design + sequenced plan that came out of the mockup review. The original tester signal (the reason this matters) stays recorded here for reference; tracking and sub-steps move to #791.

PR #610, opened 2026-06-21
fix(mobile): live-round revisited holes show shot breadcrumb + explicit 'Add a shot' (#484)

cner-smith:
## Symptom
During a live round, navigating **back** to a previously-played hole dropped the player into the live append-shot flow: the only actions were "Mark ball" (logging a *new* shot) and "Finish hole", and the live GPS ball + auto-aim redrew a line-to-green over the already-logged shots.

## Root cause
A live round always mounts `LiveRoundSession`, which has no review mode — the #593 shot-stepper lives only in `PastRoundMap`, which a live round never reaches. So a revisited played hole re-entered the `PLACE_BALL → SET_AIM` append state machine.

## Fix (Option C — owner-selected)
Treat a revisited played hole as breadcrumb-only review until the player explicitly opts in:
- New `isRevisitingPlayedHole = hasPriorShots && !appendEngaged` (`hasPriorShots` = this hole already has logged shots). `appendEngaged` resets on every hole change and is set true by **marking a ball** (keeps aids on across the natural shot-to-shot loop) and by the new **"+ Add a shot"** button.
- While revisiting: the auto-aim spawn, the GPS ball watcher/marker, and the GPS location puck are all gated off — so the line-to-green and live ball can't render; the existing shot breadcrumb stays visible.
- `MapBottomChrome` shows a "+ Add a shot" primary CTA (built with the existing `PressableTouch` + static-style + `android_ripple` pattern) plus the finish affordance. Tapping it re-arms the live append flow for that hole.

## Testing
- On-device repro recommended: in a live round, go back to a played hole → should show the shot breadcrumb only (no roaming ball/line), with "+ Add a shot" to resume logging.

## Notes
Mobile-only. Did not pursue Option B (composing `PastRoundMap` into the live session) per the owner's choice. Minor self-correcting flicker possible when navigating from an empty hole to a played one before `hasPriorShots` resolves; end state always correct.

### DIFF
--- a/apps/mobile/components/round/LiveRoundSession.tsx
+++ b/apps/mobile/components/round/LiveRoundSession.tsx
@@ -525,7 +525,10 @@ export default function LiveRoundSession({
           }
           showLocationPuck={
             finalState.roundState !== 'SHOT_DETAIL' &&
-            finalState.roundState !== 'PUTTING'
+            finalState.roundState !== 'PUTTING' &&
+            // Hide the live GPS puck while revisiting a played hole — that hole
+            // is in breadcrumb-only review mode until "Add a shot" (#484).
+            !finalState.isRevisitingPlayedHole
           }
           onSetAim={(loc) => {
             // A user drag / long-press is an explicit aim — mark it touched so
@@ -607,6 +610,7 @@ export default function LiveRoundSession({
             finalState.ball != null &&
             !ballMoved
           }
+          isRevisitingPlayedHole={finalState.isRevisitingPlayedHole}
           totalShotsThisHole={totalShotsThisHole}
           holeNumber={holeNumber}
           holeCount={data.holeCount}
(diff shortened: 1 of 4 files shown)


cner-smith (commented):
## Review — fix(mobile): live-round revisited holes show shot breadcrumb + explicit 'Add a shot' (#484)

**Concerns**
- `apps/mobile/components/round/MapBottomChrome.tsx:44` — `onAddShot` is typed as required while `isRevisitingPlayedHole` is optional. The two props are semantically a pair — `onAddShot` is only reachable through the `if (p.isRevisitingPlayedHole)` branch. Making it required forces every caller to supply a handler even when they never set `isRevisitingPlayedHole={true}`. Changing to `onAddShot?: () => void` would make the interface consistent with its optional partner and remove the burden from any future second consumer of `MapBottomChrome`.

**Nits**
- `apps/mobile/components/round/hole/useHoleState.ts` (block comment above the `appendEngaged` declaration) — the multi-sentence paragraph re-explains logic that the variable names, the JSDoc on the exported members, and the single line `isRevisitingPlayedHole = hasPriorShots && !appendEngaged` already express. CONTRIBUTING.md: "Don't narrate what the code does." The async-load caveat (`hasPriorShots can arrive a beat late without flipping the verdict`) is the part worth keeping; the structural description around it can go.

- The `onAddShot` handler in `LiveRoundSession` calls two state setters in sequence (`setAppendEngaged(true)` + `setRoundState('PLACE_BALL')`). They always need to travel together to re-arm the live flow; nothing currently enforces that. A named `engageAppend()` action in `useShotActions` — alongside the `markBallHere` action that also calls `setAppendEngaged(true)` — would encode the invariant where the rest of the hole-state actions live and make it harder to call one without the other in a future touch.

---
Looked good: single-concern fix, targets `dev`, `appendEngaged` reset on every hole change is clean, no new `@oga/core` exports or migrations, scope held tightly to the revisit case.

---

PR #614, opened 2026-06-28
fix(mobile): prompt for a pin on synthetic-course holes

cner-smith:
## What

On synthetic courses (no `holes` rows — the majority of prod courses), the live map hit the `missingHoleLayout` path: it showed only a **passive** banner ("No hole layout… place shots manually") and **explicitly suppressed** the actionable `PinFirstCta`. Without a pin there are no distances, no expected strokes, and no dispersion — the core live-round value stayed dark, which made the flow feel broken.

## Change

- Surface the existing `PinFirstCta` ("Set the pin (below) to see your To Hole distance, expected strokes, and shot pattern") whenever **no pin is placed** — synthetic or mapped holes alike. The "Place pin" button is already in the left toolbar.
- Retire the passive `MissingLayoutBanner`. Once a per-hole pin is dropped, `missingHoleLayout` flips false and the HUD lights up anyway, so the prop is now dead — removed from `HoleMap` and both callers (`LiveRoundSession`, `PastRoundMap`).

Net −37 lines, reuses existing components, no new state. Non-blocking — the player is nudged but can still skip.

## Testing

- `npm run typecheck` clean.
- On-device QA pending (the user is QA).

### DIFF
--- a/apps/mobile/components/round/HoleMap.tsx
+++ b/apps/mobile/components/round/HoleMap.tsx
@@ -25,7 +25,6 @@ import { AimDistancePill } from './markers/AimDistancePill'
 import { useHoleCamera } from './hooks/useHoleCamera'
 import {
   ExpStrokesPill,
-  MissingLayoutBanner,
   PinFirstCta,
   TeeBadge,
   ToHolePill,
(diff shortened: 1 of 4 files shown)


cner-smith (commented):
## Review — fix(mobile): prompt for a pin on synthetic-course holes

---
Looked good: dead prop and component fully swept from interface, destructuring, import, and both callers; `PinFirstCta`/`ToHolePill` mutual exclusion holds; pin placement is reachable from `PastRoundMap` in both logging and review modes; scope held to one concern; targets `dev`.

---

PR #615, opened 2026-06-28
fix(mobile): resolve Fraunces italic on Android app-wide

cner-smith:
## Problem
Custom fonts are selected by family **name** only. Layering `fontStyle: 'italic'` or `fontWeight` on an already-named face makes the Android font manager fail to resolve it and fall back to **system sans** — iOS tolerates it, so the bug only showed on Android. Result: every italic in the app (handicap index, stats numbers, practice/drill names, learn articles, round screens) rendered as plain sans on Android while looking correct on iOS. Caught comparing the iOS vs Android store screenshots.

## Fix
- **90 sites** already had a named italic face (`TYPE.serif`, `TYPE.bodyItalic`, `FONT.serifItalic`/`FONT.bodyItalic`) → stripped the redundant `fontStyle: 'italic'` + sibling `fontWeight`.
- **3 sites** were bare `fontStyle: 'italic'` with no face → swapped to `fontFamily: FONT.bodyItalic`.
- **93 sites, 41 files.** Matches the known-good splash wordmark pattern (bare `fontFamily`, nothing layered).

## fontVariant
The sweep incidentally dropped `fontVariant: ['tabular-nums']` where it sat in the same style object as a removed `fontStyle`. The remaining ~45 `tabular-nums` (scorecard/stats number columns) are **left in place** — on-device verification (Android EAS build) confirms named-face numbers with `tabular-nums` render Fraunces correctly, so it is **not** an Android fallback trigger and the column alignment is worth keeping.

## Verification
- `npm run typecheck` passes.
- **Android EAS build installed on-device — Fraunces italic now renders across the app. ✅**
- mono/kicker/bodyBold weights untouched.

### DIFF
--- a/apps/mobile/app/(app)/_layout.tsx
+++ b/apps/mobile/app/(app)/_layout.tsx
@@ -94,8 +94,6 @@ export default function AppLayout() {
           style={[TYPE.serif, {
             color: '#F2EEE5',
             fontSize: 20,
-            fontWeight: '500',
-            fontStyle: 'italic',
             marginBottom: 10,
             textAlign: 'center',
           }]}
--- a/apps/mobile/app/(app)/bag.tsx
+++ b/apps/mobile/app/(app)/bag.tsx
@@ -379,8 +379,6 @@ export default function BagScreen() {
               style={[TYPE.serif, {
                 color: '#1C211C',
                 fontSize: 22,
-                fontWeight: '500',
-                fontStyle: 'italic',
                 marginBottom: 18,
               }]}
             >
--- a/apps/mobile/app/(app)/drills.tsx
+++ b/apps/mobile/app/(app)/drills.tsx
@@ -119,7 +119,7 @@ export default function Drills() {
               {filtered.length} drill{filtered.length === 1 ? '' : 's'}
             </Text>
             {filtered.length === 0 ? (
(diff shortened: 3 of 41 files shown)


cner-smith:
## Review — fix(mobile): resolve Fraunces italic on Android app-wide

**Concerns**
- `apps/mobile/components/home/SGBreakdown.tsx:103` — `fontVariant: ['tabular-nums']` is retained after the `fontStyle`/`fontWeight` removal. The PR description lists this property as "also an Android fallback trigger" and explicitly states the intent is to remove it everywhere, with the note "Correct font > tabular alignment." The other numeric displays that previously carried this property — `StatTile`, `StatRow`, `HomeTile`, `SGValue` in `RecentRoundsList`, `GreenDiagram`, several article stats — all had it stripped. This one looks like a missed site. If the SG bar intentionally keeps tabular alignment for layout reasons, the description should say so; if not, this should be removed to match the stated fix.

**Nits**
- `apps/mobile/components/learn/primitives.tsx:325` — The nested `<Text>` for `r.by` inside `ResourceList` still carries `fontStyle: 'normal'` and `fontWeight: '400'`. Those were reset guards against the parent's former `fontStyle: 'italic'` and `fontWeight: '500'`. After this PR strips both from the parent, the resets are dead. The `fontFamily: FONT.body` override is still necessary and correct; just the two reset props can go.

---
Looked good: no security issues, no package-boundary violations, no `@oga/core` changes requiring tests, single-concern scope, targets `dev`.


---

PR #616, opened 2026-06-28
feat(web): club distances on Stats (parity with mobile)

cner-smith:
## What
Mobile's Stats screen shows per-club **total distance** (avg + min–max range + shot count) via `@oga/core` `clubDistanceStats`, but the web Stats page never got it. This adds a **Club distances** section at the bottom of web Stats, matching mobile.

## How
- New `ClubDistancesSection` (`apps/web/src/pages/stats/sections/`), reusing the shared `Section` + `useUnits` + `formatClubLabel`.
- Computed in `StrokesGainedPage` from `stats.rounds` (the rounds-with-shots the page already loads) — same flatten the mobile screen uses. No new fetch.
- Reuses the existing `@oga/core` `clubDistanceStats` (already unit-tested).

## Verify
- `pnpm typecheck` passes.
- Eyeball on the Vercel preview (signed-in → Stats, scroll to bottom). Shows avg with unit + bare min–max range + shot count per club, longest-first.

Closes the web/mobile club-distance parity gap.

### DIFF
--- a/apps/web/src/pages/stats/StrokesGainedPage.tsx
+++ b/apps/web/src/pages/stats/StrokesGainedPage.tsx
@@ -1,4 +1,5 @@
-import { useState } from 'react'
+import { useMemo, useState } from 'react'
+import { clubDistanceStats } from '@oga/core'
 import { useDetailedStats } from '../../hooks/useDetailedStats'
 import { EmptyState, Skeleton } from './components/Section'
 import { Segmented } from './components/Segmented'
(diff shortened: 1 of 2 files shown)


cner-smith (commented):
## Review — feat(web): club distances on Stats (parity with mobile)

**Concerns**
- `apps/web/src/pages/stats/sections/ClubDistancesSection.tsx` — this component has exactly one caller (`StrokesGainedPage`). CONTRIBUTING.md is explicit: "Single-caller 'components' stay co-located in the parent file." The existing `sections/` directory already houses other probable single-caller sections (Scoring, BallStriking, ShortGame, Patterns), so this PR follows that established pattern — but the pattern itself is the issue. If the project has tacitly accepted it as an exception, that exception should be documented or the rule updated; otherwise the minimum fix is to inline the ~60 lines into `StrokesGainedPage`.

- `apps/web/src/pages/stats/StrokesGainedPage.tsx:17` — the `useMemo` calls `stats.rounds.flatMap(…)` without a null guard on `stats.rounds` itself. The inner spreads have guards (`r.hole_scores ?? []`, `hs.shots ?? []`), but if `useDetailedStats` initializes `rounds` as `undefined` rather than `[]` before the query resolves, this throws on the loading render — before `stats.status === 'success'` can protect anything. Unclear from this diff alone; if `useDetailedStats` always returns an array, ignore this. If it can be `undefined`, add `(stats.rounds ?? []).flatMap(…)`.

**Nits**
- `apps/web/src/pages/stats/sections/ClubDistancesSection.tsx:37` — `borderBottom: '1px solid #EBE5D6'` is a hardcoded hex value while every other color in the file uses a Tailwind token (`text-caddie-ink`, `text-caddie-ink-dim`, `text-caddie-ink-mute`). If there's a Tailwind border token for this shade, prefer it so the value tracks the design system.

- `apps/web/src/pages/stats/sections/ClubDistancesSection.tsx:1–5` — the file-top comment block narrates what the component does ("Per-club total distance… Mirrors the mobile Stats section…"). Per CONTRIBUTING.md, comments should explain the *why* that can't be inferred from names and types. The inline comment explaining `conv` vs `toDisplay` (to avoid repeating "yd" three times) is the right shape — keep that one, drop the block.

---
Looked good: `@oga/core`'s `clubDistanceStats` reused without duplication; no new core export means the test-coverage rule isn't triggered; PR targets `dev`; scope held to one concern (no mixed concerns); no security issues, no domain pillar violations.

---

PR #617, opened 2026-06-28
fix(web): sidebar uses the oga. brandmark

cner-smith:
The app sidebar showed a plain uppercase **OGA**; the rest of the brand (landing nav, splash) uses the lowercase **oga.** wordmark (Fraunces italic + upright period). Mirror it in the sidebar.

Period is cream on the dark sidebar — matching the splash, since the forest `#1F3D2C` period would be invisible against `#1C211C`. Typography itself was already correct on web (Tailwind: Fraunces/Epilogue/Inconsolata, all loaded); this was just the wordmark.

### DIFF
--- a/apps/web/src/components/layout/Sidebar.tsx
+++ b/apps/web/src/components/layout/Sidebar.tsx
@@ -95,9 +95,13 @@ export function Sidebar() {
       >
         <div
           className="font-serif text-caddie-bg"
-          style={{ fontSize: 18, fontWeight: 500 }}
+          style={{ fontSize: 22, fontWeight: 500, fontStyle: 'italic' }}
         >
-          OGA
+          oga
+          {/* Upright period — the "o." brand mark. Cream (not forest) on the
+              dark sidebar, matching the splash treatment where the forest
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(web): sidebar uses the oga. brandmark

**Nits**
- `apps/web/src/components/layout/Sidebar.tsx:107` — `style={{ fontStyle: 'normal' }}` on the new `<span>` could be `className="not-italic"` instead. The span has no sizing that forces it into the inline style object, and the element already uses `className` for font properties (`font-serif`). Small step toward keeping Tailwind the single source for font style on this element.
- `apps/web/src/components/layout/Sidebar.tsx:105-107` — The comment's opening clause ("Upright period — the 'o.' brand mark.") narrates what the code produces. Per CONTRIBUTING.md: "Don't narrate what the code does." The color rationale that follows ("Cream (not forest) on the dark sidebar…") is the non-obvious *why* worth keeping — the first clause can be trimmed.

---
Looked good: single concern held, targets `dev`, no logic in `apps/web/src/lib/`, no `@oga/core` exports added, no security surface touched.

---

PR #618, opened 2026-06-28
docs: pre-release cleanup of public docs

cner-smith:
Pre-release sweep of **public-facing** docs (gating the dev→main release). From a docs audit cross-checked against migrations / seed.sql / @oga/core.

## Blockers fixed
- **Issue templates** (`bug`/`feature`/`refactor`): dropped the internal `CLAUDE.md` filename from the '… rules to preserve' header (renders publicly on every filed issue) → 'Project rules / invariants to preserve'.
- **CONTRIBUTING drills schema** was materially wrong. Now matches reality:
  - adds the **`drill_type` practice-mode taxonomy** (`warmup|blocked|random|skill_game|pressure_game|on_course`, migration 0033)
  - adds `goals`, `targets`, and the `source`/`source_url`/`contributor`/`verified` provenance columns
  - ⚠️ documents that **migration 0039 prunes any drill with `source IS NULL AND verified=false`** — following the old doc silently deleted contributions
  - facility uses empty array `{}` (there is no `anywhere` value)
  - drops the auto-generated `id` from the contributor field list
  - describes the structured/sourced instruction format the corpus uses
- **docs/self-hosting.md**: seed loads **125** drills, not 24.

## Also
- **CHANGELOG**: removed AI-vendor names from three historical entries (kept commit links).

Docs-only; no code touched.


cner-smith (commented):
## Review — docs: pre-release cleanup of public docs

**Nits**
- `CONTRIBUTING.md` (drill checklist, `source`/`verified` item) — "otherwise migration `0039` prunes the row" implies either field being absent is enough to trigger deletion. The warning box two lines above states the actual condition: `source IS NULL AND verified = false` — both must be absent. The checklist is stricter than the migration, which is harmless in practice, but a contributor cross-checking the two will notice a contradiction. Consider tightening to "unless both are absent, migration `0039` prunes the row" or just pointing to the warning box directly.

---
Looked good: issue template rename correctly removes the internal tool reference from public-facing templates, `self-hosting.md` drill count corrected to 125, `anywhere` facility value removed to match actual schema, CONTRIBUTING.md schema block now reflects the live columns, PR targets `dev`.

---

PR #625, opened 2026-06-28
fix(web): no-app account-deletion path for Play deletion URL

cner-smith:
Google Play's account-deletion policy requires deletion to be requestable **without the app** (uninstalled users). The `/support` and `/privacy` deletion sections only described the in-app flow (Settings → Delete account).

Adds a line to both pages: *"No longer have the app installed? Email support@oga.golf to request deletion of your account and all of its data."* (`support@oga.golf` forwarding is live.)

This lets **`https://oga.golf/support`** be used as the Play Data Safety **data-deletion URL** and satisfy the policy. Copy-only; typecheck passes.

### DIFF
--- a/apps/web/src/pages/landing/PrivacyPage.tsx
+++ b/apps/web/src/pages/landing/PrivacyPage.tsx
@@ -116,6 +116,11 @@ export function PrivacyPage() {
             stays — it isn't yours to delete.
           </P>
           <P>
+            No longer have the app installed? Email{' '}
+            <ExtLink href="mailto:support@oga.golf">support@oga.golf</ExtLink>{' '}
+            to request deletion of your account and all of its data.
+          </P>
+          <P>
             If something goes wrong with self-service deletion, open a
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — fix(web): no-app account-deletion path for Play deletion URL

**Nits**
- `apps/web/src/pages/landing/SupportPage.tsx:77` — This block uses a raw `<a>` with `style={{ color: '#1F3D2C', textDecoration: 'underline' }}`, while the identical call-to-action added in `PrivacyPage.tsx` at the same line uses `<ExtLink>`. The two pages now have the same sentence but different rendering paths — if link styles change, `SupportPage` won't track the change, and the hex color is load-bearing rather than coming from the design token. Using `<ExtLink href="mailto:support@oga.golf">` here, the same way `PrivacyPage` does, removes the divergence.

---
Looked good: copy-only change, single concern, targets `dev`, no security surface, no logic touches.

---

PR #628, opened 2026-06-28
chore(eas): pin Xcode 26 image for iOS production builds

cner-smith:
Apple requires App Store uploads to be built with **Xcode 26+** since 2026-04-28. We're on **Expo SDK 53**, whose default EAS image is Xcode 16 — so the iOS production build was rejected at submission.

Opt into Xcode 26 by pinning `"image": "macos-sequoia-15.6-xcode-26.2"` on the production iOS profile (per Expo's guidance for SDK ≤53).

⚠️ Expo notes SDK 53 isn't guaranteed compatible with Xcode 26 (depends on native libs). If the rebuild fails, the proper fix is an SDK 54 upgrade (separate effort). Config-only change.

### DIFF
--- a/apps/mobile/eas.json
+++ b/apps/mobile/eas.json
@@ -34,6 +34,9 @@
     "production": {
       "autoIncrement": true,
       "node": "20.18.0",
+      "ios": {
+        "image": "macos-sequoia-15.6-xcode-26.2"
+      },
       "env": {
         "npm_config_legacy_peer_deps": "true"
       }


cner-smith (commented):
## Review — chore(eas): pin Xcode 26 image for iOS production builds

**Concerns**
- `apps/mobile/eas.json` (production profile) — The `development` and `preview` profiles continue to use EAS's default Xcode 16 image, while `production` now pins `macos-sequoia-15.6-xcode-26.2`. The PR description already flags the known risk: SDK 53 isn't guaranteed compatible with Xcode 26. With the profiles diverging on Xcode version, any incompatibility that shows up (native module link failure, binary signing quirk, new Swift/ObjC compiler behavior) won't surface during preview distribution or TestFlight — it'll only appear at production build time, which is the worst moment to discover it. Consider pinning the same image in the `preview` profile so the pre-production cycle exercises the actual toolchain before the production build runs.

---
Looked good: change is scoped to exactly one EAS profile key, no package.json touched, targets `dev`.

---

PR #630, opened 2026-06-29
fix(web): sanitize Turnstile site key so production captcha loads

cner-smith:
## Problem

The production Android (Play Store early-access) build can't log in or sign up. The Turnstile box never loads, never verifies, and shows no error — the submit button just stays disabled.

Root cause: the production EAS env var \`EXPO_PUBLIC_TURNSTILE_SITE_KEY\` was set to \`<0x4AAAAAADJg4XtoY7a9rXJY>\` — the value was copied with the surrounding placeholder angle brackets. The app passes that key to \`https://oga.golf/captcha.html?siteKey=...\`, and \`turnstile.render()\` silently fails on the malformed key (no widget, no token, no error).

Worked in preview builds because the preview environment had a clean key; this is the first production build.

## Fix

Strip \`<\`, \`>\`, and whitespace from the site key in \`captcha.html\` before handing it to Turnstile.

Because the mobile app loads this page **remotely** from oga.golf, deploying this to production fixes **already-installed builds with no app rebuild or store resubmission** — as soon as oga.golf redeploys, existing apps can verify again.

## Follow-ups (not in this PR)
- Fix the EAS var itself to the clean value (\`0x4AAAAAADJg4XtoY7a9rXJY\`, no brackets) and rebuild — the web sanitizer is a safety net, not a substitute.
- Confirm the Supabase Auth Turnstile **secret** matches the same Cloudflare widget.
- Merge this into \`dev\` after \`main\` so the branches don't drift.

## Test
Manual: loading \`captcha.html?siteKey=<0x...>\` now renders the widget; previously blank.

### DIFF
--- a/apps/web/public/captcha.html
+++ b/apps/web/public/captcha.html
@@ -10,7 +10,15 @@
     body { display: flex; align-items: center; justify-content: center; }
   </style>
   <script>
-    const siteKey = new URLSearchParams(window.location.search).get('siteKey');
+    // Strip stray angle brackets / whitespace before handing the key to
+    // Turnstile. A site key copied from a "<your-key-here>" style placeholder
+    // (e.g. into the EXPO_PUBLIC_TURNSTILE_SITE_KEY env var) arrives wrapped in
+    // "<...>", which Turnstile rejects by silently failing to render — no
+    // widget, no error, submit button stuck disabled. Sanitizing here fixes
+    // already-shipped app builds without a rebuild, since the app loads this
+    // page remotely.
+    const rawSiteKey = new URLSearchParams(window.location.search).get('siteKey');
+    const siteKey = (rawSiteKey || '').replace(/[<>\s]/g, '');
 
     function initTurnstile() {
       turnstile.render('#widget', {


cner-smith (commented):
## Review — fix(web): sanitize Turnstile site key so production captcha loads

**Concerns**
- PR targets `main` instead of `dev`. CONTRIBUTING.md: "Open a PR against `dev` (not `main`)… contributors don't open PRs against `main`." The PR description notes the branches will need to be re-synced afterward, which confirms the deviation is known. If this is an intentional maintainer hotfix to unblock production, that's a reasonable call — but it's worth stating explicitly so the process deviation is on the record rather than implied by a follow-up bullet.

**Nits**
- `apps/web/public/captcha.html` — The 7-line comment narrates what Turnstile does internally ("rejects by silently failing to render — no widget, no error, submit button stuck disabled") and how the bug manifested. CONTRIBUTING.md: "Comments only when the *why* is non-obvious… Don't narrate what the code does." The first sentence covers the why; the rest belongs in the PR description (where it already lives) rather than the source file.

- Fix `EXPO_PUBLIC_TURNSTILE_SITE_KEY` in EAS to the clean value, as noted in the follow-ups. The sanitizer is a good safety net, but leaving the malformed value in place means any future audit of env vars will see a key that looks wrong, and a different class of copy-paste mistake (missing prefix, wrong widget ID) would still fail silently.

---
Looked good: regex handles the null case correctly via `|| ''`; no `@oga/core` changes, no migrations touched, scope held to one file and one concern.

---

PR #631, opened 2026-06-29
fix(web): sanitize Turnstile site key in captcha WebView page

cner-smith:
Ports the production hotfix #630 onto `dev` so the branches don't drift.

`captcha.html` strips `<`, `>`, and whitespace from the Turnstile site key before passing it to `turnstile.render()`. A key copied into `EXPO_PUBLIC_TURNSTILE_SITE_KEY` with placeholder brackets (`<0x4AAA...>`) silently fails to render the widget, blocking login/signup with no error.

Already merged to `main` and live on oga.golf; this keeps `dev` in sync. No app rebuild needed — the mobile app loads this page remotely.

### DIFF
--- a/apps/web/public/captcha.html
+++ b/apps/web/public/captcha.html
@@ -10,7 +10,15 @@
     body { display: flex; align-items: center; justify-content: center; }
   </style>
   <script>
-    const siteKey = new URLSearchParams(window.location.search).get('siteKey');
+    // Strip stray angle brackets / whitespace before handing the key to
+    // Turnstile. A site key copied from a "<your-key-here>" style placeholder
+    // (e.g. into the EXPO_PUBLIC_TURNSTILE_SITE_KEY env var) arrives wrapped in
+    // "<...>", which Turnstile rejects by silently failing to render — no
+    // widget, no error, submit button stuck disabled. Sanitizing here fixes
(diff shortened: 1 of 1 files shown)


cner-smith (commented):
## Review — fix(web): sanitize Turnstile site key in captcha WebView page

**Nits**
- `apps/web/public/captcha.html:13` — The comment's opening sentence ("Strip stray angle brackets / whitespace before handing the key to Turnstile.") narrates what the code does. CONTRIBUTING.md says: "Don't narrate what the code does." The two sentences that follow explain the *why* — that a key copied from a `<your-key-here>` placeholder silently breaks the widget, and that fixing it here (rather than in app code) covers already-shipped builds because the page is loaded remotely. Those sentences are justified; the first one isn't. Dropping just that line satisfies the rule.

- If `rawSiteKey` is null (param missing) or the sanitization strips every character (e.g. `siteKey=<   >`), `siteKey` becomes `''` and `turnstile.render` fails silently — the same symptom the fix addresses. A short guard before `initTurnstile()` (e.g. early-return with a visible error message when `!siteKey`) would surface misconfiguration instead of hiding it. Out of scope here since the stated goal is just the angle-bracket case.

---
Looked good: single-concern fix, targets `dev`, correct null-guard on the param read, regex handles the documented failure mode, no logic placed in the wrong package.

---

Issue #457, opened 2026-05-29
[data] Real fairway/green/bunker course geometry (polygons)


cner-smith:
Prerequisite for honest by-lie strokes-gained. Most OGA courses are synthetic (no per-hole layout). To know whether a target lands in fairway/green/rough/bunker we need course polygons (OSM/import/draw). Relates to #385 (web Course Planner) and enables the live-map by-lie SG sheet + dispersion-weighted SG.

cner-smith:
## Concrete build path (from Course Planner #385 scoping, 2026-07-01)

The pipeline is further along than this ticket implies. `scripts/crawl/holes-fetcher.ts` **already** fetches `golf=green`/`golf=tee`/`golf=hole` from Overpass with `out geom tags` (line 111) — i.e. it already receives the full green/tee **polygon** node coordinates. It then **discards them to centroids** (`geomCentroid`, lines 88–90) to fill the scalar `tee_lat/lng` + `pin_lat/lng` columns. So most of the fetch machinery exists.

**Build (path 1 — extend the existing crawler, near-free):**
1. Add area tags to the Overpass query: `golf=fairway`, `golf=bunker`, `golf=rough`, `golf=water_hazard`/`lateral_water_hazard`, `natural=water`.
2. Stop discarding geometry — store rings in a new `hole_features(course_id, hole_number, feature_type, polygon jsonb)` table as GeoJSON. **No PostGIS needed** at this scale (a few dozen polygons/course, tested against a handful of landing points); do point-in-polygon client-side in a pure `@oga/core` helper `lieAt(point, features): LieType` (ray-casting, precedence green→sand→water→fairway→rough).
3. Consume: `sgAtAim` (#385) swaps its distance-only `expectedStrokesFromPin(d)` for `lieAt(landing) -> getShotCategory(lie, ...)`. Real hazard-aware SG, **UI unchanged**.

**Coverage caveat:** only well-mapped OSM courses tag fairway/bunker/water — famous/urban courses are rich, many munis have only a hole centerline. OSM gets the well-mapped subset for free; a **community draw-tool** (course-creator idea + #548 moderation model) fills the long tail; ML segmentation of satellite tiles is a someday.

Sequencing: fast-follow after the #385 Course Planner MVP — it lifts SG accuracy app-wide (not just the planner) and proves the `hole_features` schema before any community draw-tool builds on it.

cner-smith:
### Logged-shot lie points are a second data source for these polygons

[...]

Issue #385, opened 2026-05-22
feat(web): Course Planner — pre-round strategy mode with dispersion overlays


cner-smith:
## Background

OGA already collects everything needed to be a real **pre-round strategy tool**: per-club shot dispersion ellipses (`packages/core/src/shot-patterns.ts`), course hole maps (`courses`, `holes` tables with tee/pin GPS), and SG baselines (`packages/core/src/sg-baselines.ts`). What it doesn't have is the *planning* surface — the place where a player loads a course they're about to play and figures out, hole by hole, where to actually aim.

ShotPattern (#361) does this live-round, on a phone, between shots. OGA's natural angle is **pre-round, on a laptop, the night before** — calmer setting, reflective rather than reactive, better fit for the planning task. Mobile is intentionally out of scope: planning is not a phone activity.

This is the highest-leverage way to make OGA's existing dispersion data *actionable* for the player, instead of just showing them a histogram after the fact.

## Scope

**Web-only.** No mobile parity. The course planner is a desktop/laptop surface (route: `/plan/:courseId/:holeNumber?`).

Player workflow:

1. Search for a course (same search UX as `/rounds/new`)
2. Load the course's holes with map data
3. Per hole: see the hole on a satellite map, overlay the player's per-club dispersion ellipses on the map at typical aim points, drag aim points around to see live SG-at-aim feedback (mirroring ShotPattern's strongest feature per #361 research)
4. Save a per-hole strategy as text + aim coordinates if the player wants — or just use it as a transient exploratory tool

## Why this fits OGA's positioning

- ShotPattern is **paid**, **iOS-only**, **live-round**. OGA Course Planner is **free**, **web**, **pre-round**. Distinct positioning, distinct user moment.
- Leverages the dispersion data OGA already has (#372, #243) without needing new measurement infrastructure.
- Builds on the SG-at-aim primitive (the standout ShotPattern feature) but ships it in a slower, more reflective context where the player has time to actually engage.

## Three-system shape

| Layer | What lives here |
|---|---|
| `@oga/core` | Pure helpers — already have most: `shotCurve()` (#243), `peerBracketDispersion()` (#372), `computeRoundSG()`. New: `sgAtAim(holeData, club, aimLatLng, userDispersion)` — pure function returning expected SG for an aim point. |
| `apps/web` | All UI. Routes, components, Mapbox rendering, drag interactions, save state. |
| `apps/mobile` | **None.** Web-only by explicit scope. |

[...]

PR #633, opened 2026-07-01
feat(web): Course Planner MVP — pre-round strategy tool (#385)

cner-smith:
## Course Planner MVP (web-only)  ·  #385

A read-only, pre-round strategy tool. Load a mapped course, plan an ordered sequence of shots per hole on a satellite map, drag aim points, and see **expected strokes to hole out** — Monte-Carlo'd over your own per-club dispersion. Web-only by design (planning is a laptop activity). No persistence, no migrations, no new deps.

Route: `/plan` (course search) → `/plan/:courseId/:holeNumber` (per-hole planner). New top-level **Planner** nav item.

### What's here
- **`@oga/core`** (pure, tested): `sgAtAim` + `expectedStrokesFromPin` (Monte-Carlo expected-strokes over real dispersion) and `coneRingGeoJSON` (95% dispersion-cone polygon). 8 new tests; full core suite 559/559 green.
- **Web:** `useCoursePlanData` hook, `CoursePlanPage` (search), `HoleStrategyMap` (fresh Mapbox satellite — per-leg dispersion dots + 95% cone + tee→aim→pin route), `SGFeedbackCard`, `HolePlanPage` (leg-list state + hole-total), route + nav.

### How it works
Each hole is planned as ordered "legs" (shots). Leg 1 starts at the tee; each later leg starts from the previous leg's aim. Per leg you get your dispersion overlay + expected-strokes-from-there; the hole total = `(earlier shots) + last-leg expected-strokes-to-hole-out`, shown vs par. Legs are chained from *intended* aims (how golfers actually plan), so a leg's cone shows risk without feeding the score, while moving an earlier aim relocates the next leg's start.

### Scope / constraints (intentional MVP)
- **Mapped courses only** — courses without hole geometry show "not mapped yet."
- **Real dispersion only** — clubs need ≥5 logged shots; no-data users get an empty state.
- **Read-only** — nothing persists (saving plans is a deferred phase).

### Notes for review
- **Baseline-seam clamp:** the shared SG baselines dip across the 30→50 yd seam (around-green@30 > approach@50 in every bracket), so `expectedStrokesFromPin` floors the ≥30 yd branch to stay monotonic. `sg-baselines.ts` is left untouched; the underlying table bug is filed as **#632**.
- **Defensive hardening (no repro reached):** `HolePlanPage`'s course+hole-scoped leg-reset guard and memoized `tee`/`pin` prevent a plan-wipe / marker-churn that isn't reachable in current code but would surface if the dispersion hook ever refetched.
- **Fast-follow #457** (real lie-type polygons) will replace the distance-only SG with hazard-aware categories — a drop-in swap, UI unchanged.

### Follow-ups (non-blocking, out of scope here)
- Route-line dark contrast outline on bright satellite tiles.
- Strengthen the `sgAtAim` convexity test to use equal-mean-distance clouds.
- Browser QA on real dev data (signed-in, mapped course, dispersion drag).

Verified: `pnpm test` 559/559 · `pnpm typecheck` clean · `pnpm --filter @oga/web build` clean.


cner-smith (commented):
## Review — feat(web): Course Planner MVP — pre-round strategy tool (#385)

**Blockers**
- `apps/web/src/components/plan/HoleStrategyMap.tsx:176-182` — the camera-recenter effect is gated by `initialPositionDoneRef` (declared line 126), which is set `true` once and never reset. `/plan/:courseId/:holeNumber` doesn't remount `HolePlanPage`/`HoleStrategyMap` when only `holeNumber` changes (React Router keeps the same element mounted across a param-only navigation), so `goToHole`'s Prev/Next buttons update `tee`/`pin`/`legs`/markers correctly but the map view stays parked on whichever hole loaded first. On any course where holes aren't clustered together, browsing past hole 1 leaves the map pointed at the wrong (or empty) patch of ground — this breaks the hole-to-hole navigation that's a headline interaction of this PR. Suggest keying the "have we positioned" guard on the course/hole (same pattern as `initedKeyRef` in `HolePlanPage.tsx`) instead of a fire-once ref.

**Concerns**
- `apps/web/src/pages/plan/HolePlanPage.tsx:158` — `// eslint-disable-next-line react-hooks/exhaustive-deps` has no comment explaining why, unlike the other two disables in this same file (lines 84 and the map-init one), which both have detailed justification above them. CONTRIBUTING.md: "No `// eslint-disable-next-line` without a comment explaining why." Worth a line noting why `pin?.lat, pin?.lng` is used instead of `pin`.
- `apps/web/src/pages/plan/HolePlanPage.tsx:76-81` and `:107-112` — the initial aim (and each new leg's aim, via `addLeg`) is computed automatically from the club's median carry distance along the tee→pin/aim→pin bearing, rather than starting from a user-placed point. The user can drag it afterward, but the default is an algorithmically suggested aim point. CONTRIBUTING.md lists "aim point is always explicit user input" as one of the project's four design pillars. I think this may be intended to scope to live-round shot tracking rather than a read-only pre-round planner, but it's close enough to the pillar's wording that it seems worth a maintainer call rather than assuming it's fine.
- `packages/core/src/sg-at-aim.ts:39-40` (call sites `apps/web/src/pages/plan/HolePlanPage.tsx:138`, `:150`) — the `sgAtAim` parameter is named `tee`, but for every leg after the first it's actually the previous leg's aim point, not the tee. This is a public `@oga/core` export other surfaces will call — the name will mislead future callers into thinking it must be the hole's tee. Consider `origin` (matches `coneRingGeoJSON`'s naming for the same concept).

- `expectedStrokesFromPin` in `packages/core/src/sg-at-aim.ts` is distance-only (no lie type), which the PR body already flags with a filed fast-follow (#457) and matches the existing distance-based `getShotCategory`/`sg-calculator.ts` model, so not a new regression — just noting it brushes up against the "lie-slope-aware tracking" pillar in case a maintainer wants to weigh in before this ships more broadly.
- `HoleStrategyMap.tsx`'s render effect tears down and recreates every Mapbox `Marker` (tee, pin, each leg's aim) on every `legs`/`focusedLeg` change instead of updating existing marker positions in place. Fine at MVP scale (a handful of legs) but worth revisiting if the planner grows more legs or gets more interactive.

---
Looked good: pure math (`sgAtAim`, `expectedStrokesFromPin`, `coneRingGeoJSON`) correctly lives in `packages/core` with tests for both new exports (559/559 passing per the PR body); no migrations, secrets, or service-role usage in this diff; scope stayed to the one stated concern (planner MVP) despite touching several new files.

---

cner-smith:
## Review — feat(web): Course Planner MVP — pre-round strategy tool (#385)

**Concerns**
- `apps/web/src/pages/plan/HolePlanPage.tsx:80-85` and `:111-116` — the leg-init effect and `addLeg` both auto-compute the initial aim point (`destinationYards(tee, bearing, carryDist)` using the player's median club carry along the straight tee→pin bearing) before the user has touched anything. CONTRIBUTING.md's "Suggesting features" section lists "aim point is always explicit user input" as one of the project's four design pillars — the highest bar for a feature to clear. The user can drag the marker afterward, but the *first* rendered aim is a suggested/computed point, not a placement the user made. Worth confirming with the maintainer whether a planning surface is exempt from this pillar (as opposed to live-round tracking, which is presumably what the rule was written for), or whether legs should start un-aimed (e.g. pinned to the tee/previous aim with 0 offset) until the user drags. The PR description pre-emptively justifies the baseline-seam clamp but doesn't address this one.
- `packages/core/src/sg-at-aim.ts:16-17` — `expectedStrokesFromPin` is explicitly "keyed on distance only (no lie data yet)," i.e. the new expected-strokes number shown to users doesn't account for lie/slope. This brushes against the "lie-slope-aware tracking" pillar. The PR description does self-disclose this (citing fast-follow #457), which is the right move per CONTRIBUTING's "say why up front" — flagging here mainly so the maintainer explicitly signs off on shipping a lie-agnostic expected-strokes number in the meantime, since it's presented in the UI (`SGFeedbackCard`) as a fairly precise-looking stat (`toFixed(1)`).
- `apps/web/src/hooks/useCoursePlanData.ts:5` — this hook has exactly one caller (`HolePlanPage.tsx`) and is a thin composition of three existing hooks (`useAuth`, `useHolesForCourse`, `useClubDispersion`). CONTRIBUTING.md's 3-caller rule: "Don't extract a helper, hook, or component into its own function or file unless it has three or more callers, or the inline version is genuinely unreadable... Single-caller 'components' stay co-located in the parent file." Inlining these three lines into `HolePlanPage` looks straightforward and wouldn't hurt readability there.

**Nits**
- `apps/web/src/pages/plan/HolePlanPage.tsx:52-71` — the `initedKeyRef` guard is described in the PR as "defensive hardening (no repro reached)" for a scenario that "isn't reachable in current code but would surface if the dispersion hook ever refetched." Per CONTRIBUTING's "ship the minimum" principle, guarding against a bug that can't currently happen is speculative; if it's ever reachable, a regression test at that point would catch it and justify the guard then.

- Route-line contrast outline on bright satellite tiles and the `sgAtAim` convexity test strengthening are already called out as follow-ups in the PR description — agree with deferring both.

---
Looked good: `sgAtAim`/`expectedStrokesFromPin`/`coneRingGeoJSON` are correctly placed in `@oga/core` with tests alongside (`sg-at-aim.test.ts`, `shot-dispersion-geo.test.ts`); no migrations or service-role access introduced; scope is held to one concern (web-only planner, explicitly no persistence); the `HoleStrategyMap` orphaned-source cleanup when a leg is removed (tracking `renderedLegCountRef`) is careful, easy to miss, correctly done.

---

cner-smith (commented):
## Review — feat(web): Course Planner MVP — pre-round strategy tool (#385)

**Concerns**
- `apps/web/src/pages/plan/HolePlanPage.tsx:80-84` and `apps/web/src/pages/plan/HolePlanPage.tsx:111-115` — the "pick a club by distance, cap carry at the remaining distance, project the aim point" logic is duplicated verbatim between the init effect and `addLeg`. CONTRIBUTING.md calls this the most-cited rule: "All shared business math, stats engines, SG calculation, units conversions, format helpers, post-round inference, and shared domain types live in `packages/core/src/`. Web and mobile apps wire UI; they do not own logic." This reads as exactly that kind of domain math (default-aim selection), not UI wiring, and as written it also ships without a test. Worth lifting to a small `packages/core` helper (e.g. `defaultLegAim(origin, pin, selectClub)`) with a unit test, then calling it from both sites.
- `apps/web/src/pages/plan/HolePlanPage.tsx:84` — the initial leg aim is computed automatically (`destinationYards(tee, bearing, carryDist)`) rather than starting from user input; the user then drags it via `HoleStrategyMap`. CONTRIBUTING.md lists "aim point is always explicit user input" as one of the project's design pillars, and asks PRs that touch a pillar to say why up front. The PR description addresses two other pillar-adjacent tradeoffs (the SG-baseline seam clamp, the defensive-hardening guards) but not this one. I think there's a reasonable case that a *planning* tool's default suggestion is different in kind from inferring aim on a *logged* shot — but since the diff doesn't make that case explicitly, flagging it so a maintainer can confirm the distinction is intentional.

**Nits**
- PR title "Course Planner MVP — pre-round strategy tool (#385)" isn't in imperative present tense as CONTRIBUTING.md's "Opening a PR" section asks for (its example: "Add lie-slope filter to patterns").

- `packages/core/src/sg-at-aim.ts` computes expected strokes from distance only, since the club dispersion type it consumes (`AimRelativeDispersion` in `shot-patterns.ts`) already drops per-point lie type when aggregating. The PR description already flags this as a known MVP gap with a fast-follow filed (#457), so just noting it lines up with that pillar's spirit for anyone landing here later.

---
Looked good: no migrations/secrets touched so no RLS surface here, scope is held to one feature (planner), and the two new `@oga/core` exports (`sgAtAim`/`expectedStrokesFromPin`, `coneRingGeoJSON`) both ship with tests exercising the interesting cases (monotonicity, the baseline-seam clamp, Jensen's-inequality direction, ring closure).

---

PR #637, opened 2026-07-02
feat(web): up-the-hole camera + two-column map rollout (round detail, mini-map, planner)

cner-smith:
Rolls the Course Planner's map treatment out to the other web Mapbox surfaces. **Web-only.**

**Stacked on #633** (`feature/course-planner`) — it needs `HoleStrategyMap`. When #633 merges to `dev`, this rebases onto `dev` and retargets. Spec: `docs/internal/web-map-uptheholecamera-rollout.md` + `docs/superpowers/specs/2026-07-01-web-map-treatment-rollout-design.md` (both gitignored/local).

## What

- **Shared helper** `apps/web/src/lib/mapbox.ts` → `fitHoleUpward(map, tee, pin)` — rotates the camera so tee→pin runs bottom→top and `fitBounds` frames both. Pure bearing math stays in `@oga/core`.
- **Round-detail `RoundMap`** (via `useMapSetup`): up-the-hole framing when tee+pin are known; north-up fallback for pin-only / course-centroid / OKC.
- **`MapView`**: two-column portrait layout — controls + per-hole shot list left, portrait map right; collapses map-first on narrow. Overlays stay map-anchored.
- **`ShotMiniMap`**: bearing on its `fitBounds` so the shot runs up-the-map.
- **`HoleStrategyMap`** (planner): refactored onto the shared helper — no behavior change.

## Not touched

`useMapLayers` drag/tap semantics, DemoMap (landing), Shot Patterns (SVG). No mobile.

## Verification

- Workspace typecheck green.
- Signed-in browser QA at 1440 + 390 (desktop two-column + phone map-first): rotated camera, shot list, overlays, planner unchanged.

### DIFF
--- a/apps/web/src/components/plan/HoleStrategyMap.tsx
+++ b/apps/web/src/components/plan/HoleStrategyMap.tsx
@@ -1,6 +1,6 @@
-import { useEffect, useMemo, useRef, useState } from 'react'
-import { bearingDegrees, coneRingGeoJSON, scatterGeoJSON, type GeoPoint } from '@oga/core'
-import { mapboxgl, MAPBOX_TOKEN_PRESENT } from '../../lib/mapbox'
+import { useEffect, useRef, useState } from 'react'
+import { coneRingGeoJSON, scatterGeoJSON, type GeoPoint } from '@oga/core'
+import { mapboxgl, MAPBOX_TOKEN_PRESENT, fitHoleUpward } from '../../lib/mapbox'
 import type { ClubDispersion } from '../../pages/rounds/hooks/useClubDispersion'
 
 export interface Leg {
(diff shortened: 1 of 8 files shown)


cner-smith (commented):
## Review — feat(web): up-the-hole camera + two-column map rollout (round detail, mini-map, planner)

**Concerns**
- `apps/web/src/pages/rounds/components/MapView.tsx:256` (two-column grid) and `:400` (new `ShotList` component) — this PR bundles three distinct changes: the shared up-the-hole camera rollout (the stated headline feature), a layout restyle of `MapView` into two columns, and a brand-new shot-list feature that didn't exist before. CONTRIBUTING.md's one-concern-per-PR rule is explicit: "UI restyle, bug fix, and new feature do not belong in the same PR." The camera rollout across `HoleStrategyMap`/`useMapSetup`/`ShotMiniMap` reads as one coherent concern; the two-column layout + `ShotList` reads as a second, separable one. Might be worth a split even though the title discloses both up front.
- No CI run exists for this PR (`https://github.com/cner-smith/opengolfapp/pull/637` shows only a Vercel deployment check, no `typecheck`/`test`/`mobile` runs). `.github/workflows/ci.yml:5` scopes `pull_request` triggers to `branches: [main, dev]`, and this PR's base is `feature/course-planner` (per the stacked-PR note in the description), so the workflow never fires here. CONTRIBUTING.md: "CI runs the same four commands. A PR with a red CI gate will not be reviewed until it's green." Nothing here suggests the code is broken — the description's local verification is thorough — but the mandatory gate is unverified by CI as of this SHA, presumably until the retarget onto `dev`.

**Nits**
- `apps/web/src/lib/mapbox.ts:17` — the `fitHoleUpward` doc comment says it's "Shared by the planner map (HoleStrategyMap), the round-detail map (useMapSetup), and the shot mini-map," but `ShotMiniMap.tsx` doesn't call `fitHoleUpward` — it reimplements the bearing + `fitBounds` composition inline (`ShotMiniMap.tsx:92-100`) using its own `bearingDegrees` import. Comment overstates actual reuse; worth fixing the comment or actually routing `ShotMiniMap` through the shared helper if its multi-point bounds needs can be accommodated.
- `apps/web/src/pages/rounds/components/MapView.tsx:417` — `const cat = s.category ?? 'approach'` defaults any shot with an unset `category` to "Approach" (and the tee-colored badge in `categoryColor` is unreachable via this fallback). If `category` is ever legitimately null on a tee or putt row, the shot list will mislabel it rather than showing "unknown."

---
Looked good: `fitHoleUpward` is a clean extraction with three real call sites (`HoleStrategyMap`, and twice in `useMapSetup`), the bearing math itself correctly stays pure in `@oga/core` per the project's package-boundary rule, no new `@oga/core` exports so no missing-test gap, and the diff stays web-only with no migrations, secrets, or mobile changes.

---

cner-smith:
Added `2403370` — restores the **Set pin** button for past-round logging. It was gated to render only when a hole had no pin coordinate; now that the OSM crawler fills pins on prod holes, `effectivePin` is never null so the button stopped showing. Re-gated on past/edit logging (`!isLiveEntry && activeHoleGeo != null`) — shown even when a crawled pin exists so it can be overridden, hidden during live-aim. Renamed the strip prop `needsPin`→`showPinButton` and relabeled "Place pin"→"Set pin". Verified with Playwright (visible on past-round Map tab, hidden on `?view=map`).

PR #645, opened 2026-07-06
fix(mobile): live-round field bugs — camera orientation/zoom, GPS remount, putting scroll (#639–643)

cner-smith:
Live-round field bugs surfaced during on-course play (2026-07-05/06). Four fixes, one commit each. #644 (swipe-to-dismiss for bottom sheets) is split into a follow-up PR — it's a larger, on-device-QA-gated sheet refactor.

## Fixes

- **#639 — Finish round never finalizes** (from last session): finalize live rounds on finish; stop stray/duplicate rounds.
- **#641 / #642 — Camera** (`useHoleCamera.ts`): apply a tee/ball→pin bearing as the camera heading in the initial, hole-change and PLACE_BALL reframe effects (not just the fleeting SET_AIM step that reset to north on commit), via the canonical `bearingDegrees`. Add short-game zoom tiers so a greenside chip frames tight instead of flatlining at the 80-yd zoom; fall the PIN snap back to ball/GPS on synthetic holes so tapping the pin tool frames the green.
- **#640 — GPS button / auto-place after remount** (`useHoleState.ts`): keep the GPS watcher subscribed while revisiting a played hole so the recenter button + Mark-ball CTA stay live (the #484 review posture is now enforced at the `setBall` gate, not by skipping the whole watcher). Auto-engage the component's first-mounted hole (round resume) so auto-place resumes without a "+ Add a shot" tap.
- **#643 — Putting sheet not scrollable on first open** (`PuttingSheet.tsx`): give the putt ScrollView an absolute window-derived `maxHeight` so its scroll range resolves in the first layout pass, independent of the modal slide transform and the parent percentage height.

## Verification

- `npm run typecheck` (mobile) green.
- Not yet device-QA'd — all four are field-reproduced; behaviour changes are camera/GPS/layout and want an on-device pass before merge.

Closes #639
Closes #640
Closes #641
Closes #642
Closes #643

### DIFF
--- a/apps/mobile/app/(app)/index.tsx
+++ b/apps/mobile/app/(app)/index.tsx
@@ -299,7 +299,10 @@ export default function Home() {
 
         <Entrance index={4}>
           <RecentRoundsList
-            rounds={rounds}
+            // Exclude the active round — it already shows as the Resume banner
+            // above. Without this it renders in both places, so deleting it
+            // clears both and reads as "deleted two rounds". (#639)
+            rounds={rounds.filter((r) => r.id !== activeRound?.id)}
             onRequestDelete={(id, name) => setPendingDelete({ id, name })}
           />
         </Entrance>
--- a/apps/mobile/app/(app)/round/new.tsx
+++ b/apps/mobile/app/(app)/round/new.tsx
@@ -23,6 +23,7 @@ import {
   createCourse,
   createHoles,
   createRound,
+  deleteRound,
   getCourseByExternalId,
   searchCourses,
   upsertCourseTees,
@@ -77,6 +78,11 @@ export default function NewRound() {
   } | null>(null)
   const [gps, setGps] = useState<GpsState>({ status: 'idle' })
   const searchAbort = useRef<AbortController | null>(null)
+  // Synchronous double-tap guard for round creation. `busy` state updates a
+  // tick later, so a fast double-tap on "Start round" could fire startWith
+  // twice and create two rounds — a ref blocks the second call immediately.
+  // (Mirrors persistShotInFlightRef in useShotActions.) (#639)
+  const startInFlightRef = useRef(false)
   const insets = useSafeAreaInsets()
 
   // Debounce 300ms.
@@ -178,8 +184,11 @@ export default function NewRound() {
     tee?: { courseTeeId: string | null; teeColor: string | null },
   ) {
     if (!user) return
+    if (startInFlightRef.current) return
+    startInFlightRef.current = true
     setBusy(true)
     setError(null)
+    let createdRoundId: string | null = null
     try {
       const today = todayLocalDate()
       const { data: round, error: roundError } = await createRound(supabase, {
@@ -198,6 +207,7 @@ export default function NewRound() {
           : {}),
       })
       if (roundError || !round) throw roundError ?? new Error('Round insert failed')
+      createdRoundId = round.id
 
       const { data: existingHoles, error: holesError } = await supabase
         .from('holes')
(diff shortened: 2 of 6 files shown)


cner-smith:
## Review — fix(mobile): live-round field bugs — camera orientation/zoom, GPS remount, putting scroll (#639–643)

**Concerns**
- `apps/mobile/components/round/hooks/useHoleCamera.ts:78-124` — The initial-center effect and the hole-change effect both compute `headingUpTheHole(center, roundPin ?? pin)` but their dependency arrays are `[styleLoaded, center.lat, center.lng]` / same — `roundPin`/`pin` aren't listed. `cameraInitialized.current` latches true after the first `setCamera` call in the initial effect, so if `roundPin`/`pin` are still `null` at the moment `center` first resolves (a very plausible race if pin geometry is a separate/slower fetch than the hole's tee/center), the heading falls back to north and is never recomputed for that hole — the "up the hole" orientation this PR adds would silently never engage until the next hole change. Worth confirming pin data is guaranteed to be present by the time `center` first settles, or adding the pin coords to these two effects' deps.
- `apps/mobile/components/round/hole/useShotActions.ts:480-491` — `finishHole()` now calls `void handleEndRound()` on the last hole, but `handleEndRound` (line 491) has no synchronous in-flight guard, only the async `ending` state (set via `setEnding(true)`, which — per this same PR's own reasoning at `round/new.tsx:81-84` and the pre-existing `persistShotInFlightRef` — updates a tick too late to block a fast double-tap). A quick double-tap of "Finish" on the 18th hole could fire `completeRound` twice concurrently, which is the exact race class this PR fixes elsewhere via `startInFlightRef`/`persistShotInFlightRef`. Might be worth the same ref guard here, or confirming the Finish control already disables on `ending`/`saving` (not visible in this diff).
- One-concern-per-PR (CONTRIBUTING.md: *"One concern per PR. UI restyle, bug fix, and new feature do not belong in the same PR"*) — this PR bundles four independently-rooted fixes (round finalize/double-tap-create in `round/new.tsx` + `index.tsx`, camera heading/zoom in `useHoleCamera.ts`, GPS watcher/auto-engage in `useHoleState.ts`, putting-sheet scroll in `PuttingSheet.tsx`) across 6 files. Appreciate that each is its own commit and that #644 was already split out as a separate PR — but by the letter of the rule these read as four separate PRs rather than one concern each touching different symptoms of the same root cause.

---
Looked good: `bearingDegrees` (already in `@oga/core`, already tested) is reused instead of re-deriving a new planar bearing calc — actually removes a duplicated, less-accurate `atan2` bearing from `useHoleCamera.ts` in favor of the shared great-circle helper; `deleteRound` rollback in `round/new.tsx` and the double-tap ref guard follow the same pattern already established by `persistShotInFlightRef`; base branch is correctly `dev`.


---

cner-smith:
Addressed the review:

- **C2 (finish double-tap) — fixed** in `9d86038`. Added a synchronous `endInFlightRef` guard on `handleEndRound`, mirroring the existing `persistShotInFlightRef`, cleared in `finally`. Confirmed there was no button-level `disabled` on Finish, so this was a real double-finalize window.

- **C1 (camera heading race) — fixed** in `29d2ee8`. Note: the suggested quick fix (adding pin coords to the two effects' deps) wouldn't actually work — both effects are latched (`cameraInitialized` / center-equality), so a pin-only change can't re-fire them. Instead added a `PLACE_BALL`-gated effect that re-frames once the pin resolves, keyed on pin coords so it fires once per hole (not on GPS ticks). Verified the common case (real hole) has no race: `center`/`storedPin` come from the same `currentHole` row; only synthetic holes with a separately-fetched round pin hit the window.

- **C3 (one-concern-per-PR) — intentional.** This is the deliberate "live-round field bugs" batch, one commit per issue (#639–643), with #644 already split out. Keeping as one PR by choice.

Still wants an on-device pass before merge (camera/GPS/layout behavior).

Issue #496, opened 2026-06-02
[ios] ShotLogger embeds PuttingSheet without GestureHandlerRootView — aim-drag dead


cner-smith:
## Summary
`components/round/ShotLogger.tsx:144` renders `<PuttingSheet>` (which contains a gesture-handler `<GestureDetector>` for the GreenDiagram aim handle) **inside its `<Modal>` wrapped only in a plain `<View>`** — no local `GestureHandlerRootView`. RN `<Modal>` renders to a separate native window, so the app-root root doesn't apply → **the aim-handle drag is dead** in this path.

Found in the device-free Android→iOS portability audit (#308).

## Mechanism (confirmed by code)
- `ShotLogger.tsx:131` `<Modal>` → `:137` plain `<View>` → `:144` `isOnGreen` branch renders `<PuttingSheet>`.
- `PuttingSheet` → `GreenDiagram` → `<GestureDetector gesture={pan}>`.
- The **sibling** PuttingSheet in `components/round/hole/HoleModals.tsx:136` IS wrapped in `<GestureHandlerRootView style={{ flex: 1 }}>`, with a comment stating exactly this. The ShotLogger copy is missing the wrap.

## Note — cross-platform, not iOS-only
The gesture-handler-in-Modal requirement applies on Android too, so this path is likely **already broken on Android** and simply rarely hit (the normal on-green flow goes through `roundState === 'PUTTING'` → the *wrapped* HoleModals sheet). Confirm reachability on device.

## Repro
Pending device verification: open the standard shot logger, set lie type to "green" so it swaps to the embedded PuttingSheet → try to drag the aim handle.

## Fix
Wrap ShotLogger's modal content (or at least the `isOnGreen` branch) in `<GestureHandlerRootView style={{ flex: 1 }}>`, mirroring `HoleModals.tsx:136`.

label: bug — real defect (cross-platform); repro/reachability to confirm on device.

cner-smith:
Shipped in #507 (dev→main, merged to production 2026-06-03). The release PR's comma-separated `Closes` list only auto-closed the first ref (#299), so closing manually.

Issue #293, opened 2026-05-13
Mobile: multiple ConfirmDialogs can render simultaneously — back button only closes the topmost


cner-smith:
**Severity:** low (recoverable stuck-state, unlikely state machine path)
**Audit source:** mobile-app-developer agent, 2026-05-13

## Mechanism

`apps/mobile/app/(app)/round/[id]/hole/_components/HoleModals.tsx` renders **seven ConfirmDialogs and two Modals**, each with its own `visible` flag. The parent state allows multiple to be true at once (e.g. `onGreenPromptOpen && confirmDelete`).

React Native stacks Modals; the last one rendered wins visually. On Android, `onRequestClose` (back button) fires on the topmost. If both `confirmDelete` and `aimPromptOpen` are open, hitting back only closes `aimPromptOpen` — the delete confirm stays mounted.

In practice the state machine transitions in `useShotActions` make this unlikely, but the dialog-opening handlers don't explicitly close other dialogs first.

## Proposed fix

Two options:

1. In each "open X" handler, explicitly close all other dialog state setters before opening the new one.
2. Refactor to a single `activeDialog: 'delete' | 'leave' | 'onGreen' | null` discriminated union state — only one can be open at a time by construction.

Option 2 is cleaner long-term; option 1 is one-line-per-handler and ships faster.

## Files

- `apps/mobile/app/(app)/round/[id]/hole/_components/HoleModals.tsx` (seven ConfirmDialogs)
- `apps/mobile/app/(app)/round/[id]/hole/_hooks/useShotActions.ts` (dialog-opening handlers)

---

## Validation 2026-05-22

**Verdict: already fixed — safe to close.**

Verified `apps/mobile/components/round/hole/HoleModals.tsx:46` on `dev`. The six separate `visible` flags are now collapsed into a single `activeDialog: ActiveDialog` discriminated union (defined in `apps/mobile/components/round/hole/types.ts`) — option 2 from the proposal. By construction only one dialog can be open at a time, so the Android back-button stacking concern is structurally impossible. Merged in PR #381 (commit `f7067c2`).

No remaining code work; closing recommended.

cner-smith:
## Additional iOS context (now that iOS is on the roadmap)

The body's Android framing (`onRequestClose` reaching only the topmost) is correct for Android but understates the iOS half. Worth capturing before this gets touched.

### iOS-specific behavior
[...]

PR #646, opened 2026-07-06
feat(mobile): swipe-down-to-dismiss for live-round bottom sheets (#644)

cner-smith:
Adds swipe-down-to-dismiss to the live-round bottom sheets (scorecard, putting, shot logger, past-hole shots), per issue #644.

## Approach
Scoped down from an earlier "shared BottomSheet that owns the Modal + backdrop + card" idea after a 3-agent review — that shape is a **hard blocker** (PuttingSheet is mounted twice, once *inside* ShotLogger's own Modal, so a Modal-owning wrapper double-nests Modals, reviving #293/#495). Instead:

- New `components/ui/useSwipeToDismiss.ts` (~40 lines): a `translateY` shared value + a `Gesture.Pan` bound to the **grabber region only** (never the card or an inner ScrollView, so it doesn't fight the scroll or GreenDiagram's own pan). Returns `{ pan, cardStyle }`.
- Adopted **additively** in each sheet: wrap the existing card in `Animated.View style={cardStyle}`, wrap the existing grabber in `GestureDetector`. Each sheet keeps its own Modal, backdrop, KeyboardAvoidingView, `maxHeight`, and Close button (backdrop-tap asymmetry preserved — ShotLogger/putting intentionally omit tap-to-close so a stray tap can't discard an in-progress log).
- Added `GestureHandlerRootView` inside the scorecard + past-hole-shots Modals (the putting/shot-logger ones already had one) — RN Modal is a separate native window on Android the app-root gesture handler can't reach (#496).
- **@gorhom/bottom-sheet rejected**: compatible with the pinned deps, but more invasive (swaps ScrollView/TextInput internals + a portal provider) and rides reanimated's worklet ABI right as SDK 54 / reanimated 4 (which pulls react-native-worklets) is on the roadmap.

## Protecting #643
The putting sheet's card is driven via an animated `KeyboardAvoidingView`; the `translateY` is a post-layout transform, so it can't affect the ScrollView's absolute `windowHeight*0.72` scroll range — the #643 fix is untouched. Tree structure and `maxHeight:'90%'` unchanged.

## Verification
- `npm run typecheck` (mobile) green.
- **Not yet device-QA'd** — swipe gesture, the grabber drag-target size, and the Android separate-window behavior want an on-device pass before merge.

Closes #644

### DIFF
--- a/apps/mobile/components/round/PastHoleShotsSheet.tsx
+++ b/apps/mobile/components/round/PastHoleShotsSheet.tsx
@@ -1,6 +1,12 @@
 import { useEffect, useState } from 'react'
 import { Modal, Pressable, ScrollView, Text, TextInput, View } from 'react-native'
+import {
+  GestureDetector,
+  GestureHandlerRootView,
+} from 'react-native-gesture-handler'
+import Animated from 'react-native-reanimated'
 import { PressableTouch } from '../ui/PressableTouch'
+import { useSwipeToDismiss } from '../ui/useSwipeToDismiss'
 import {
   DEFAULT_BAG,
   LIE_SLOPES_FORWARD,
@@ -152,6 +158,8 @@ export function PastHoleShotsSheet({
   // One <Modal> with discriminated content (list vs editor) — NOT two
   // sibling Modals. iOS allows one presented modal per presenter, so the
   // old stacked-Modal edit flow silently failed to present (#293/#495).
+  const { pan, cardStyle } = useSwipeToDismiss(onClose)
+
   return (
     <Modal
       visible={visible}
@@ -159,6 +167,9 @@ export function PastHoleShotsSheet({
       animationType="slide"
       onRequestClose={editingShot ? () => setEditingShot(null) : onClose}
     >
+      {/* GHRootView required for the swipe-to-dismiss pan: RN Modal is a
+          separate native window on Android the app-root can't reach (#496). */}
+      <GestureHandlerRootView style={{ flex: 1 }}>
       <View style={{ flex: 1, backgroundColor: 'rgba(0,0,0,0.5)' }}>
         <Pressable
           style={{ flex: 1 }}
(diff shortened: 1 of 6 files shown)


cner-smith:
## Review — feat(mobile): swipe-down-to-dismiss for live-round bottom sheets (#644)

**Blockers**
- `apps/mobile/components/ui/useSwipeToDismiss.ts:24-27,42,48` — `translateY` is never reset to 0 on the dismiss branch (`onEnd`, when `translationY > DISMISS_DISTANCE || velocityY > DISMISS_VELOCITY`), only on the "spring back" branch. The comment justifies this by assuming the Modal "unmounts its children" on close, disposing the shared value. That assumption doesn't hold for any of the four call sites:
  - `ScorecardModal` is rendered unconditionally inside `<Modal visible={scorecardOpen}>` in `apps/mobile/components/round/hole/HoleModals.tsx:210-232` with no `key` — it's the same mounted instance across close/reopen, `visible` only toggles the native window.
  - `PuttingSheet` (`HoleModals.tsx:128-146`) and `ShotLogger` (same file, `key={shotEntryKey}`) only remount when `shotEntryKey` changes. `closeLogger`/`closePuttingSheet` in `apps/mobile/components/round/hole/useShotActions.ts:453-461` don't touch `shotEntrySeq`, so a plain close-then-reopen (no save) reuses the same instance.
  - `PastHoleShotsSheet` is rendered with no `key` at `apps/mobile/app/(app)/round/[id]/index.tsx:985`, same as `ScorecardModal`.

  Net effect: swipe a sheet closed past the threshold, then reopen it (a very ordinary flow — e.g. open the shot logger, swipe it away, tap to open it again) — the card renders already translated down by whatever distance/velocity carried the last dismiss, since `translateY.value` was left non-zero. It only self-corrects once the user drags again (the "else" branch springs it back to 0). Worth fixing before the on-device pass called out in the PR description — this is exactly the kind of thing that pass should catch, but it's deterministic enough to fix now: reset `translateY.value = 0` (no need for `withSpring`, since the Modal's own slide-out is covering the exit) before/in the `runOnJS(onClose)()` branch.

**Nits**
- `apps/mobile/components/ui/useSwipeToDismiss.ts:35` — the pan gesture is bound only to the grabber pill's wrapping `View` (`alignItems: 'center', paddingBottom: 14`, pill itself is 32×4). That's a small target for a drag gesture; might be worth a touch/hitSlop expansion on the grabber row, though not blocking.

---
Looked good: scope held to one concern (swipe-to-dismiss only, no unrelated restyle/logic creep), `useSwipeToDismiss` correctly extracted to `apps/mobile/components/ui/` given 5+ call sites (3-caller rule), no `@oga/core` changes so no test-coverage gap, no new/bumped mobile deps, PR correctly targets `dev`.

---

cner-smith:
Triaged — the blocker's premise doesn't hold, so folding the residual into the on-device QA pass rather than changing code now.

**On the blocker:** RN 0.79 `<Modal>` renders `null` when not visible (`react-native/Libraries/Modal/Modal.js:255-266`), so children **unmount** on close — Android immediately, iOS after the dismiss animation (`isRendered`). On a normal close→reopen the hook remounts and `useSharedValue(0)` starts fresh at 0, so the "swipe closed, reopen, card still translated" flow doesn't reproduce (the instance was torn down between). The `key`/`shotEntryKey` distinction is moot once the Modal returns null.

**Real residual (narrow, iOS-only):** reopening *during* the ~300ms exit animation (before `isRendered` flips false) keeps the instance alive at the last drag offset. Frame-perfect edge. A defensive `translateY.value = 0` on the dismiss branch would cover it, but whether that snap introduces a visible up-jump as the native modal slides out is itself a device question — so verifying on-device either way.

**On-device QA checklist for this PR:**
- [ ] iOS: swipe a sheet closed, then immediately reopen *during* the dismiss animation — card should render at rest, not translated.
- [ ] Grabber drag-target feel — the 32×4 pill's wrapping row may want a taller hit area / `hitSlop`.
- [ ] All 5 surfaces: swipe-dismiss works and doesn't fight inner scroll / the GreenDiagram pan (putting sheet).
- [ ] Android: gesture reaches inside the scorecard + past-hole-shots modals (the newly added GestureHandlerRootView).

PR #680, opened 2026-07-06
fix(core): stats correctness — SG inflation, GIR inference, mobile handicap baseline (#668, #669, #673)

cner-smith:
Stats-correctness cluster from the 2026-07-06 internal audit — three fixes, one commit each, all numerically verified.

## Fixes

- **#668 — `approachByDistance` + `girPct`** (`packages/core/src/stats.ts`): a next shot with no derivable expected strokes was treated as a hole-out (`endExpected = 0`), awarding up to **+2.95 SG** per data-less shot and making the "SG by distance" chart disagree with `calculateRoundSG` on identical data — now mirrors the round engine's skip. `girPct` counted gir-null holes (scorecard-only rounds) in the denominator as misses — now denominates on holes where GIR is actually known. **First test coverage for both exports** (they were on the audit's untested list — exactly where the bugs lived).
- **#669 — dead `'holed'` GIR branch** (`packages/core/src/holeInference.ts`): `shot_result='holed'` is unreachable (0021 CHECK constraint; no writer), so aces/holed approaches never counted as GIR — and the dead branch's arithmetic would have wrongly counted chip-ins for par (green-start shots need ≤ par−1; off-green hole-outs reach the green ON the holing stroke, so ≤ par−2). Replaced with an explicit `holedOut` param: web's review flow passes `true` (holed-out by construction — final marker ends at the pin, `score = rows.length`); mobile omits it (no reliable off-green holing signal — documented at the call site). A single logged shot without `holedOut` now infers fairway `null` (partial log) instead of `false`. Old tests fed the unreachable input and masked the bug — rewritten with reachable inputs, including the chip-in-for-par case.
- **#673 — mobile stats DEFAULT_HANDICAP** (`apps/mobile/app/(app)/stats.tsx`): SG baselines were computed against the default bracket for every player; now fetches `profile.handicap_index` (fallback DEFAULT_HANDICAP), matching web's `useDetailedStats`.

## Behavior notes
- Mobile GIR for off-green hole-outs stays conservatively false (no signal exists on mobile) — same as before, but now documented; web logging of the same round infers it correctly.
- `hole_scores(*)` is already selected by `getRoundsWithDetails`, so the girPct change needs no query change.

## Verification
- `pnpm test`: **568 passed** (was ~547; new cases for approachByDistance, ballStrikingStats, and the rewritten holeInference suite).
- `pnpm typecheck` (3/3) + mobile `npm run typecheck` green.

Closes #668
Closes #669
Closes #673

### DIFF
--- a/apps/mobile/app/(app)/stats.tsx
+++ b/apps/mobile/app/(app)/stats.tsx
@@ -16,7 +16,7 @@ import {
   type DetailedStats,
   type SGAverages,
 } from '@oga/core'
-import { getRoundsWithDetails } from '@oga/supabase'
+import { getProfile, getRoundsWithDetails } from '@oga/supabase'
 import { supabase } from '../../lib/supabase'
 import { useAuth } from '../../hooks/useAuth'
 import { useUnits } from '../../hooks/useUnits'
@@ -58,6 +58,26 @@ export default function Stats() {
   const [n, setN] = useState<number>(10)
   const [rounds, setRounds] = useState<DetailedRound[]>([])
   const [loading, setLoading] = useState(true)
+  // Player's handicap for SG baselines. Web (useDetailedStats) already uses
+  // the profile value; hardcoding DEFAULT_HANDICAP here benchmarked every
+  // player against the default bracket and made mobile disagree with web on
+  // the same rounds (#673).
+  const [handicap, setHandicap] = useState<number | null>(null)
+
+  useEffect(() => {
+    if (!user) return
+    let active = true
+    getProfile(supabase, user.id).then(({ data }) => {
+      if (!active) return
+      setHandicap(
+        (data as { handicap_index?: number | null } | null)?.handicap_index ??
+          null,
+      )
+    })
+    return () => {
+      active = false
+    }
+  }, [user?.id])
   const { width: screenWidth } = useWindowDimensions()
   const { unit, toDisplay } = useUnits()
 
@@ -132,8 +152,11 @@ export default function Stats() {
   )
 
   const stats: DetailedStats | null = useMemo(
-    () => (rounds.length > 0 ? computeDetailedStats(rounds, DEFAULT_HANDICAP) : null),
-    [rounds],
+    () =>
+      rounds.length > 0
+        ? computeDetailedStats(rounds, handicap ?? DEFAULT_HANDICAP)
+        : null,
+    [rounds, handicap],
   )
 
   return (
--- a/apps/mobile/app/(app)/stats.tsx
+++ b/apps/mobile/app/(app)/stats.tsx
@@ -67,8 +67,12 @@ export default function Stats() {
   useEffect(() => {
     if (!user) return
     let active = true
-    getProfile(supabase, user.id).then(({ data }) => {
+    getProfile(supabase, user.id).then(({ data, error }) => {
       if (!active) return
+      if (error) {
+        // eslint-disable-next-line no-console
+        console.error('[stats/getProfile]', error.message)
+      }
       setHandicap(
         (data as { handicap_index?: number | null } | null)?.handicap_index ??
           null,
--- a/apps/mobile/lib/completeRound.ts
+++ b/apps/mobile/lib/completeRound.ts
@@ -89,11 +89,14 @@ export async function completeRound({
     const hole = holesById.get(hs.hole_id)
     if (!hole) continue
     const holeShots = shots.filter((s) => s.hole_score_id === hs.id)
+    // No holedOut arg: mobile has no reliable signal that the last logged
+    // shot finished in the cup (no 'holed' shot_result exists), so off-green
+    // hole-outs conservatively don't count GIR here (#669). Web's review
+    // flow, which IS holed-out by construction, passes true.
     const inferred = inferHoleStats(
       holeShots.map((s) => ({
         shot_number: s.shot_number,
         lie_type: s.lie_type,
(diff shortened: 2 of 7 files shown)


cner-smith:
## Review — fix(core): stats correctness — SG inflation, GIR inference, mobile handicap baseline (#668, #669, #673)

**Concerns**
- Scope: this PR bundles three independently-closable issues (#668, #669, #673) touching unrelated code paths — `approachByDistance`/`girPct` inflation, the dead `'holed'` GIR branch (web + mobile + core), and mobile's hardcoded SG baseline handicap. CONTRIBUTING.md's "Opening a PR" section is explicit: *"One concern per PR. UI restyle, bug fix, and new feature do not belong in the same PR."* These are all bug fixes (not a restyle+feature mix), and each has its own commit, but they're still three distinct root causes bundled under one "cluster" framing. Not blocking — I think this is understandable for an audit sweep — but worth naming since it's the literal rule as written.
- `apps/mobile/lib/completeRound.ts:88-102` — the comment says mobile "has no reliable signal that the last logged shot finished in the cup," so `holedOut` is always omitted. But `hs` (the full `hole_scores` row) is already in scope here with `hs.score` available. When `holeShots.length === hs.score`, the shot log is complete for that hole and the last shot must have holed out — that reliable signal exists and isn't used. As written, mobile will permanently under-report GIR/fairway on any ace, holed approach, or chip-in, even on holes with a fully-logged shot sequence. Might be worth a follow-up (`holedOut: holeShots.length === hs.score`) rather than accepting the conservative gap as permanent.

**Nits**
- `packages/core/src/__tests__/holeInference.test.ts:125-128` — test title says "chip-in for par on par 4," but the fixture (`[tee(1), lie(2,'rough'), lie(3,'rough')]`, 3 shots, holed on stroke 3) is 3 strokes on a par 4 — a birdie, not a par. The assertion itself (`gir: false`, stroke 3 > par−2) is correct; just the label is off, which could mislead someone extending this suite later.
- `apps/mobile/app/(app)/stats.tsx:70-76` — `getProfile` result is cast to an ad-hoc inline type (`{ handicap_index?: number | null } | null`) rather than reusing whatever type `getProfile` already infers — web's `useDetailedStats.ts` reads `profile.data?.handicap_index` off the same query with no cast, so this looks unnecessary. Also, unlike the other two mobile call sites for `getProfile` (`index.tsx`, `profile.tsx`), the `error` half of the response is dropped silently here — a failed fetch just falls back to `DEFAULT_HANDICAP` with nothing logged, which is safe but harder to debug than the sibling call sites.

---
Looked good: `approachByDistance`'s fixed skip genuinely mirrors `calculateRoundSG`'s existing holed-out assumption (verified against `sg-calculator.ts`/`sg.ts` — same "no next shot in the per-hole array ⇒ holed" model, not a new risk), `girPct` denominator fix is correct and has direct test coverage, `holeInference.ts`'s `holedOut` param and its GIR/fairway branches check out against the actual golf rules (green-start vs. hole-from-off-green needing one stroke stricter), new `@oga/core` exports (`approachByDistance`, `ballStrikingStats`) get their first tests per CONTRIBUTING's rule, no `any`, no migrations/secrets/mobile-dep changes, PR correctly targets `dev`.

---

cner-smith:
## Review — fix(core): stats correctness — SG inflation, GIR inference, mobile handicap baseline (#668, #669, #673)

**Concerns**
- `apps/mobile/app/(app)/stats.tsx:70-76` — the new profile-fetch effect destructures only `{ data }` and drops `error` entirely, so a failed `getProfile` call silently falls back to `DEFAULT_HANDICAP` with no signal to the user or logs. The very next `useEffect` in the same file (`getRoundsWithDetails`, lines ~85-95) destructures `{ data, error }` and explicitly `console.error`s on failure — this one is inconsistent with that adjacent pattern. Falling back to `DEFAULT_HANDICAP` is a reasonable degradation, but the silent swallow means a persistent profile-fetch failure (e.g., RLS/auth issue) would look identical to "user has no handicap set," making it hard to diagnose. Suggest at least logging the error like the sibling effect does.
- This PR folds three separately-numbered fixes (#668, #669, #673) into one PR ("three fixes, one commit each"). Per CONTRIBUTING's one-concern-per-PR rule this is worth a mention, though all three are same-kind (bug fix) changes from one audit pass touching overlapping stats/inference code, so — as with #681 — this reads more like reasonable batching than scope creep. Flagging for maintainer discretion, not blocking.

---
Looked good: `inferHoleStats`'s new `holedOut` param is a clean fix for the dead/wrong `'holed'` branch — traced both `inferFairway` and `inferGir` against all the new test cases (ace, holed-approach-for-eagle, chip-in-for-par) and the off-by-one reasoning (par-2 for off-green hole-outs vs. par-1 for green-start) holds up; confirmed via `git grep` that both real call sites (`useRoundActions.ts` web, `completeRound.ts` mobile) were updated and no third caller was missed (the live-round scorecard call in `apps/mobile/app/(app)/round/[id]/index.tsx` intentionally keeps the conservative no-`holedOut` default, which matches the "mobile has no reliable holing signal" documented tradeoff); `approachByDistance`'s skip-instead-of-holed-out fix mirrors `calculateRoundSG`'s existing behavior and is covered by a new test for the exact regression case; `girPct` denominator fix is correct and tested for both the mixed and all-null cases; new tests land in the same package for all new/previously-untested `@oga/core` exports touched; no mobile dependency changes; PR targets `dev`.

---

PR #681, opened 2026-07-06
fix(mobile): harden shot sync queue + completeRound sync race (#651, #652)

cner-smith:
Two audit-filed defect sets in the mobile pending-shot sync pipeline, one commit each.

## #652 — Sync-queue hardening (`lib/sync.ts`, `lib/db.ts`)

| Defect | Fix |
|---|---|
| **Poison pill** — chunks insert atomically; a server-permanently-rejected row (23505/23503) failed its whole chunk, and all future sync, forever | `isPermanentError()` classifies Postgres class 22/23 as deterministic; on a permanent chunk failure, `syncRowsIndividually()` inserts one row at a time, quarantines the poison row(s) via `markShotBroken`, and lets the rest through. Network/5xx keep the transient single-retry. |
| **Chunk index misalignment** — `data[j]` (payloads order) mapped against `chunk[j]` (chunk order) once a corrupt row was skipped | Parse loop builds parallel `validRows`/`payloads` arrays; success reconciles by the client-generated id against a `Set` of returned ids — no ordering trust at all. |
| **No idempotency** — retry after a committed-but-lost-response insert duplicated shots | `insertPendingShot` bakes a client-generated uuid into the payload; `insertChunk` upserts `onConflict: 'id'`, so a re-send lands on `DO UPDATE` (local payload is the freshest copy while pending — safe refresh). Legacy queued rows get an id backfilled and persisted on first sync pass (`updatePendingShotPayload`). |
| **`pending_shots` never pruned + dead `deletePendingShot`** | Synced rows purged at db open; dead function removed (replaced by `updatePendingShotPayload`, which the backfill needs). |

## #651 — completeRound sync-lock no-op race (`lib/sync.ts`, `lib/completeRound.ts`, `useShotActions.ts`)

The boolean lock returned an instant `{synced: 0, failed: 0}` no-op to concurrent callers; `completeRound` discarded it and stamped `completed_at` + totals/SG while the final putt's save-triggered background sync was still inserting. Fix: the lock becomes a module-level promise of the active run — concurrent callers **join** the in-flight run and get its real result. `handleEndRound` drains the queue before calling `completeRound`, re-runs once if rows landed after the joined run's snapshot, then checks `pendingCount()`; if rows still can't reach the server, an Alert forces a choice: **Keep round open** (round stays live, retry later) or **Finish anyway** (knowingly finalize over what synced).

## Design decisions

- **uuid source:** `uuid.v4()` from `expo-modules-core` — already an exact-pinned transitive dep (2.5.0), typed export, backed by the native Expo runtime generator (`globalThis.expo.uuidv4`). Zero new dependencies; `expo-crypto` is NOT installed so it was not an option without touching `package.json`.
- **Idempotency shape:** upsert-on-PK rather than "treat 23505-on-PK as already-synced" — the upsert path never surfaces the ambiguous 23505 at all, so any 23505 that remains is the natural `unique(hole_score_id, shot_number)` key = genuine poison. Shots RLS is one `FOR ALL` own-rows policy, so the `ON CONFLICT DO UPDATE` arm passes.
- **Permanent-error classification:** conservative — only Postgres classes 22/23. Blanket 4xx would quarantine the entire queue on an expired JWT (PGRST301) or a signed-out RLS denial; those stay transient/pending.
- **Join vs retry-loop:** join-in-flight (module-level promise). The promise *is* the lock, so it can't go stale like the boolean; the 30s TTL survives as "don't join a zombie run, start fresh" to keep the old hung-fetch recovery.
- **Prune strategy:** purge `status='synced'` at db open rather than delete-on-sync — `setPendingShotEnd` patches a just-synced shot's end coords via `remote_id` within a session (`markBallHere` remote-patch path), so deleting at sync time would break that. The ref driving that path (`lastSavedShotLocalIdRef`) is in-memory and doesn't survive a restart, so startup purge loses nothing. `broken` rows are kept as diagnostic evidence.
- **Warn-UX:** pre-stamp Alert in `handleEndRound` (not a return value from `completeRound`) so the warning genuinely precedes `completed_at`. The check is device-global `pendingCount()` rather than round-scoped — any pending row is un-synced shot data, so the superset warning is still truthful and avoids a new round-scoped SQLite query. "Keep round open" simply returns; the end dialog closes and the player can retry End Round later.

## Manual verification reasoning (no mobile unit-test infra)

Walking each defect through the final code:

1. **Poison row:** queue = [A(poison-23505-natural-key), B, C]. `insertChunk` upserts all three → 23505 → `isPermanentError('23505')` true → no transient retry burned → `syncRowsIndividually`: B and C each upsert alone and succeed → `markShotSynced(B)`, `markShotSynced(C)`; A alone → 23505 → `markShotBroken(A)`. Next pass: `listPendingShots` filters `status='pending'` → queue empty. Sync unblocked, no reinstall.
2. **Corrupt row mid-chunk:** queue = [X(corrupt JSON), Y, Z]. Parse loop quarantines X (`markShotBroken`, `failed+=1`) and never pushes it to `validRows`/`payloads`. Success loop iterates `validRows` = [Y, Z] and checks each `payload.id` against the returned-id `Set` — Y and Z get their own ids as `remote_id`; nothing is off-by-one, X's quarantine is never overwritten (the old bug marked X 'synced' with Y's remote id).
3. **Lost response:** insert of chunk with ids [i1, i2] commits server-side, response lost → supabase-js surfaces a transport error with empty `code` → not permanent → one 2s retry → upsert `onConflict:'id'` re-sends i1/i2 → `DO UPDATE` no-op-refresh, returns both rows → reconcile marks both synced with their client ids. Zero duplicates. Cross-run case identical because the id is persisted in the SQLite payload (at creation for new rows; backfilled + written back for legacy rows).
4. **Prune:** `getDb()` runs `DELETE FROM pending_shots WHERE status='synced'` once per session before any read. Rows synced this session survive until next launch, keeping the `setPendingShotEnd` → remote end-coords patch working; prior sessions' rows are gone. `deletePendingShot` no longer exists; `grep` confirms zero callers existed.
5. **#651 race:** final putt saves → `persistShot` fires background sync (holds `activeRun`). Player taps End Round → `handleEndRound` → `syncPendingShots()` returns *that same promise* (run snapshotted after the putt's `insertPendingShot`, which persistShot awaits before firing) → genuinely waits for the putt's insert. Belt-and-suspenders: if a joined run predated some rows, `pendingCount() > 0` triggers one fresh pass. Only then does `completeRound` read `getShotsForRound`.
6. **#651 failed-sync silent finalize:** airplane-mode sync leaves rows pending → both passes fail → `unsynced > 0` → Alert. "Keep round open" returns before `completeRound`; `finally` resets `ending`/`endInFlightRef` and closes the end dialog — round still resumable. "Finish anyway" proceeds, which is the documented acknowledged fallback (completeRound's internal drain keeps its `.catch(() => undefined)` for exactly this path).

`cd apps/mobile && npm run typecheck` passes after each commit.

**On-device QA still wanted:** poison-row quarantine + duplicate-prevention need a real flaky-network round (or airplane-mode toggling) to observe end-to-end; the Alert copy/flow should be eyeballed on Android + iOS.

Closes #651
Closes #652

### DIFF
--- a/apps/mobile/components/round/hole/useShotActions.ts
+++ b/apps/mobile/components/round/hole/useShotActions.ts
@@ -11,6 +11,7 @@ import { deleteRound, getProfile } from '@oga/supabase'
 import { supabase } from '../../../lib/supabase'
 import {
   insertPendingShot,
+  pendingCount,
   setPendingShotEnd,
   type ShotPayload,
 } from '../../../lib/db'
@@ -498,6 +499,46 @@ export function useShotActions(input: UseShotActionsInput): UseShotActionsResult
     endInFlightRef.current = true
     setEnding(true)
     try {
+      // Drain the queue before finalizing (#651). syncPendingShots joins
+      // an in-flight run instead of no-oping, but a joined run may have
+      // snapshotted the queue before the final putt's row landed — if
+      // anything is still pending after the first pass, run once more now
+      // that the previous run has settled.
+      await syncPendingShots().catch(() => undefined)
+      if ((await pendingCount()) > 0) {
+        await syncPendingShots().catch(() => undefined)
+      }
+      const unsynced = await pendingCount()
+      if (unsynced > 0) {
+        // Shots that never reached the server would silently vanish from
+        // totals/SG (completeRound reads the server's shot set). Make the
+        // player choose: keep the round open and retry with a better
+        // connection, or knowingly finalize over what synced. cancelable:
+        // false so the Android back button can't dismiss without
+        // resolving.
+        const finishAnyway = await new Promise<boolean>((resolve) => {
+          Alert.alert(
+            'Some shots have not synced',
+            `${unsynced} shot${unsynced === 1 ? ' has' : 's have'} not reached the server. ` +
+              'Finishing now will compute totals and strokes gained without ' +
+              `${unsynced === 1 ? 'it' : 'them'}. You can keep the round open and finish later with a better connection.`,
+            [
+              {
+                text: 'Keep round open',
+                style: 'cancel',
+                onPress: () => resolve(false),
+              },
+              {
+                text: 'Finish anyway',
+                style: 'destructive',
+                onPress: () => resolve(true),
+              },
+            ],
+            { cancelable: false },
+          )
+        })
+        if (!finishAnyway) return
+      }
       const { data: profile } = await getProfile(supabase, user.id)
       const handicap =
         (profile as { handicap_index?: number | null } | null)?.handicap_index ??
--- a/apps/mobile/lib/completeRound.ts
+++ b/apps/mobile/lib/completeRound.ts
@@ -42,6 +42,13 @@ export async function completeRound({
   userId,
   handicap,
 }: CompleteArgs): Promise<void> {
+  // Drain the pending queue before reading shots. syncPendingShots joins
+  // an in-flight background run (#651), so this genuinely waits for the
+  // final hole's shots instead of no-oping while a save-triggered sync
+  // holds the lock. The caller (handleEndRound) has already drained,
+  // re-checked pendingCount, and made the player acknowledge any rows
+  // that still can't reach the server — so a failure here proceeds by
+  // design: finalizing over what synced is the acknowledged fallback.
   await syncPendingShots().catch(() => undefined)
 
   const [holesRes, holeScoresRes, shotsRes, teesRes, roundRes] = await Promise.all([
--- a/apps/mobile/lib/db.ts
+++ b/apps/mobile/lib/db.ts
@@ -1,4 +1,5 @@
 import { AppState, type AppStateStatus } from 'react-native'
+import { uuid } from 'expo-modules-core'
 import * as SQLite from 'expo-sqlite'
 import type { Database } from '@oga/supabase'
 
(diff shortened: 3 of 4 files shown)


cner-smith:
## Review — fix(mobile): harden shot sync queue + completeRound sync race (#651, #652)

**Concerns**
- This PR bundles two separately-numbered defects (#651 `completeRound` sync-lock race, #652 sync-queue hardening) into one PR, each with its own commit. CONTRIBUTING.md's workflow section is explicit: "**One concern per PR.** UI restyle, bug fix, and new feature do not belong in the same PR." The examples given are about mixing change *types*, and both changes here are bug fixes touching the same `sync.ts`/`db.ts` pipeline, so this may be a reasonable judgment call rather than a violation in spirit — but per the letter of the rule these are two independently-described, independently-fixable defects that could have shipped as two PRs. Not blocking, flagging for maintainer discretion.

**Nits**
- `apps/mobile/lib/sync.ts:1-` — new `isPermanentError`/`syncRowsIndividually`/`insertChunk` split adds several single-purpose helpers to one file; each has 1-2 callers, at the edge of (but I think justified by) the "3-caller rule"'s "genuinely unreadable inline" exception given how much branching `runSync` already has.

---
Looked good: chunk reconciliation now keys off the client-generated id (`returnedIds` Set) rather than trusting response ordering, closing the exact misalignment bug described in #652; `isPermanentError` scoped conservatively to Postgres classes 22/23 so an expired-JWT or RLS-denial 4xx can't quarantine the whole queue; upsert-on-PK idempotency reasoning for the lost-response case checks out; the join-in-flight-promise lock is a real fix for the boolean no-op race in #651 and I don't see a way it can leave a phantom lock (the stale-run branch correctly avoids clobbering a newer run in its `finally`); `deletePendingShot` removal confirmed dead via `git grep` — no callers left in the tree; no new mobile dependency, no `@oga/core` changes so no test-coverage gap, PR targets `dev`.

---

PR #682, opened 2026-07-06
fix(mobile): live-round data-loss traps — 9-hole phantom resume + load-error delete (#650, #653)

cner-smith:
Two mobile data-loss traps from the 2026-07-06 audit — both let an ordinary flow destroy a fully-logged round. One commit each.

## #650 — 9-hole resume → phantom hole 10 → delete trap
`useActiveRound` computed the resume hole as `min(18, maxScoredHole + 1)`. On a fully-played 9-hole round that's hole 10, which doesn't exist → the hole screen's error branch → a one-tap "discard the round" (which #653 also fixes). Now: fetch **all** of the round's `hole_scores` (batch-created at round start, so their numbers give the true hole count), infer the count via `inferHoleCount`, and clamp the resume hole to it. A 9-hole round resumes at 9, never 10.

## #653 — load-error screen deleted the round
The error branch (`data.error || !currentHole`) served **one** delete dialog — "Nothing's been logged yet, so the round will be discarded" → `deleteRound` — for both a transient **load failure** (a resumable, holes-deep round under bad signal) and a genuinely missing hole. A network blip on resume could destroy the whole round behind copy asserting it was empty.

`handleExitFromError` no longer deletes — it navigates home. Rationale: a load error is resumable, and synthetic no-layout courses became playable when the no-layout banner was removed (#614), so the "unplayable → discard" case that justified the delete is gone. The round stays resumable and is still deletable from the home round list. Copy corrected on both the screen and the (now non-destructive) confirm; button recolored from alarm-red to neutral, "Exit round" → "Leave to home".

## Notes
- The `[id]/index.tsx` URL clamp still bounds a raw `?hole=` param to 1–18 — but with #650 the banner never emits an out-of-range hole, `navigateHole` already guards `> holeCount`, and #653 makes the error screen safe if one ever slips through. Left as-is (it can't cheaply know the hole count pre-mount).
- Merge note: `useShotActions.ts` is also touched by the in-flight sync PR #681 (a different function, `handleEndRound`) — no line overlap, but sequence the merges.

## Verification
Mobile `npm run typecheck` green. Not device-QA'd — wants a 9-hole resume repro + a bad-signal resume repro on device.

Closes #650
Closes #653

### DIFF
--- a/apps/mobile/components/round/LiveRoundSession.tsx
+++ b/apps/mobile/components/round/LiveRoundSession.tsx
@@ -312,8 +312,8 @@ export default function LiveRoundSession({
       ? 'Something went wrong loading this round.'
       : `Hole ${holeNumber} isn't set up for this round yet.`
     const subline = data.error
-      ? 'Check your connection and try again, or exit to clear the round.'
-      : 'This usually means the course was created without per-hole layout data. Exit to discard the round and start fresh.'
+      ? 'Check your connection and try again, or leave and resume this round later.'
+      : 'Try again, or leave and pick this round back up from the home screen.'
     return (
       <View
         style={{
@@ -380,10 +380,10 @@ export default function LiveRoundSession({
         </Pressable>
         <Pressable
           accessibilityRole="button"
-          accessibilityLabel="Exit round and discard"
+          accessibilityLabel="Leave and go to the home screen"
           onPress={() => setActiveDialog('exit')}
           style={{
-            backgroundColor: '#A33A2A',
+            backgroundColor: '#5C6356',
             borderRadius: 2,
             paddingVertical: 14,
             paddingHorizontal: 24,
@@ -400,17 +400,15 @@ export default function LiveRoundSession({
               },
             ]}
           >
-            Exit round
+            Leave to home
           </Text>
         </Pressable>
         <ConfirmDialog
           visible={activeDialog === 'exit'}
           title="Leave this round?"
-          message="Nothing's been logged yet, so the round will be discarded."
-          confirmLabel="Leave round"
+          message="Your round is saved — you can resume it from the home screen."
+          confirmLabel="Leave"
           cancelLabel="Stay"
-          destructive
-          busy={actions.deleting}
           onConfirm={actions.handleExitFromError}
           onCancel={() => setActiveDialog(null)}
         />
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — fix(mobile): live-round data-loss traps — 9-hole phantom resume + load-error delete (#650, #653)

**Concerns**
- Scope: this bundles two separately-numbered, separately-closable defects (#650 resume-hole clamping in `useActiveRound.ts`, #653 the destructive error-screen exit in `useShotActions.ts`/`LiveRoundSession.tsx`) into one PR. CONTRIBUTING.md's rule is explicit: "**One concern per PR.** UI restyle, bug fix, and new feature do not belong in the same PR." Both are bug fixes from the same audit and the PR keeps them as one commit each, so — consistent with how #680/#681 were flagged — this reads as reasonable batching from one audit pass rather than scope creep, but it's worth naming since it's the letter of the rule.
- `apps/mobile/components/round/LiveRoundSession.tsx:315-316` — the new subline for the genuinely-missing-hole case ("Try again, or leave and pick this round back up from the home screen.") and the confirm dialog's "Your round is saved — you can resume it from the home screen." now apply to *both* branches of `data.error || !data.round || !data.currentHole || !data.currentHoleScore`. The PR body's rationale (synthetic no-layout courses became playable via #614, so there's no "genuinely unplayable" round left to justify a discard) is credible but I can't verify #614's coverage from this diff — if there's any remaining path where `currentHole`/`currentHoleScore` come back missing for a reason other than a transient load blip (e.g. a data integrity gap), "your round is saved, resume it" would be told to a round that in fact can't resume. Worth a mental double-check against #614's actual scope before merge; not blocking since the round stays deletable from the home list regardless.

**Nits**
- `apps/mobile/hooks/useActiveRound.ts:58-83` — dropping the `.gt('score', 0)` server-side filter in favor of fetching all `hole_scores` rows and re-deriving the scored-only max client-side is correct (verified the `(row.score ?? 0) > 0` reduce condition reproduces the old filter) and is the right call since the unscored rows are what let `inferHoleCount` see the true hole count. Just flagging that this pulls slightly more data per focus-refetch than before — inconsequential at 9–18 rows, not worth changing.

---
Looked good: `inferHoleCount` is an existing, already-tested `@oga/core` export (`packages/core/src/round.ts:77`, tests in `packages/core/src/__tests__/round.test.ts`) reused here, not a new untested one; the pattern matches its other mobile call sites (`useHoleData.ts`, `ShareableScorecardCard.tsx`); `handleExitFromError`'s `Promise<void> → void` signature change has a single caller and `ConfirmDialog.onConfirm` already accepts `() => void | Promise<void>`; `deleteRound`/`Alert`/`deleting` are still live via `handleDeleteRound`, no dead code left behind; no `@oga/core` changes needing new tests, no migrations/secrets, no mobile dependency changes, PR targets `dev`, CI is green.

---

PR #683, opened 2026-07-06
fix(web): save-path data-integrity cluster (#660, #662, #664, #661)

cner-smith:
Four web save-path data-integrity bugs from the 2026-07-06 audit, one commit each. All `apps/web`; no `@oga/core`, migrations, secrets, or deps.

## #660 — putts classified by raw distance, not intent
The save path + review UI treated any shot within 30 yd of the pin as a putt regardless of the "No, I'm not putting" dialog — a near-green chip lost `distance_to_target` and inflated the putt count. On unmapped courses `distanceToPin` is 0 for every row, so **every** shot became a putt. Putt-ness now keys on user intent only: lie `green` (pinned when a putt is placed) or club `putter`.

## #662 — manual pin never persisted; drag then nulled distance
On unmapped courses no `hole_scores` row exists until `saveReviewedHole`, so `persistRoundPin`'s UPDATE no-oped and a placed pin lived only in `pinOverride` — gone on reload. Then a shot-marker drag with no resolved pin wrote `distance_to_target: null`, destroying a valid stored distance. Now the upsert persists `pinOverride`, and drag only rewrites distance when a pin resolves.

## #664 — fire-and-forget scorecard + practice-plan saves
`HoleScoreCard.persist` void-fired the upsert (and left `ensureRealHole` rejections unhandled); practice-plan feedback + drill-toggle mutations had no `onError`. A failed write kept the typed value on screen with nothing saved and no message. Added a try/catch + inline error slot on the scorecard, and error surfaces on both practice mutations.

## #661 — ShotEntryModal second consecutive add failed silently
`cancelEdit()` recomputed the draft `shot_number` from a stale render closure, so a second add reused the number and hit `unique (hole_score_id, shot_number)`; `save()` only rethrew and the onClick never caught → unhandled rejection, no UI. Now `shot_number` is derived fresh at save time (max+1) and failures surface inline.

**Runtime-reproduced #661 in a browser** against the dev fixture (two "Add shot" clicks → only shot 5 persisted, console 23505 + 409); re-verified after the fix that both shots persist cleanly.

## Verification
- `pnpm typecheck` (workspace) green; `pnpm --filter @oga/web typecheck` green per commit.
- #661 browser repro before + after (bundled chromium + dev storageState).
- Still wants a manual browser QA pass on the #660/#662 map/review flows and the #664 failure paths.

Closes #660
Closes #662
Closes #664
Closes #661

### DIFF
--- a/apps/web/src/components/round/HoleReviewSheet.tsx
+++ b/apps/web/src/components/round/HoleReviewSheet.tsx
@@ -5,6 +5,7 @@ import {
   buildInitialRows,
   formatClubLabel,
   haversineYards,
+  isPuttShot,
   type Club,
   type LieType,
   type PuttDirectionResult,
@@ -350,7 +351,7 @@ function ShotRow({
   // Putt-ness is user intent only — lie 'green' (set when a putt is placed)
   // or club 'putter'. Raw distance must NOT classify: a chip/bunker inside
   // 30 yd is not a putt, and unmapped rows read distanceToPin 0 (#660).
-  const isPutt = row.lieType === 'green' || row.club === 'putter'
+  const isPutt = isPuttShot(row.lieType, row.club)
   const { toDisplay, toDisplayFt } = useUnits()
   const { bag } = useUserBag()
   // Source the club options from the user's bag, falling back to
--- a/apps/web/src/components/round/HoleReviewSheet.tsx
+++ b/apps/web/src/components/round/HoleReviewSheet.tsx
@@ -2,7 +2,6 @@ import { useEffect, useMemo, useRef, useState } from 'react'
 import {
   DEFAULT_BAG,
   LIE_TYPES,
-  NEAR_GREEN_YARDS,
   buildInitialRows,
   formatClubLabel,
   haversineYards,
@@ -124,8 +123,11 @@ export function HoleReviewSheet({
       baseRows.map((row, idx) => {
         const inline = putts[idx]
         if (!inline) return row
+        // A placed putt is on the green by definition — pin lieType so putt
+        // classification keys off real intent (lie/club), not raw distance.
         return {
           ...row,
+          lieType: 'green',
           puttMade: inline.puttMade,
           puttDistanceResult: inline.puttDistanceResult,
           puttDirectionResult: inline.puttDirectionResult,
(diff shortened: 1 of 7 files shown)


cner-smith:
## Review — fix(web): save-path data-integrity cluster (#660, #662, #664, #661)

**Concerns**
- `apps/web/src/components/round/HoleReviewSheet.tsx:353`, `apps/web/src/pages/rounds/hooks/useRoundActions.ts:585-586`, `apps/web/src/pages/rounds/hooks/useRoundActions.ts:644` — the putt-classification predicate (`lieType === 'green' || club === 'putter'`) is now duplicated verbatim in three places (plus a fourth close variant, `editIsPutt`, in `ShotEntryModal.tsx`'s `handleMiniMapDrag`). Each copy carries a comment saying "match HoleReviewSheet's isPutt" — i.e., the fix already documents that these must be kept in sync by hand. That's the exact failure mode #660 itself was: one of these copies (the old `NEAR_GREEN_YARDS` distance check) had drifted from the others. Per CONTRIBUTING.md, "shared business math... post-round inference... shared domain types live in `packages/core/src/`... don't add helpers [web/mobile apps] would want to reuse — lift them to `@oga/core`." Shot/putt classification feeds SG calc and persisted stats, so it reads as exactly this category. Would rather see one `isPutt(shot)` helper in `@oga/core` (with a test) imported at all sites than four independently-maintained copies.
- `apps/web/src/components/rounds/ShotEntryModal.tsx:152-154` — `shotNumber` is derived from `holeShotsRef.current`, which mirrors the React Query cache. `useCreateShot`'s `onSuccess` (`apps/web/src/hooks/useShots.ts:48`) calls `invalidateShotMutationKeys` without returning/awaiting the resulting promise, so `mutateAsync` — and therefore `cancelEdit()`, which re-enables the "Add shot" button — resolves before the shot list has actually refetched. A fast second "Add shot" click before that background refetch lands would still derive the same `shot_number` from the stale cache and hit the same unique-constraint collision #661 reports. It's a genuine improvement (the failure now surfaces via `saveError` instead of vanishing as an unhandled rejection), but the race itself isn't fully closed, just narrowed and made visible. Worth being explicit in the PR description that this is a mitigation, not a full fix, unless I'm missing something that prevents the fast-second-click case.
- Scope: this PR bundles four independently-tracked, independently-diagnosed bugs (#660, #661, #662, #664) — the PR body itself frames it as "one commit each." CONTRIBUTING's one-concern-per-PR example is about mixing restyle/bugfix/feature, so this may not violate the letter of the rule, but "ship the minimum... then stop" and revert-ability both argue for four small PRs rather than one that touches five files across unrelated code paths. Not blocking since each fix is well-isolated into its own commit, but flagging in case a split was considered and rejected for a reason I'm not seeing.

- Three near-identical inline error-banner blocks now exist (`HoleScoreCard.tsx`, `ShotEntryModal.tsx`, `PracticePlanPage.tsx`). Now that there's a third call site, a shared `<InlineError>` might be worth it — not needed for this PR.

---
Looked good: `!isPutt && effectivePin` guard on the drag-distance rewrite (`useRoundActions.ts:382`) correctly stops the null-out on unmapped/pin-less holes; `pinOverride` is properly scoped and reset per-hole via `SWITCH_HOLE`/`AFTER_SAVE` in the reducer, so the new pin-persist write doesn't leak across holes; error handling on the scorecard/modal/practice-plan saves is a real fix for previously-silent failures; no RLS, secrets, `@oga/core`, or mobile-dependency changes to worry about, and the PR targets `dev`.

---

cner-smith:
Addressed the self-review in b178ef9:

- **isPutt duplication** → extracted `isPuttShot(lieType, club)` to `@oga/core` (with tests) and replaced all five hand-copied predicates. Single source of truth now — the exact drift that caused #660 can't recur.
- **#661 race** → added a monotonic `lastIssuedNumberRef` (reset per hole), so `shot_number = max(freshMax, lastIssued) + 1`. The fast-second-add-before-refetch case is now **fully closed**, not just narrowed. Re-verified in the browser: two consecutive adds persist both shots, zero 23505 errors.

Scope-bundling and the shared `<InlineError>` suggestion left as-is (non-blocking / out of scope).

PR #684, opened 2026-07-06
chore: CI web-build job + audit doc/hygiene corrections (#674, #675, #676)

cner-smith:
Audit CI + docs chores, one commit per issue.

## #674 — CI never ran the web production build
Added a `web-build` job (`pnpm --filter @oga/web build`) to `ci.yml` and `preview.yml`. CI ran typecheck/test/check-vendor/mobile but never the Vite/rollup build, though README/CONTRIBUTING/PR-template claim it does and `vercel.json`'s buildCommand is exactly that — a broken production build could pass CI and fail only at Vercel deploy. Verified the build passes locally before adding the job.

## #675 — public-doc + hygiene corrections
- README quickstart `pnpm dev --filter web` → `pnpm --filter web dev` (turbo needs `@oga/web`).
- CONTRIBUTING: `pnpm typecheck` = 3 workspace packages, not 4 (mobile is separate).
- PR template typecheck checkbox `(4/4)` → `(3/3 workspace packages)`.
- self-hosting.md: dropped the nonexistent "README deploy button" reference.
- Deleted the stray untracked root `app.json`.
- **`.env.example` dup-keys item was a false positive** — each key appears once; the `SUPABASE_*`/`VITE_*`/`EXPO_PUBLIC_*` families are distinct per-runtime vars. No change.

## #676 — stale runbook items
- `docs/ios-submission.md`: `eas.json` production carries `ios.image` (Xcode pin) — corrected the "omits a platform key" claim. (The audit's "no iOS keys" premise was itself outdated — the keys exist.)
- Removed the dead "JetBrainsMono-Medium not bundled" font gap — zero JetBrainsMono refs remain in mobile source (replaced by bundled Inconsolata).
- Version-bump staleness lives in the gitignored internal runbook; corrected there directly.

## Verification
`pnpm --filter @oga/web build` green locally; workflow YAML mirrors the existing typecheck job's setup.

Closes #674
Closes #675
Closes #676

PR #638, opened 2026-07-04
feat(web): live-round putting parity with mobile

cner-smith:
Brings the web **live-round** putting sheet up to the mobile `PuttingSheet` feature set. Previously the web live putt sheet captured only distance + short/long + miss left/right + made; mobile (and the web *past-round* `ShotEntryModal`) capture the full green read.

## What's added to `WebPuttingSheet`
- Interactive **green diagram** (reuses existing `round/GreenDiagram`) — drag the aim/start line, which derives horizontal break via `horizontalBreakFromAim`
- **Break (slope)** — uphill / flat / downhill
- **How much** — slope intensity 0–4
- **Speed** — slow / medium / fast
- **Notes**

## Persistence
The live save path (`useRoundActions.saveReviewedHole`) now writes `break_direction[_vertical|_horizontal]`, `green_speed`, `putt_slope_pct`, `aim_offset_yards`, and `notes` — mirroring the past-round `ShotEntryModal.save()` writer. All target columns already exist (migrations 0005 / 0006 / 0028); **no migration, no new `@oga/core` export**.

## Changes
- `@oga/core` `ReviewedShotRow` gains optional putt fields (additive; `buildInitialRows` leaves them undefined)
- `WebPuttData` expanded to the mobile `PuttingValue` shape
- `HoleReviewSheet` threads the new fields from the tap-time sheet onto each row
- `useRoundActions` persists them on save

## Verification
- `pnpm typecheck` ✅ (3/3 packages)
- `pnpm test` ✅ (551 core tests)

## Notes
- v1 persists what the tap-time sheet captures; adding edit UI for these fields to the review-sheet rows is a follow-up if wanted.
- Independent of the map redesign in #637 (disjoint files).

PR #685, opened 2026-07-07
release(web): save-path fixes, stats correctness, Course Planner → prod

### LINKED ISSUE #385: feat(web): Course Planner — pre-round strategy mode with dispersion overlays

cner-smith:
Web-only release to `main`/prod. Cherry-picked the web + core commits from `dev`; the mobile live-round fixes (#645/#646/#681/#682) stay on `dev` for on-device QA and ship in a separate `dev→main` PR after testing.

## Included (5 commits, in dev order)
- **#638** — live-round putting parity (web)
- **#633 / #385** — Course Planner MVP (web)
- **#680** — stats correctness: SG inflation, GIR inference, mobile handicap baseline (core; also carries 2 `apps/mobile` stats files — `stats.tsx` handicap + `completeRound.ts` GIR — pulled in as a dependency of #683, benign stats refinements, not the live-round fixes under device test)
- **#683** — web save-path data-integrity cluster + `isPuttShot` core extraction
- **#684** — CI web-build job + doc/hygiene corrections

## Why #680 rides along
#638 → #680 → #683 form a chain on `useRoundActions.ts` / `HoleReviewSheet.tsx` / `round.ts` (the `inferHoleStats` `holedOut` param + `isPuttShot`), so the web save-path fix can't cherry-pick without it.

## Verification
`pnpm test` (573 core), `pnpm typecheck` (3 pkgs), `pnpm --filter @oga/web build` all green on this branch. #661 was browser-reproduced + re-verified fixed earlier.

## Deploy note
Merging deploys web to prod via Vercel. Mobile is unaffected — no store submission happens on merge.

Closes #660
Closes #661
Closes #662
Closes #664
Closes #668
Closes #669
Closes #673
Closes #674
Closes #675
Closes #676
Closes #385


cner-smith (commented):
## Review — release(web): save-path fixes, stats correctness, Course Planner → prod

**Blockers**
- `apps/web/src/components/rounds/ShotEntryModal.tsx:164-167,229-237` — `lastIssuedNumberRef.current` is bumped *before* the insert (`if (!editing) lastIssuedNumberRef.current = shotNumber`, line 167), but the `catch` block (229-237) only calls `setSaveError(...)` and never rolls it back. If a non-editing save throws (transient Supabase error, RLS denial, network blip) and the user retries, `shotNumber` is computed as `Math.max(freshMax, lastIssuedNumberRef.current) + 1` — since the ref is already bumped past `freshMax`, the retry skips a number. The failed shot number is never issued, leaving a permanent gap in `shot_number` for that hole, which breaks the `editingNext` lookup a few lines below (relies on contiguous numbering) and silently corrupts downstream stats inference. This is the same class of save-path bug #661 (referenced in the comment right above this code) was written to fix — the happy-path race is fixed, but the failure/retry path introduces a new instance of it.

**Concerns**
- `apps/web/src/pages/rounds/hooks/useRoundActions.ts:705` and `apps/web/src/components/rounds/ShotEntryModal.tsx:216` — the putt aim-offset inches→yards conversion (`Math.round((x / 36) * 10) / 10`) is duplicated verbatim across two web call sites in this same PR. CONTRIBUTING.md: "units conversions... live in `packages/core/src/`... Don't add helpers to `apps/web/src/lib`... that other surfaces would want to reuse." This already has two callers within one PR — it should be a named `@oga/core` helper, not a bare inline formula in two places (and mobile will need the same conversion eventually).
- `apps/web/src/hooks/useCoursePlanData.ts:1-17` — new hook extracted into its own file with exactly one call site (`HolePlanPage.tsx`). It's a trivial composition of `useAuth` + `useHolesForCourse` + `useClubDispersion` with no readability barrier. Per the 3-caller rule ("Don't extract a helper, hook, or component into its own function or file unless it has three or more callers, or the inline version is genuinely unreadable"), this reads like it should be inlined in `HolePlanPage.tsx`.
- `apps/web/src/pages/plan/HolePlanPage.tsx:107-119` (`addLeg`) — the new Course Planner auto-computes an initial aim point from median club-carry distance and bearing before any user input (`destinationYards(prevLeg.aim, bearing, carryDist)`), for every leg added. CONTRIBUTING.md's design pillars include "aim point is always explicit user input." This planner is read-only and doesn't write to the round/shot dispersion dataset, and the actual round-logging path (`useRoundActions.ts` around line 2140) explicitly guards against an untouched auto-spawned aim entering that dataset — so this may be intentional as a distinct planning/advisory surface rather than a violation of the shot-tracking pillar. Flagging for a maintainer call on whether the pillar is meant to extend here, since the UI does present a computed point as the starting aim.
- `apps/mobile/lib/completeRound.ts:96` — `const holedOut = holeShots.length === hs.score` infers whether the last logged shot was the holing stroke purely from a count match. It's a reasonable heuristic (and the comment above explains the intent, including that incomplete logs stay conservative), but there's no check that the *last* shot by `shot_number` is actually the one on the green — a duplicate/corrective row plus a manually-corrected `score` could produce a count match without the assumption holding. `holedOut` directly gates the tightened GIR/fairway threshold in `inferHoleStats`, so a false positive here would mis-credit a GIR. Possibly acceptable given how narrow the edge case is — worth a second look.

**Nits**
- `apps/web/src/pages/plan/HolePlanPage.tsx:150-159` — the hole-level expected-score sum (`legs.length - 1 + sgAtAim(...).expectedStrokes`) is SG-domain math built inline in the page component rather than exported from `packages/core` alongside `sgAtAim`. Only one call site today, so lower urgency than the items above, but it's the same category of logic the rest of this PR correctly lifted to core.
- `apps/web/src/pages/plan/HolePlanPage.tsx:162` — `// eslint-disable-next-line react-hooks/exhaustive-deps` has no comment explaining why the deps are narrowed to `pin?.lat, pin?.lng`. CONTRIBUTING.md requires an inline reason for any eslint-disable.
- `apps/web/src/components/round/WebPuttingSheet.tsx:811-822` — `distanceFt` (module scope) and `dist` inside `commit()` recompute the same `Math.max(MIN_PUTT_FT, Math.round(inputToFeet(parsedDist)))` independently, differing only in null-handling. Likely intentional, but the duplication risks silent drift on a future edit.

---
Looked good: no secrets, no migrations/RLS surface touched, no `any`-silencing, no mobile dependency/`nativewind` changes; every new `@oga/core` export in this diff (`sgAtAim`, `expectedStrokesFromPin`, `coneRingGeoJSON`, `isPuttShot`, `inferHoleStats`'s `holedOut` param) has accompanying tests; GIR/fairway threshold logic, SG geometry, and the `isPuttShot` call-site migration all check out against their tests; the previously fire-and-forget save paths now surface errors instead of swallowing them.

---

cner-smith:
## Review — release(web): save-path fixes, stats correctness, Course Planner → prod

**Concerns**
- `apps/web/src/components/rounds/ShotEntryModal.tsx:168` (and `:249`) — `isPuttSave`/`isPutt` still use the old narrow predicate `draft.lieType === 'green'`, while this same file now imports and uses `isPuttShot(lieType, club)` two other places (`:264`, `:643`). `round.ts`'s own doc comment on `isPuttShot` says it exists as the "single source of truth... every surface must agree" specifically to kill the #660 class of hand-copied-predicate drift. The club picker here (`ChipGroup` at the "Club" field) is only hidden when `isPutt` is true, i.e. it's visible — and offers `putter` as an option (`DEFAULT_BAG` includes `{ club_type: 'putter', sort_order: 13 }`, `packages/core/src/constants.ts:167`) — whenever `lieType !== 'green'`. So a "Texas wedge" (putter played from fringe/fairway) saves with `isPuttSave = false`: it's written with `distance_to_target` in yards and null putt fields. But `formatShotSummary` (`:643`) and every other read path classify the same row as a putt via `isPuttShot` (since `club === 'putter'`), and read `distance_to_target` yards as if it were feet via `formatPuttDistance`, producing a ~3x-wrong displayed distance. This looks like exactly the read/write classification mismatch this PR's `isPuttShot` extraction was meant to eliminate, just missed in one spot. Not introduced by this PR (the predicate predates it), but the PR touches this exact file for the exact same bug class and leaves this instance in place.
- `packages/core/src/holeInference.ts:2496-2503` (JSDoc on `inferHoleStats`) vs `apps/mobile/lib/completeRound.ts:156-160` — the doc comment says "Mobile has no reliable holing signal for off-green hole-outs... so mobile callers omit it and get the conservative default," but `completeRound.ts` now computes `holedOut = holeShots.length === hs.score` and passes it explicitly. The comment is stale/inaccurate relative to the actual mobile behavior being shipped in this same PR.
- `apps/web/src/pages/plan/HolePlanPage.tsx:1501-1518` — the Course Planner auto-computes an initial aim point (`destinationYards(tee, bearing, carryDist)` from the player's median carry for an auto-selected club) on load, before any user interaction. It's clearly a draggable, non-persisted planning default (read-only surface, no round/shot writes), so I don't think it's the kind of auto-aim CONTRIBUTING's "aim point is always explicit user input" pillar is aimed at (that reads as being about live shot-tracking integrity), but this is a new surface that pillar wasn't written with in mind, and I'm not fully certain the maintainers intended the pillar's scope to exclude planning tools. Worth a second look.
- PR description frames this as a "web-only release" with mobile "unaffected," and it's true no store submission happens on merge — but `apps/mobile/lib/completeRound.ts` and `apps/mobile/app/(app)/stats.tsx` do land on `main` here, ahead of the mobile live-round fixes (#645/#646/#681/#682) still awaiting on-device QA on `dev`. Given the mobile files are genuinely part of #680's stated stats-correctness scope (not just a compile dependency of #683), this is probably fine, but flagging so it's a deliberate call rather than an incidental one for whoever cuts the next mobile build off `main`.

**Nits**
- `apps/web/src/components/plan/HoleStrategyMap.tsx` (343 lines) and `SGFeedbackCard.tsx` (90 lines) are single-caller components split into their own files, which is a literal read of CONTRIBUTING's 3-caller rule ("don't extract... unless it has three or more callers, or the inline version is genuinely unreadable"). Inlining a full Mapbox integration into `HolePlanPage.tsx` would plausibly qualify as "genuinely unreadable," so this looks like a defensible use of that carve-out rather than a violation.

---
Looked good: `isPuttShot`/`inferHoleStats(holedOut)`/`approachByDistance` core changes are well-isolated in `packages/core`, come with direct unit tests for the new/changed branches, and the GIR/fairway inference fixes checked out correctly against the par/stroke-number logic by hand.

---

PR #686, opened 2026-07-07
fix(mobile): live-round camera — past-round up-the-hole orientation + greenside PLACE_BALL zoom

cner-smith:
Two small live-round camera fixes found during on-device (sideload) QA, one commit each. Both in the shared camera path (`useHoleCamera` / `PastRoundMap`).

## 1. Past-round map now orients "up the hole" (like live)
`PastRoundMap` centered on the pin (`effectivePin ?? tee ?? …`), but `useHoleCamera` derives the up-the-hole heading from `center → pin`. Centering on the pin made origin == target → `headingUpTheHole`'s `<5yd` guard tripped → **north-up**. The live view centers tee-first, which is why only it oriented.

Flip the center priority to `tee ?? effectivePin ?? courseCenter ?? OKC_FALLBACK`, mirroring `LiveRoundSession`. **Parity, not beyond:** synthetic holes with no tee geometry still fall through and stay north-up — same as live. Verified `center` feeds only the `<HoleMap>` prop (no pin-centered logic depends on it).

## 2. Greenside PLACE_BALL frame stays tight
Marking a shot on/around the green snapped to a flat `zoom 17` (fairway-approach framing) — a jarring zoom-out from a green close-up when logging a chip/putt. SET_AIM already scales by ball→pin distance; the PLACE_BALL reframe didn't. Gave it the same signal, **capped at 17 on the loose end** so ≥80-yd approaches keep their current framing and only the short end tightens (60-80→17.5, 30-60→18, <30→19). No resolvable pin → keep 17.

### On the GPS question
Deliberately *not* GPS-based — GPS says where you are, not where the green is. Remaining distance (ball→pin) is the real "near the green" signal, and it's the same one SET_AIM already uses.

## Verification
`npm run typecheck` (mobile) clean. **Not yet device-tested** — wants an on-device pass on the next EAS build (both are camera-framing changes; unit tests don't cover Mapbox camera behavior).

### DIFF
--- a/apps/mobile/components/round/PastRoundMap.tsx
+++ b/apps/mobile/components/round/PastRoundMap.tsx
@@ -185,9 +185,12 @@ export function PastRoundMap({
   )
   // Tee-first so the camera orients "up the hole" — useHoleCamera computes
   // heading from center→pin, and centering on the pin (as before) made
-  // origin == target, collapsing the bearing to north-up. Mirrors the live
-  // view's center priority (LiveRoundSession: tee ?? courseCenter ?? ball).
-  const center = tee ?? effectivePin ?? courseCenter ?? OKC_FALLBACK
+  // origin == target, collapsing the bearing to north-up. Pin is deliberately
+  // NOT in the fallback chain, matching the live view (LiveRoundSession:
+  // tee ?? courseCenter ?? ball): a hole with a round-pin but no surveyed tee
+  // would otherwise resolve center == pin and re-collapse the heading. When no
+  // tee resolves, courseCenter still gives a non-coincident origin.
+  const center = tee ?? courseCenter ?? OKC_FALLBACK
 
   const [placed, setPlaced] = useState<PlacedShot[]>([])
   const [activeIdx, setActiveIdx] = useState(0)
--- a/apps/mobile/components/round/PastRoundMap.tsx
+++ b/apps/mobile/components/round/PastRoundMap.tsx
@@ -183,7 +183,11 @@ export function PastRoundMap({
         : null,
     [currentHole],
   )
-  const center = effectivePin ?? tee ?? courseCenter ?? OKC_FALLBACK
+  // Tee-first so the camera orients "up the hole" — useHoleCamera computes
+  // heading from center→pin, and centering on the pin (as before) made
+  // origin == target, collapsing the bearing to north-up. Mirrors the live
+  // view's center priority (LiveRoundSession: tee ?? courseCenter ?? ball).
+  const center = tee ?? effectivePin ?? courseCenter ?? OKC_FALLBACK
 
   const [placed, setPlaced] = useState<PlacedShot[]>([])
   const [activeIdx, setActiveIdx] = useState(0)
--- a/apps/mobile/components/round/hooks/useHoleCamera.ts
+++ b/apps/mobile/components/round/hooks/useHoleCamera.ts
@@ -273,12 +273,26 @@ export function useHoleCamera({
     if (phase !== 'PLACE_BALL') return
     if (!cameraRef.current) return
     if (!ball) return
+    // Tighten the frame near the green so marking a ball on/around the green
+    // stays a green close-up. A flat zoom 17 (fairway-approach framing) read
+    // as a jarring zoom-out when marking a greenside/putt shot. Remaining
+    // distance (ball→pin) is the signal — same one SET_AIM uses. Capped at 17
+    // on the loose end so ≥80-yd approaches keep the existing framing; only
+    // the short end tightens. No pin resolvable → keep 17.
+    const target = roundPin ?? pin ?? null
+    const distYd = target ? distanceYards(ball, target) : null
+    const zoom =
+      distYd == null ? 17
+      : distYd >= 80 ? 17
+      : distYd >= 60 ? 17.5
+      : distYd >= 30 ? 18
+      : 19
     try {
       cameraRef.current.setCamera({
         centerCoordinate: toCoord(ball),
-        zoomLevel: 17,
+        zoomLevel: zoom,
         pitch: 0,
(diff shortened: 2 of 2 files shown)


cner-smith (commented):
## Review — fix(mobile): live-round camera — past-round up-the-hole orientation + greenside PLACE_BALL zoom

**Concerns**
- `apps/mobile/components/round/PastRoundMap.tsx:190` — the new priority is `tee ?? effectivePin ?? courseCenter ?? OKC_FALLBACK`, and the comment above it (line 186-189) claims this "mirrors the live view's center priority (LiveRoundSession: tee ?? courseCenter ?? ball)." It doesn't: `LiveRoundSession.tsx:157-163` explicitly excludes the pin from the fallback chain — "Pin/green is intentionally NOT a fallback; it would mis-frame the hole every time" — and falls straight from `tee` to `courseCenter`. This PR's version puts `effectivePin` *before* `courseCenter`. That reintroduces the exact bug this PR fixes, just gated on a different condition: for a synthetic-course hole with no surveyed tee (`tee` null) but a pin already set for the round (`roundPin`/`storedPin` non-null — a normal post-round state once a player has used PIN mode during live play), `center` resolves to `effectivePin`, and `useHoleCamera`'s target for the initial/hole-change frames is `roundPin ?? pin` — the same point. Origin == target again → `headingUpTheHole`'s `<5yd` guard trips → north-up, silently, for exactly the "synthetic hole, no tee" case the PR's own description calls out as an accepted no-op ("still fall through and stay north-up — same as live"). It isn't the same as live, because live never lets pin become the anchor. Might be worth dropping `effectivePin` from this fallback chain entirely (matching live) rather than fixing the comment, since the whole point of the PR is to stop pin-as-origin collapses.

**Nits**
- `apps/mobile/components/round/hooks/useHoleCamera.ts:284-289` vs `:334-340` — the new PLACE_BALL zoom staircase (17 / 17.5 / 18 / 19 at the 80/60/30-yd breaks) duplicates the tail of the existing SET_AIM staircase almost value-for-value, just without the 150/300-yd bands SET_AIM needs for its wider start zoom. Two call sites, so this is inside the 3-caller-rule's "two similar lines are fine" allowance, but if a third phase ever wants distance-based zoom, a shared `zoomForDistance` helper would be the move rather than a third copy.
- Two independent camera bugs (tee-first orientation in `PastRoundMap`, greenside zoom in `useHoleCamera`'s PLACE_BALL reframe) are bundled in one PR, each with its own commit. They share a file/subsystem and both came out of the same on-device QA pass, so this reads more like "related cleanup" than the restyle+bugfix+feature mixing CONTRIBUTING.md's one-concern-per-PR rule is aimed at — flagging only because they are two independently-revertable fixes, not because I think this needs a split.

---
Looked good: no secrets, no `@oga/core` surface touched (so no test-coverage gap), no migrations, no mobile dependency/`nativewind` changes, targets `dev`; the `headingUpTheHole` reuse and the ball→pin distance signal for zoom are consistent with the existing SET_AIM pattern in the same file; PR description is candid about not being device-tested yet.

---

cner-smith:
## Review — fix(mobile): live-round camera — past-round up-the-hole orientation + greenside PLACE_BALL zoom

**Concerns**
- `apps/mobile/components/round/PastRoundMap.tsx:186-190` — inserting `effectivePin` into the center chain (`tee ?? effectivePin ?? courseCenter ?? OKC_FALLBACK`) can reproduce the exact bug this PR sets out to fix. When a hole has no tee coords but does have a resolved pin (a round-logged pin with no crawled tee — the "synthetic holes" case the PR body itself calls out), `center` resolves to `effectivePin`, and `useHoleCamera`'s heading effects (`apps/mobile/components/round/hooks/useHoleCamera.ts:89` and `:124`) compute `headingUpTheHole(center, roundPin ?? pin)` with origin == target again → the `<5yd` guard trips → north-up, same collapse the PR is fixing. The added comment says this "Mirrors the live view's center priority (LiveRoundSession: tee ?? courseCenter ?? ball)," but `LiveRoundSession.tsx:152-153` is explicit that pin is deliberately excluded ("Pin/green is intentionally NOT a fallback; it would mis-frame the hole every time"). The new PastRoundMap chain doesn't actually mirror that — it reintroduces the exact case live avoids. Unclear from the diff how common "pin set, no tee" holes are in practice; if that's genuinely rare this may be a non-issue, but worth confirming or dropping `effectivePin` from the chain to match live's stated invariant.

- This PR bundles two independently-discovered bug fixes (orientation center-priority in `PastRoundMap`, and PLACE_BALL zoom-by-distance in `useHoleCamera`) as two commits in one PR. CONTRIBUTING.md: "One concern per PR. UI restyle, bug fix, and new feature do not belong in the same PR" and "PRs that add scope beyond their stated concern will be asked to split." Both are bug fixes in the same shared file, but they're unrelated symptoms with separate root causes and are independently revertible — arguably two PRs rather than one, even though neither is a feature/restyle mix.

**Nits**
- `apps/mobile/components/round/hooks/useHoleCamera.ts:282-289` duplicates the low end of the SET_AIM zoom-by-distance staircase (`useHoleCamera.ts:333-340`) almost verbatim — same 80/60/30 thresholds, same 17/17.5/18/19 values. Two similar chains are within CONTRIBUTING's "two similar lines are fine" allowance, but if a third call site shows up, a shared `zoomForDistance` helper would avoid the two chains drifting out of sync.

- If the zoom-by-distance staircase pattern grows a third use site, consider lifting it into a small shared helper rather than a third inline ternary chain.

---
Looked good: the zoom tier boundaries (80/60/30, capped at 17 on the loose end) are internally consistent and monotonic with no gaps/off-by-ones, match SET_AIM's existing thresholds, and the null-pin fallback to zoom 17 preserves prior behavior rather than silently masking an error.

---

PR #687, opened 2026-07-07
fix(mobile): swipe-dismissed sheets reopen off-screen (#644 residual)

cner-smith:
Bug found during on-device QA of the past-round logger.

## Repro
Log a past round → open the shot sheet → swipe down to dismiss → reopen it → **it's off-screen and never comes back** (workaround: leave the round and re-enter, which unmounts the sheet).

## Cause
The #646 swipe hook left `translateY` at the drag offset on dismiss, relying on the sheet's `<Modal>` unmounting its children to dispose the shared value ("No manual reset needed"). That holds for sheets rendered **inside** a parent `<Modal>` (PuttingSheet, ScorecardModal — they fully unmount), but **not** for sheets that are always mounted and only toggle their **own** inner Modal's `visible` (**ShotLogger**, **PastHoleShotsSheet**) — the hook lives in the component body, so the offset persists across close/reopen. This is exactly the residual the #646 self-review flagged (and the triage wrongly dismissed).

## Fix
`useSwipeToDismiss(onClose, isOpen?)` — resets `translateY` to 0 on each (re)open, done while the sheet is hidden so the slide-out exit animation is untouched (resetting on *dismiss* would break the "slide out from where you left it" exit). Passed `visible` at the two always-mounted call sites. The nested `EditShotSheet` is a keyed ternary (remounts per open) and the parent-`<Modal>` sheets unmount, so they reset for free — `isOpen` defaults to `true`, leaving them unchanged.

## Verification
`npm run typecheck` (mobile) clean. **Not yet device-tested** — wants an on-device pass on the next EAS build (gesture/animation behavior; no unit coverage for Reanimated).

### DIFF
--- a/apps/mobile/components/round/PastHoleShotsSheet.tsx
+++ b/apps/mobile/components/round/PastHoleShotsSheet.tsx
@@ -158,7 +158,7 @@ export function PastHoleShotsSheet({
   // One <Modal> with discriminated content (list vs editor) — NOT two
   // sibling Modals. iOS allows one presented modal per presenter, so the
   // old stacked-Modal edit flow silently failed to present (#293/#495).
-  const { pan, cardStyle } = useSwipeToDismiss(onClose)
+  const { pan, cardStyle } = useSwipeToDismiss(onClose, visible)
 
   return (
     <Modal
--- a/apps/mobile/components/round/ShotLogger.tsx
+++ b/apps/mobile/components/round/ShotLogger.tsx
@@ -137,7 +137,7 @@ export function ShotLogger({
   const isOnGreen = value.lieType === 'green'
   // Swipe-dismiss for the non-putt card; the on-green branch delegates to
   // PuttingSheet, which owns its own swipe-dismiss.
-  const { pan, cardStyle } = useSwipeToDismiss(onClose)
+  const { pan, cardStyle } = useSwipeToDismiss(onClose, visible)
 
   return (
     <Modal
--- a/apps/mobile/components/ui/useSwipeToDismiss.ts
+++ b/apps/mobile/components/ui/useSwipeToDismiss.ts
@@ -1,4 +1,4 @@
-import { useEffect } from 'react'
+import { useLayoutEffect } from 'react'
 import { Gesture } from 'react-native-gesture-handler'
 import {
   runOnJS,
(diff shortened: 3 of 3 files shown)


cner-smith:
## Review — fix(mobile): swipe-dismissed sheets reopen off-screen (#644 residual)

**Concerns**
- `apps/mobile/components/ui/useSwipeToDismiss.ts:43-45` — the reset runs in a plain `useEffect`, which fires after commit/paint, not before. On reopen, the newly-mounted `Animated.View`/`GestureDetector` subtree (unmounted while the parent `<Modal>` was hidden) first renders with `cardStyle` still reflecting the *stale* `translateY.value` from the previous dismiss, then snaps to 0 once the effect runs. Since the RN `Modal` slide-in animation plays over the same window, this could show as a visible jump partway through the open animation rather than a clean open — the opposite of what the fix intends ("done while hidden"). Using `useLayoutEffect` instead would apply the reset before the first paint of the reopened subtree. The PR body already flags this as not yet device-tested ("no unit coverage for Reanimated"); worth specifically watching for a jump-on-reopen when doing that pass.

**Nits**
- PR title — CONTRIBUTING.md's "Opening a PR" section asks for "Title in imperative present tense (`Add lie-slope filter to patterns`)." "swipe-dismissed sheets reopen off-screen" describes the bug rather than the fix; something like `fix(mobile): reset sheet position on reopen (#644 residual)` would match the convention used by sibling merged PRs in this repo's history (e.g. "harden shot sync queue...").

---
Looked good: I traced all 5 call sites of `useSwipeToDismiss` (`ShotLogger.tsx:140`, `PastHoleShotsSheet.tsx:161,482`, `Scorecard.tsx:54`, `PuttingSheet.tsx:191`) against how each sheet actually mounts/unmounts (RN `<Modal visible>` unmounts its children, so `PuttingSheet`/`ScorecardModal` and the keyed `EditShotSheet` get a fresh shared value for free) and the `isOpen = true` default and the two new `visible` passes at `ShotLogger.tsx:140` / `PastHoleShotsSheet.tsx:161` line up correctly with that; the guard against resetting on the dismiss path (only resetting `if (isOpen)`) is correct and preserves the exit animation as intended; single concern, appropriately small diff.

---

Issue #524, opened 2026-06-08
[mobile] Make the Learn 'yardage book' discoverable — signal it's a list of articles


cner-smith:
**From on-device QA (2026-06-08).**

On mobile, the Learn content (the 'yardage book') is hard to stumble onto — it's not obvious it's a library of helpful articles unless you go looking. Make the entry point clearer.

- Clearer entry-point label/affordance that signals 'helpful articles / guides', not a single page.
- The dashboard already nudges '↓ New to a stat? The yardage book's at the bottom' — make the destination itself read as a browsable list (title + 'short reads on every stat' framing, article count, etc.).
- Decide the right surface (tab? prominent card on Home/Practice? section header) deliberately.

Note: the Learn articles already render on mobile (per-article RN ports). This is purely discoverability/IA, not content.

cner-smith:
Shipped in the 0.9.0 release (PR #552, via #535) — now live on `main`/prod. Closing as done.

PR #688, opened 2026-07-07
feat(mobile): Learn access point in the Practice header (#524)

cner-smith:
Option A from the discoverability discussion — a title-bar entry point for Learn.

## Why
Learn (20 articles) was only reachable from the home-page CTA + a buried body link, so it was easy to miss. Adds a **"Learn" chip** (book icon + label) to the Practice `AppBar`'s free `right` slot, routing to the existing `/(app)/learn` route. The Practice screen is the natural home for it (improvement-focused), and it's now always visible in the header.

## Notes
- Uses the standard `PressableTouch` (iOS press-dim) + `<Link asChild>` pattern; matches the AppBar's cream-on-forest styling.
- Non-destructive: the existing home CTA + body links stay.
- Scoped to Practice per your suggestion. Stats' `right` slot is occupied by the L5/L10/L20 toggle — easy to add there too later if you want it in more places.

Addresses #524. **Not yet device-tested** — wants a visual confirmation on the next EAS build (typecheck green; RN layout not covered by unit tests).

### DIFF
--- a/apps/mobile/app/(app)/practice.tsx
+++ b/apps/mobile/app/(app)/practice.tsx
@@ -113,6 +113,8 @@ export default function Practice() {
         right={
           <Link href={'/(app)/learn' as never} asChild>
             <PressableTouch
+              accessibilityRole="button"
+              accessibilityLabel="Learn articles"
               style={{
                 flexDirection: 'row',
                 alignItems: 'center',
@@ -124,8 +126,8 @@ export default function Practice() {
                 borderRadius: 2,
               }}
             >
-              <MaterialCommunityIcons name="book-open-variant" size={14} color="#F2EEE5" />
-              <Text style={{ color: '#F2EEE5', fontSize: 12, fontWeight: '600', letterSpacing: 0.3 }}>
+              <MaterialCommunityIcons name="book-open-variant" size={14} color={CREAM} />
+              <Text style={{ color: CREAM, fontSize: 12, fontWeight: '600', letterSpacing: 0.3 }}>
                 Learn
               </Text>
             </PressableTouch>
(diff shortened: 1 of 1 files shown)


cner-smith (commented):
## Review — feat(mobile): Learn access point in the Practice header (#524)

**Nits**
- `apps/mobile/app/(app)/practice.tsx:127-128` — the new chip hardcodes `color="#F2EEE5"` twice, but this exact file already defines `const CREAM = '#F2EEE5'` (line 23) and uses it elsewhere (`backgroundColor: CREAM`). Worth reusing the constant instead of a second magic-string copy of the same value.
- `apps/mobile/app/(app)/practice.tsx:110-134` — the new chip has no `accessibilityRole`/`accessibilityLabel`. Not a hard rule here (the plain `<Pressable>` "Browse all drills →" links in this same file also skip it), but the `PressableTouch` CTAs elsewhere in this file (e.g. `NoPlan`'s "Generate this week's plan", line ~269) do set both — may be worth matching for a screen-reader-announced nav chip.

---
Looked good: single, well-scoped concern (one nav affordance, no unrelated changes); reuses the existing `AppBar` `right` slot and `PressableTouch`/`Link` patterns already established in this same file rather than introducing new abstractions; correctly kept inline rather than extracted into its own component (single caller, under the 3-caller rule); no `@oga/core` or Supabase surface touched so no test/RLS gap.

---

cner-smith:
## Review — feat(mobile): Learn access point in the Practice header (#524)

**Nits**
- PR title — CONTRIBUTING.md's "Opening a PR" section asks for "Title in imperative present tense (`Add lie-slope filter to patterns`)." "Learn access point in the Practice header" describes the addition rather than commanding it; something like `feat(mobile): add Learn chip to Practice header` would match the convention.

---
Looked good: single concern (one nav entry point, no drive-by changes to the home CTA or body link it's additive to), inline `right` slot usage matches `AppBar`'s existing `ReactNode` prop rather than adding a new one, no extraction into a shared component despite only one caller (correctly avoids the 3-caller rule rather than tripping it), and the target route (`/(app)/learn`) already exists.

---

PR #689, opened 2026-07-07
chore(mobile): set up EAS Update (OTA) for review-free JS shipping

cner-smith:
Sets up over-the-air updates so pure-JS fixes ship to installed apps **without an App Store / Play review**. Apple/Google permit interpreted-code (JS) updates that add no native code and don't change the app's purpose — which is exactly the shape of the camera/swipe/Learn fixes we've been making.

## What changed
- **expo-updates 0.28.18** (SDK-53 version), exact-pinned like `expo` core.
- **app.config.ts:** `updates.url` (project `f852bb53…`) + `runtimeVersion: { policy: 'fingerprint' }`.
- **eas.json:** per-profile `channel` (development / preview / production) → update branches.

## Why `fingerprint` runtimeVersion
It's the safe policy: an OTA update only lands on a build whose **native fingerprint matches**, so a JS update can never reach a binary missing its native code. The simpler `appVersion` policy has a footgun (JS update landing on a native-mismatched build → crash) that fingerprint removes.

## ⚠️ Activation (important)
`expo-updates` is a native module, so OTA is **not live yet**. It activates only after a **new binary built with this config is submitted + approved once** (the currently-shipped store build has no updates runtime). After that:
```
eas update --branch production --message "fix: ..."
```
ships JS fixes to the production build with no review. Native changes still need a reviewed binary.

## Verification
`npm run typecheck` clean; `expo config` resolves updates + runtimeVersion; `expo-doctor` 17/18 (the 1 fail is the pre-existing intentional `metro.config.js` resolver override, unrelated). **Not device-tested** — needs the first expo-updates binary build to validate OTA end-to-end.

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -9,6 +9,22 @@ const config: ExpoConfig = {
   slug: 'oga',
   scheme: 'oga',
   version: '1.0.0',
+  // EAS Update (OTA). Ships JS/asset-only fixes to installed builds WITHOUT an
+  // App Store / Play review — Apple/Google permit interpreted-code updates that
+  // don't add native code or change the app's purpose. Native changes (SDK/RN
+  // bumps, new modules, permission or icon changes) still need a new reviewed
+  // binary. The `fingerprint` runtimeVersion policy enforces exactly that: an
+  // OTA update only lands on a build whose native fingerprint matches, so a JS
+  // update can never reach a binary missing its native code. Channels → update
+  // branches are set per-profile in eas.json. NOTE: the FIRST expo-updates
+  // build must be submitted + approved once before OTA is live (the already-
+  // shipped store build has no updates runtime).
+  updates: {
+    url: 'https://u.expo.dev/f852bb53-02fc-46a1-b509-4b2170cb6d84',
+  },
+  runtimeVersion: {
+    policy: 'fingerprint',
+  },
   orientation: 'portrait',
   // New Architecture stays OFF for the SDK 53 migration. SDK 53 flips it
   // on by default, so this is an active opt-out (set during the SDK 52 step
--- a/apps/mobile/eas.json
+++ b/apps/mobile/eas.json
@@ -7,6 +7,7 @@
     "development": {
       "developmentClient": true,
       "distribution": "internal",
+      "channel": "development",
       "node": "20.18.0",
       "android": {
         "buildType": "apk"
@@ -20,6 +21,7 @@
     },
     "preview": {
       "distribution": "internal",
+      "channel": "preview",
       "node": "20.18.0",
       "android": {
         "buildType": "apk"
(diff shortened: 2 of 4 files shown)


cner-smith (commented):
## Review — chore(mobile): set up EAS Update (OTA) for review-free JS shipping

**Concerns**
- `apps/mobile/package.json:35` vs `apps/mobile/package-lock.json:31` — `package.json` pins `expo-updates` exact (`"0.28.18"`), but the lockfile's root manifest snapshot (`packages[""].dependencies`) records it with a tilde (`"~0.28.18"`). These two are supposed to mirror each other; they don't here. CONTRIBUTING.md's mobile-pinning rule exists specifically because "EAS does fresh installs and any range bump risks runtime version mismatches" — if the EAS build step runs `npm ci` (common for EAS Build when a lockfile is present), a package.json/lockfile mismatch like this fails the install outright. Even if the actual build step runs plain `npm install` and self-heals it, the lockfile as committed doesn't match the stated exact pin. Worth regenerating (`rm apps/mobile/package-lock.json && npm install --legacy-peer-deps`) so both files agree on `0.28.18` before merge.

**Nits**
- `apps/mobile/package-lock.json:2,6` — the `@oga/mobile`/`@oga/core`/`@oga/supabase` version fields jump from `0.0.1` to `0.9.0` in this diff. Confirmed these already read `0.9.0` in `apps/mobile/package.json` on `dev`, so this is just the lockfile catching up to pre-existing drift via the `npm install` this PR ran — not new scope — but it's unexplained in the PR description and could look like scope creep on a quick skim.

---
Looked good: single, well-scoped concern (OTA wiring only — app.config.ts, eas.json channels, the one dependency add); `fingerprint` runtimeVersion policy is the safer choice and the tradeoff vs `appVersion` is explained; targets `dev`; the "not live until a new binary ships" caveat is called out clearly; no secrets, RLS, or domain-pillar surface touched.

---

cner-smith:
## Review — chore(mobile): set up EAS Update (OTA) for review-free JS shipping

**Nits**
- `apps/mobile/package.json:35` — `expo-updates` is added as `~0.28.18`, a tilde range. CONTRIBUTING.md's mobile-pinning section says dependencies "prefer **exact pins** over caret/tilde ranges — EAS does fresh installs and any range bump risks runtime version mismatches." That said, every other `expo-*` entry in this same file (`expo-constants`, `expo-dev-client`, `expo-splash-screen`, etc.) already uses a tilde range, consistent with `expo install`'s own convention — so this follows the file's existing pattern rather than deviating from it. Flagging only because it's a literal read of the stated rule; not asking for a change.

---
Looked good: scoped to exactly one concern (the `chore/` PR the mobile-pinning section itself asks for when adding a dependency), `nativewind` untouched, `runtimeVersion: 'fingerprint'` is the safer of the two policies and the PR body explains why, and the "not live until a reviewed binary ships" caveat is called out up front rather than glossed over.

---

PR #690, opened 2026-07-07
fix(web): bump shot_number guard only after a successful insert

cner-smith:
Ports the #685 review fix to `dev` (the buggy code came from #683, merged here earlier).

**Bug:** `lastIssuedNumberRef` was bumped *before* the insert, so a failed save + retry recomputed `max(freshMax, ref)+1` off the already-bumped ref — skipping the failed number and leaving a permanent `shot_number` gap, which breaks the `editingNext` lookup and stats inference (both assume contiguous numbering). Same class as #661.

**Fix:** move the bump inside the `try`, after the insert resolves, so a failure leaves the ref untouched and the retry reuses the number.

Identical change to #685's fix commit (cherry-picked); web typecheck verified there.


cner-smith:

---

PR #692, opened 2026-07-07
release(mobile): live-round fixes, sync hardening, camera, swipe, Learn, EAS Update → prod

cner-smith:
Ships the accumulated **mobile** work to `main`/prod. The web/core changes are already on `main` (via #685), so this diff is mobile-only: **22 files, +871/−270**.

## Contents
- **Live-round field bugs** (#645 → #639–643): camera orientation/zoom, GPS locate button + auto-place, putting-sheet scroll, "Finish round" finalization.
- **Sync-queue hardening** (#681 → #651/#652): poison-pill quarantine, client-id chunk reconciliation, idempotent inserts, prune synced rows, completeRound sync-lock race.
- **Live-round data-loss traps** (#682 → #650/#653): 9-hole phantom-resume clamp; load-error screen no longer offers a destructive delete.
- **Swipe-to-dismiss** (#646 → #644) + reopen-off-screen fix (#687).
- **Camera** (#686): past-round up-the-hole orientation + greenside PLACE_BALL zoom.
- **Learn access point** in the Practice header (#688 → #524).
- **EAS Update / OTA setup** (#689): `expo-updates` + `runtimeVersion: fingerprint` + channels. ⚠️ OTA activates only after the first production binary built with this config is submitted + approved.

## Review status
- CI green; all `@oga/core` changes (already on main) are unit-tested.
- **A 3-agent code review (logic / UI-UX / ripple effects) is running on this diff.** Findings will be posted before merge.
- **Not yet on-device QA'd** — the plan is: cut a production build off `main` after merge → Play internal track / TestFlight → smoke on device → promote. Do not wide-release without that pass.

## Deploy note
Merging does not deploy mobile anywhere (no store submission on merge). It sets `main` as the base for the next production build.

Closes #639
Closes #640
Closes #641
Closes #642
Closes #643
Closes #644
Closes #650
Closes #651
Closes #652
Closes #653
Closes #524

PR #693, opened 2026-07-07
fix(mobile): review polish — Learn chip hitSlop + PastRoundMap OKC guard

cner-smith:
Two one-liners from the #692 pre-release 3-agent review (both LOW severity):

- **Learn chip tap target** (~26px → ~44pt via `hitSlop`; visual unchanged).
- **PastRoundMap OKC edge:** append `effectivePin` as the last-resort center so a manually-added course with no tee + null lat/lng frames on the pin, not Oklahoma. Pin stays *after* tee/courseCenter, so it can't reintroduce the #686 heading collapse.

Merges to `dev` so it rides into the #692 release. Mobile typecheck clean.

### DIFF
--- a/apps/mobile/app/(app)/practice.tsx
+++ b/apps/mobile/app/(app)/practice.tsx
@@ -115,6 +115,9 @@ export default function Practice() {
             <PressableTouch
               accessibilityRole="button"
               accessibilityLabel="Learn articles"
+              // Extend the touch target to the ~44pt minimum without bloating
+              // the compact chip's visual height (padding stays 6).
+              hitSlop={{ top: 10, bottom: 10, left: 6, right: 6 }}
               style={{
                 flexDirection: 'row',
                 alignItems: 'center',
--- a/apps/mobile/components/round/PastRoundMap.tsx
+++ b/apps/mobile/components/round/PastRoundMap.tsx
@@ -190,7 +190,11 @@ export function PastRoundMap({
   // tee ?? courseCenter ?? ball): a hole with a round-pin but no surveyed tee
   // would otherwise resolve center == pin and re-collapse the heading. When no
   // tee resolves, courseCenter still gives a non-coincident origin.
-  const center = tee ?? courseCenter ?? OKC_FALLBACK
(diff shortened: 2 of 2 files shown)


cner-smith:
## Review — fix(mobile): review polish — Learn chip hitSlop + PastRoundMap OKC guard

**Concerns**
- `apps/mobile/components/round/PastRoundMap.tsx:188-189` — The pre-existing comment directly above the changed line still reads "Pin is deliberately NOT in the fallback chain, matching the live view." That's no longer true: line 197 now reads `tee ?? courseCenter ?? effectivePin ?? OKC_FALLBACK`, putting pin back in the chain as a last resort. The new comment block (193-196) explains the new tier, but the old sentence wasn't updated, so the two comments now contradict each other a few lines apart — a future reader skimming just the first block will believe pin is excluded when it isn't. Related: the PR description says the ordering means this "can't reintroduce the #686 heading collapse," but in the specific case this change targets (no tee, no `courseCenter`, pin present) `center` does resolve to `effectivePin`, which is exactly the origin==target condition that collapses `useHoleCamera`'s bearing to north-up — the in-code comment even concedes this ("north-up, but on the hole"). That's a reasonable degraded fallback, but the PR body overstates the guarantee; worth tightening the old comment (188-189) so it doesn't flatly assert something the code three lines later contradicts.

**Nits**
- This PR bundles two unrelated one-line fixes (the Learn chip `hitSlop` in `practice.tsx` and the `PastRoundMap` fallback-chain change) touching unrelated components. CONTRIBUTING.md's "One concern per PR" rule calls out mixing a UI fix and a bug fix in the same PR. Both are trivial and explicitly scoped as a pre-release cleanup batch, so this is low-stakes, but it's the kind of bundling the rule is written against.

---
Looked good: both changes are minimal and match their stated intent (hitSlop math gets the chip to ~44pt without changing its visual size; the fallback re-ordering keeps pin behind tee/courseCenter so it can't override a resolvable non-coincident origin). No `@oga/core` exports touched, no security/RLS/paywall/lie-slope/aim-point concerns, targets `dev` correctly.

---

PR #695, opened 2026-07-08
fix(mobile): EAS Update runtimeVersion policy fingerprint → appVersion (unblocks prod builds)

cner-smith:
Both production builds failed 2026-07-07 at the **Configure expo-updates** phase (opaque "Unknown error").

**Root cause (researched):** the `fingerprint` runtimeVersion policy makes EAS recompute a native fingerprint on its build servers, which diverges/fails in a **monorepo** (server-side node_modules/lockfile resolution differs from local) — specifically worse with **expo-sqlite** present. Both true for OGA. `fingerprint:generate` runs fine locally, but EAS's resolution of the `file:fingerprint` placeholder fails. Matches [expo/expo#43831](https://github.com/expo/expo/issues/43831) (open, no upstream fix).

**Fix:** `appVersion` policy — runtimeVersion derives from `version` (`1.0.0`), computed identically local + on EAS. **OTA is fully retained**; the only habit it asks is bumping `version` on releases that change native code.

Verified: `expo config` shows `runtimeVersion.policy: appVersion`; mobile typecheck clean. Builds to be re-kicked tomorrow (billing cycle reset).

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -13,17 +13,25 @@ const config: ExpoConfig = {
   // App Store / Play review — Apple/Google permit interpreted-code updates that
   // don't add native code or change the app's purpose. Native changes (SDK/RN
   // bumps, new modules, permission or icon changes) still need a new reviewed
-  // binary. The `fingerprint` runtimeVersion policy enforces exactly that: an
-  // OTA update only lands on a build whose native fingerprint matches, so a JS
-  // update can never reach a binary missing its native code. Channels → update
-  // branches are set per-profile in eas.json. NOTE: the FIRST expo-updates
-  // build must be submitted + approved once before OTA is live (the already-
-  // shipped store build has no updates runtime).
+  // binary — bump `version` when you ship one so the OTA runtimeVersion moves
+  // with it (see policy note below). Channels → update branches are set
+  // per-profile in eas.json. NOTE: the FIRST expo-updates build must be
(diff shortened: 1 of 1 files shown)


cner-smith (commented):
## Review — fix(mobile): EAS Update runtimeVersion policy fingerprint → appVersion (unblocks prod builds)

**Concerns**
- `apps/mobile/app.config.ts:31-32` — `appVersion` ties OTA/native compatibility to a manually-bumped `version` field, while `eas.json`'s `autoIncrement: true` + `appVersionSource: "remote"` only auto-increments the native build number, not `version`. If a future native-code change ships without a `version` bump, an OTA JS update could land on a binary with mismatched native code — the exact hazard `fingerprint` existed to prevent. The comment documents the discipline required, which helps, but there's no enforcement (e.g., a CI check diffing native deps against `version`) backing it up. Given this PR is an unblock for a real outage, that may be an acceptable trade-off for now — flagging so it's a conscious one rather than a rule that quietly stops being followed.

**Nits**
- PR title uses `fix(mobile): ... fingerprint → appVersion (unblocks prod builds)` rather than an imperative sentence (CONTRIBUTING.md: "Title in imperative present tense"). Minor, doesn't affect review.

---
Looked good: single-file, single-concern change (base is `dev`, not `main`); the rationale and prior-failure context are captured directly in comments rather than left implicit, which matches the "why" bar in CONTRIBUTING.md's code-style section; no test/package-boundary/security implications since this is Expo config only.

---

PR #696, opened 2026-07-08
release(mobile): EAS Update runtimeVersion fix (fingerprint → appVersion)

cner-smith:
Single-commit release: the `runtimeVersion` policy fix (#695) that unblocks the production builds. Both prod builds failed 2026-07-07 at 'Configure expo-updates' due to the `fingerprint` policy diverging on EAS servers (monorepo + expo-sqlite; expo/expo#43831). Switched to `appVersion` — OTA fully retained. Getting it onto `main` so tomorrow's production builds pick it up.

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -13,17 +13,25 @@ const config: ExpoConfig = {
   // App Store / Play review — Apple/Google permit interpreted-code updates that
   // don't add native code or change the app's purpose. Native changes (SDK/RN
   // bumps, new modules, permission or icon changes) still need a new reviewed
-  // binary. The `fingerprint` runtimeVersion policy enforces exactly that: an
-  // OTA update only lands on a build whose native fingerprint matches, so a JS
-  // update can never reach a binary missing its native code. Channels → update
-  // branches are set per-profile in eas.json. NOTE: the FIRST expo-updates
-  // build must be submitted + approved once before OTA is live (the already-
-  // shipped store build has no updates runtime).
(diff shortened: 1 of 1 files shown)


cner-smith (commented):
## Review — release(mobile): EAS Update runtimeVersion fix (fingerprint → appVersion)

**Concerns**
- `apps/mobile/app.config.ts:29-38` — switching `runtimeVersion.policy` from `fingerprint` to `appVersion` trades an automatic compatibility check for a manual one: `fingerprint` guaranteed an OTA JS update could never reach a binary whose native code didn't match, computed automatically. `appVersion` only ties compatibility to the `version` string, so the safety net now depends entirely on remembering to bump `version` whenever native code changes (new module, SDK bump, permission change). The comment calls this out honestly, but nothing in the diff or CI enforces it — a future PR that adds a native dep without bumping `version` would ship an incompatible OTA update silently. Given this is an urgent unblock for a broken production build, I don't think this PR needs to solve that (see Suggestions below), but it's worth having eyes open on.

- Consider a lightweight CI/release-checklist guard that flags when a mobile native dependency (`ios`/`android` config, native module) changed since the last tag but `version` in `app.config.ts` wasn't bumped — would restore some of the automatic safety `fingerprint` provided without reintroducing the EAS build-server divergence bug.

---
Looked good: single, well-scoped concern (one commit, one file), no secrets/domain-pillar issues, reasoning and root cause (expo/expo#43831) thoroughly documented in both the code comment and commit message, and this correctly targets `main` as the maintainer-cut `dev → main` release PR described in CONTRIBUTING.md's branching section rather than a contributor PR against `main`.

---

PR #698, opened 2026-07-08
fix(mobile): bump version to 1.0.1 for App Store resubmission

cner-smith:
App Store Connect rejected iOS build `1.0.0 (8)`:

- **90186** — the `1.0.0` pre-release train is closed for new submissions.
- **90062** — `CFBundleShortVersionString` must exceed the approved `1.0.0`.

Bumps the mobile marketing `version` to `1.0.1` so iOS accepts a new build. iOS production build already re-kicked at 1.0.1 (build e99812ab).

Android is untouched — its in-flight `1.0.0` build (versionCode 4) is valid for Play. The runtimeVersion (appVersion policy) moves to 1.0.1 on iOS only, which is fine per-platform.

Merge to keep `main` in sync with the shipped iOS binary.

### DIFF
--- a/apps/mobile/app.config.ts
+++ b/apps/mobile/app.config.ts
@@ -8,7 +8,7 @@ const config: ExpoConfig = {
   name: 'OGA',
   slug: 'oga',
   scheme: 'oga',
-  version: '1.0.0',
+  version: '1.0.1',
   // EAS Update (OTA). Ships JS/asset-only fixes to installed builds WITHOUT an
   // App Store / Play review — Apple/Google permit interpreted-code updates that
   // don't add native code or change the app's purpose. Native changes (SDK/RN


cner-smith (commented):
## Review — fix(mobile): bump version to 1.0.1 for App Store resubmission

**Concerns**
- `apps/mobile/app.config.ts:11` — this PR targets `main` directly from a `fix/` branch. CONTRIBUTING.md's workflow section is explicit: "Branch from `dev`, never from `main`" and "Open a PR against `dev` (not `main`) … contributors don't open PRs against `main`" (dev→main is reserved for maintainer release merges). I checked `dev` at `daf8d2f` and it's still on `version: '1.0.0'` — this fix isn't there. If `dev` diverges further before the next release-please cut, merging `dev → main` could silently revert this bump (or produce a merge/conflict surprise), since `main` will be ahead of `dev` on this field. Given this is an urgent App Store unblock tied to an already-resubmitted binary, landing directly on `main` is understandable, but it's worth also bumping `version` on `dev` (or rebasing dev onto main after this merges) so the branches don't drift.

---
Looked good: single-line, single-concern change; PR body clearly documents the App Store rejection reasons and confirms Android is intentionally untouched; no dependency, migration, or core-logic changes to worry about.

---

PR #702, opened 2026-07-08
fix(mobile): round-detail delete no longer wedges on "Deleting…"

cner-smith:
Fixes #701.

## Problem
Deleting a round from the round-detail/summary page hangs on "Deleting…" and never completes; the home-list swipe-delete works.

## Root cause
The detail handler (`app/(app)/round/[id]/index.tsx`) had no try/catch — its only exits were success→`router.replace('/(app)')` or a *returned* error→reset. If `await deleteRound` **rejects** (thrown fetch failure), neither branch runs: `deleting` stays `true` forever, no error surfaces, and the top-of-handler `if (… || deleting) return` guard then wedges the button so every subsequent tap no-ops (reads as "delete never works"). The list swipe (`rounds.tsx`) calls the same `deleteRound` but survives via `try/finally`.

## Fix
Wrap in try/catch: navigate only on confirmed success, otherwise reset `deleting` and surface the error via `Alert`. Clears the button-wedge, gives the user a retry, and reveals the reject cause if it recurs. Mirrors the working list handler.

## Testing
- `npm run typecheck` (mobile) clean.
- Structural defect + symptom confirmed by inspection; exact first-reject trigger not reproduced on-device — fix is resilience + error-surfacing, not a pinned network root cause (per repro-before-fix convention).

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -416,13 +416,16 @@ export default function RoundIndex() {
           style: 'destructive',
           onPress: async () => {
             setDeleting(true)
-            const { error: delErr } = await deleteRound(supabase, round.id, user.id)
-            if (delErr) {
+            try {
+              const { error: delErr } = await deleteRound(supabase, round.id, user.id)
+              if (delErr) throw delErr
+              router.replace('/(app)')
+            } catch (e) {
+              // Reset so the `deleting` guard can't wedge the button on retry;
+              // surface the reason instead of hanging on "Deleting…" forever.
               setDeleting(false)
-              Alert.alert('Delete failed', delErr.message)
-              return
+              Alert.alert('Delete failed', (e as Error).message)
             }
-            router.replace('/(app)')
           },
         },
       ],


cner-smith:
## Review — fix(mobile): round-detail delete no longer wedges on "Deleting…"

**Concerns**
- `apps/mobile/app/(app)/round/[id]/index.tsx:419-427` — the PR description says the exact reject trigger for `await deleteRound(...)` wasn't reproduced on-device, and attributes this to a "repro-before-fix convention," but CONTRIBUTING.md's only mention of reproduction ("Reproducible steps beat lengthy descriptions every time") is guidance for filing bug *reports*, not a documented PR convention. That's a minor mischaracterization, not a blocker — the fix is defensively correct regardless of the exact rejection cause (any thrown error, from `delErr` or a genuine promise rejection, now resets `deleting` and surfaces via `Alert` instead of wedging). Worth confirming in a follow-up if the wedge recurs, since the root cause is still technically unconfirmed.

---
Looked good: single concern (bug fix only), correctly targets `dev`, no package-boundary or security issues, and the `try { ... } catch (e) { setDeleting(false); Alert.alert(..., (e as Error).message) }` shape matches the established error-handling pattern used throughout `apps/mobile` (e.g. `bag.tsx`, `RoundTeeSelector.tsx`, `useShotActions.ts`) rather than inventing a new one.

---

PR #703, opened 2026-07-08
release: round-detail delete fix → prod

cner-smith:
Promotes `dev` to `main` (production).

## Ships
- **fix(mobile): round-detail delete no longer wedges on "Deleting…"** (#702) — wraps the delete handler in try/catch so it navigates on success, else resets `deleting` and surfaces the error instead of hanging + wedging the button.

Fixes #701.

Note: pure-JS mobile change — OTA-shippable once the aligned 1.0.1 binaries are live; until then it rides the next binary.


cner-smith:
## Review — release: round-detail delete fix → prod

**Concerns**
- `apps/mobile/app/(app)/round/[id]/index.tsx:422` — `router.replace('/(app)')` now runs inside the `try` block, after `deleteRound` has already succeeded. If navigation itself throws for any reason, execution falls into the `catch` and shows `Alert.alert('Delete failed', ...)` even though the round was already deleted — a misleading message for a delete that actually worked. Moving the `router.replace` call outside the `try` (after an early `return` from the error branch) would keep "delete failed" scoped to actual delete failures.

**Nits**
- `apps/mobile/app/(app)/round/[id]/index.tsx:427` — `(e as Error).message` assumes the caught value is Error-shaped. `delErr` (a Supabase `PostgrestError`) isn't an `Error` instance but does have `.message`, so this works today via duck typing; a non-object throw would just render `undefined` rather than crash. Not a new pattern in this PR (same cast is already used at lines 190, 376, 473), so this is pre-existing style, not something introduced here — flagging only in case it's worth a follow-up `instanceof` guard across the file.

---
Looked good: minimal, single-concern fix — wraps the actual bug (#701's unhandled rejection wedging `deleting`) in try/catch, correctly resets `deleting` and surfaces the error on failure, and doesn't touch anything outside the delete handler. Pure-JS mobile change consistent with the "no direct push to `main`, maintainers promote `dev → main`" release flow described in `CONTRIBUTING.md`.

---

Issue #278, opened 2026-05-13
Mobile: Location.requestForegroundPermissionsAsync has no timeout — hole screen can wedge


cner-smith:
**Severity:** high (UX wedge, same family as the #273 hang fix)
**Audit source:** debug-hunter agent, 2026-05-13

## Mechanism

\`apps/mobile/app/(app)/round/[id]/hole/_hooks/useHoleState.ts:108\` awaits \`Location.requestForegroundPermissionsAsync()\` with no timeout.

On Android, if the permission dialog is open and the user rotates / loses focus, the promise can hang indefinitely. The hole screen sits in \`PLACE_BALL\` with no GPS, indefinitely. The user has no signal that anything's wrong.

If the user navigates away during the hang, \`active\` is set false but the promise continues; cleanup returns before the await resolves, and \`subscription\` may or may not get installed and removed correctly.

Same hang family as #273's \`getCurrentPositionAsync\` fix. \`getCurrentPositionAsync\` was made fire-and-forget; the permission request was missed.

## Proposed fix

Wrap the permission request in a \`Promise.race\` against a 10s timeout. Identical pattern to the profile fetch in \`(app)/_layout.tsx:29-31\`.

\`\`\`ts
const perm = await Promise.race([
  Location.requestForegroundPermissionsAsync(),
  new Promise<never>((_, rej) => setTimeout(() => rej(new Error('permission timeout')), 10000)),
]).catch(() => ({ status: 'undetermined' as const }))
\`\`\`

## Files

- \`apps/mobile/app/(app)/round/[id]/hole/_hooks/useHoleState.ts:108\`

cner-smith:
Done in PR #343 — `Location.requestForegroundPermissionsAsync` is now `Promise.race`'d against a 10s timeout at both call sites (`useHoleState.ts`, `round/new.tsx`). Mechanism confirmed against native Kotlin source. Merged to `dev`; keeping open until `dev → main`.

Issue #275, opened 2026-05-13
Mobile: NaN coords from Mapbox onDragEnd silently corrupt ball state + Kalman


cner-smith:
**Severity:** critical (silent corruption → guaranteed DB write failure)
**Audit source:** debug-hunter agent, 2026-05-13

## Mechanism

\`apps/mobile/components/round/HoleMap.tsx:64-71\` \`extractCoord\` rejects non-arrays and arrays shorter than 2 elements, but writes \`{ lat, lng }\` straight to ball/aim state **without checking \`Number.isFinite\`**.

\`@rnmapbox/maps\` is known to fire \`onDragEnd\` with NaN coords on Android when a \`PointAnnotation\` is dragged off the visible map.

Cascade:
1. \`setBall({ lat: NaN, lng: NaN })\` fires
2. The \`distanceYards(ball, c) < 5\` guard in \`HoleMap.tsx:412\` returns \`false\` for NaN comparisons, so the next NaN reading isn't filtered
3. Kalman filter integrates NaN → \`kalmanStateRef.current = { lat: NaN, lng: NaN, variance: NaN }\` propagates forever
4. Every subsequent GPS update produces NaN
5. The map silently breaks for the rest of the session
6. On shot save, NaN goes to Postgres → \`invalid input syntax for type double precision\` → \`persistShot\` throws → Alert dialog → user wedged

## Proposed fix

In \`extractCoord\`, require \`Number.isFinite(lat) && Number.isFinite(lng)\` before returning. Two lines.

\`\`\`ts
if (!Number.isFinite(lat) || !Number.isFinite(lng)) return null
return { lat, lng }
\`\`\`

Same finite check should also live at \`PuttingSheet.tsx:117-123\` (filed separately) where NaN-tainted distances can flow through.

## Files

- \`apps/mobile/components/round/HoleMap.tsx:64-71\` (root site)
- Affected downstream: \`HoleMap.tsx:412\` (distance guard), Kalman path in \`_hooks/useHoleState.ts:130-138\`

cner-smith:
Done across PR #318 (initial onDragEnd guards) and PR #341 (three remaining NaN propagation guards in GPS callback, markBallHere, dropAimFromScreenPoint). Merged to `dev`; keeping open until `dev → main`.

PR #706, opened 2026-07-09
fix(mobile): source live-round GPS from Mapbox's location engine (#704)

cner-smith:
Fixes #704.

## Problem
Field-confirmed on the Android internal-test build: with permission granted and the Mapbox puck showing, `gpsPosition` stays null all session — recenter crosshairs grayed, GPS auto-center dead, ball never GPS-tracks. Consistent across cold relaunches.

## Root cause
expo-location `~18.1.6` on Android never delivers a fix to JS (all three read paths), while Mapbox's own native location engine works on the same device — matches upstream expo/expo#39851 (exact SDK 53 + 18.1.6 match, unfixed). Regression window = the SDK 53 bump; the 2026-06-01 on-course test explicitly had working GPS tracking pre-SDK-53 (#483).

## Fix
`useHoleState` now subscribes to `@rnmapbox/maps` `locationManager` — the same native engine that drives the LocationPuck, so `gpsPosition` can never disagree with the visible blue dot. expo-location stays for the permission prompt only. Kalman smoothing, manual-place freeze, and #484 revisit posture unchanged. Functional `setGpsPosition` with identity bailout keeps the engine's ~1 Hz at-rest cadence from re-rendering the hole screen. The effect's silent catch now `console.warn`s.

## Notes
- Pure JS → OTA-shippable.
- `round/new` nearest-course search (map-less) still uses expo-location and remains affected — tracked in #704's checklist along with re-testing expo-location at the SDK 54 / expo-location 19 bump.

## Testing
- Mobile typecheck clean.
- On-device validation pending (needs a build/OTA — the failing layer is native-adjacent and not reproducible in CI).

### DIFF
--- a/apps/mobile/components/round/hole/useHoleState.ts
+++ b/apps/mobile/components/round/hole/useHoleState.ts
@@ -332,15 +332,28 @@ export function useHoleState({
       active = false
       subscription?.remove()
       gpsSubscriptionRef.current = null
-      // Phase exit clears the filter so re-entry to PLACE_BALL on the
-      // next shot starts smoothing from a fresh fix rather than an
-      // old anchor that may now be hundreds of yards away. Also clears
-      // the manual-placement freeze so the next PLACE_BALL cycle
-      // resumes GPS auto-tracking unless the player drags again.
-      kalmanStateRef.current = null
-      manuallyPlacedRef.current = false
+      // Teardown ONLY — do NOT reset manuallyPlacedRef / kalmanStateRef
+      // here. This cleanup also fires on gpsNonce (every foreground
+      // return) and isRevisitingPlayedHole re-runs, where a manual ball
+      // placement must survive: resetting here let the next tick (or the
+      // addListener cache replay) snap a dragged ball back to raw GPS
+      // (#713). Genuine exits reset the refs elsewhere — the hole-change
+      // effect below, and the phase-exit effect keyed on roundState alone.
     }
   }, [currentHoleId, isPastMode, isRevisitingPlayedHole, roundState, gpsNonce])
+
+  // Phase exit (leaving PLACE_BALL: shot committed, logger opened, putting)
+  // clears the filter + manual-place freeze so the NEXT PLACE_BALL cycle
+  // starts smoothing from a fresh fix and resumes GPS auto-tracking unless
+  // the player drags again. Constraint: keyed on roundState ALONE — the
+  // subscription effect's cleanup can't own this reset because it also
+  // fires on gpsNonce/revisit re-runs, which must preserve a manual
+  // placement (#713).
+  useEffect(() => {
+    if (roundState === 'PLACE_BALL') return
+    kalmanStateRef.current = null
+    manuallyPlacedRef.current = false
+  }, [roundState])
 
   // Hole change resets per-hole state. The screen is resident (#264) so
   // nothing remounts on a hole switch — without an explicit reset the
--- a/apps/mobile/components/round/hole/useHoleState.ts
+++ b/apps/mobile/components/round/hole/useHoleState.ts
@@ -207,7 +207,7 @@ export function useHoleState({
             setTimeout(() => rej(new Error('perm-timeout')), 10_000),
           ),
         ]).catch((e: Error) => {
-          // eslint-disable-next-line no-console
+          // eslint-disable-next-line no-console -- dev-visible diagnostic for the Android permission hang (#278)
           console.warn('[useHoleState perm-timeout]', e.message)
           return { status: 'undetermined' as const }
         })
@@ -218,9 +218,11 @@ export function useHoleState({
         // getCurrentPositionAsync, watchPositionAsync) silently deliver
         // nothing on Android under SDK 53 / expo-location 18.1.6 — the
         // native fused request never starts (expo/expo#39851; field-
-        // confirmed 2026-07-08, #704). locationManager is the same native
-        // subscription that drives the LocationPuck, so it works wherever
-        // the blue dot does and gpsPosition can never disagree with it.
+        // confirmed 2026-07-08, #704). locationManager is a Mapbox-native
+        // provider like the LocationPuck's (the puck renders via the native
+        // RNMBXNativeUserLocation view — a separate subscription, not this
+        // one); both bypass expo-location's broken fused-provider path, so
+        // this works wherever the blue dot does.
         // expo-location stays for the permission prompt above only.
         // Re-test expo-location at the SDK 54 / expo-location 19 bump
         // (#704 checklist) before considering a switch back.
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — fix(mobile): source live-round GPS from Mapbox's location engine (#704)

**Concerns**
- `apps/mobile/components/round/hole/useHoleState.ts:231` + `:242-296` + `:300` — `locationManager.addListener()` replays the cached `_lastKnownLocation` synchronously into the new listener the moment it's registered (that's the library's `addListener` behavior: `if (this._lastKnownLocation) { listener(this._lastKnownLocation) }`). Since `getLastKnownLocation()` was already awaited just above (line 231), that cached point immediately runs through `onLocation` → Kalman → `setBall(smoothed)` (line 295) before any fresh fix arrives. On a fresh `PLACE_BALL` cycle (e.g. walking onto the next hole), "last known" could still be the previous hole's coordinates, so the ball marker could visibly snap to a stale location for the ~1s until a live tick corrects it. The old code only fed `gpsPosition` from the last-known fix, never `ball`/Kalman — this is a new path into the ball marker that wasn't there before.
- `apps/mobile/components/round/hole/useHoleState.ts:221-223` — the comment states locationManager "is the same native subscription that drives the LocationPuck." Per `@rnmapbox/maps` source, `LocationPuck` renders through the native `RNMBXNativeUserLocation` view and doesn't go through the JS `locationManager` singleton at all — they're two independent native location providers that happen to both work on Android (likely because both bypass expo-location's broken fused-provider path), not the same subscription. The stronger claim two lines down ("gpsPosition can never disagree with the visible blue dot") doesn't follow from that and may mislead whoever debugs this next if the two ever do disagree.
- `apps/mobile/components/round/hole/useHoleState.ts:297-303` — no `locationManager.setMinDisplacement()` call before `addListener()`. Per the library's `start()`, an unset displacement defaults to `-1` (no distance filter), matching this PR's own "~1 Hz continuously" comment. The removed expo-location config was deliberately throttled (`distanceInterval: 5`, `timeInterval: 5000`) with inline reasoning about battery tradeoffs. `setGpsPosition` got an identity bailout to absorb the higher tick rate, but `setBall`/Kalman (line 288-295) did not — it now re-renders the ball marker roughly every ~1s whenever GPS-driven tracking is active, not just every ~3.5s while walking / 5s at rest as before. Worth either restoring an equivalent `setMinDisplacement(5)` or a note that the tradeoff is intentionally being dropped.

**Nits**
- `apps/mobile/components/round/hole/useHoleState.ts:306` — `// eslint-disable-next-line no-console` has no inline reason, which CONTRIBUTING.md calls for ("No `// eslint-disable-next-line` without a comment explaining why"). Mirrors the pre-existing instance at line 210 in the same file, so not new to this PR, but worth tightening both while here.

---
Looked good: CI is fully green (typecheck, test, mobile-typecheck, web-build, check-vendor all pass); scope stays to the one stated concern (GPS source swap, no unrelated drive-by changes); the #484 revisit posture and the #275 NaN guard are both carried over unchanged; and the async cleanup race the old code guarded against (`if (!active) { subscription.remove(); return }`) is correctly recognized as unnecessary now since `addListener`/`removeListener` are synchronous.

---

cner-smith (commented):
## Review — fix(mobile): source live-round GPS from Mapbox's location engine (#704)

**Concerns**
- `apps/mobile/components/round/hole/useHoleState.ts:273-277` — the `setGpsPosition` identity bailout is meant to stop "at-rest ticks from re-rendering the hole screen," but it compares raw GPS floats (`rawPoint.lat === prev.lat && rawPoint.lng === prev.lng`) for exact equality. I pulled the actual `locationManager` source from `@rnmapbox/maps@10.2.10` (pinned in `apps/mobile/package.json`) to confirm its update cadence — real device GPS readings jitter at the sub-meter level even at rest, so back-to-back fixes from its ~1 Hz stream will essentially never be bit-identical. The bailout will rarely fire, so the render-thrash problem the comment claims to solve is likely still present. Worse, `setBall` via the Kalman filter (line 295) runs unconditionally on every tick regardless of this bailout — there's no equivalent to the old `watchPositionAsync`'s `distanceInterval`/`timeInterval` throttle. Net effect: the hole screen may now re-render continuously at ~1 Hz while the player is standing still, which is the opposite of what the comment states. Consider an epsilon/rounded comparison, or a real distance/time gate (`locationManager.setMinDisplacement()` exists on the singleton) if the at-rest render cost matters.
- On-device validation is explicitly pending per the PR description ("needs a build/OTA... not reproducible in CI"). This exact code path has regressed in the field multiple times already (#275, #484, #640, #278). Given it's OTA-shippable, worth confirming on a real Android device before merge rather than after — a bad GPS regression here is user-facing on every live round.

**Nits**
- `apps/mobile/components/round/hole/useHoleState.ts:371-378` — the component's own `AppState` listener now duplicates lifecycle handling `locationManager` already does internally (its `_handleAppStateChange` stops on `background` and restarts `start()` on `active` when listeners are registered). Not wrong, just redundant now that the source is Mapbox's own manager rather than expo-location; could likely be dropped in a later pass.

---
Looked good: the root cause is well-diagnosed and cross-referenced against the actual upstream issue (expo/expo#39851), the `expo-location` → `locationManager` swap is a clean single-concern change with the permission-request/timeout logic left untouched, and the NaN guard, manual-placement freeze, and #484 review posture are all preserved unchanged.

---

cner-smith:
Review findings verified against `@rnmapbox/maps` source — all three confirmed real and fixed in 1b17d4b:

1. **Stale cache replay → ball/Kalman**: confirmed `addListener` replays `_lastKnownLocation` synchronously. Added a >10 s staleness guard after `setGpsPosition` — stale fixes update the GPS chip only, never the ball/Kalman path (restores the old last-known semantics).
2. **Missing displacement filter**: confirmed `start(-1)` = unfiltered. `setMinDisplacement(5)` restores the old 5 m gating; comment documents the at-rest tradeoff (no ticks, held position stays correct).
3. **LocationPuck comment overclaim**: confirmed the puck is `RNMBXNativeUserLocation` (native view), not the JS `locationManager`. Comment reworded — independent providers that both bypass expo-location's broken path.

Nit: both `eslint-disable no-console` lines (new + pre-existing perm-timeout) now carry reasons.

cner-smith:
Scope grew by two hardening commits on the same GPS effect this PR rewrote:

- **f9f8fe6** — the subscription effect's cleanup no longer resets `manuallyPlacedRef`/`kalmanStateRef` (it fires on every foreground-return `gpsNonce` bump, not just phase exits, so any app-state blip snapped a dragged ball back to raw GPS). Cleanup is teardown-only now; genuine exits are covered by the existing hole-change reset plus a new phase-exit effect keyed on `roundState` alone, preserving the leave-PLACE_BALL unfreeze for the next shot. Closes #713.
- **a92cb5d** — #720 items 1, 3, 4 (item 2 was fixed in PR #724): bail before `locationManager.addListener` if the app backgrounded during the setup awaits (addListener unconditionally start()s the engine); new `gpsFixAtRef` records each fix's own timestamp so `markBallHere` declines a gpsPosition-only fallback older than 30 s via the existing 'No GPS yet' alert; the `getLastKnownLocation` seed gets the same `Number.isFinite` guard as the listener path. Closes #720.

Typecheck green; needs the same on-device pass as the rest of this PR.

PR #722, opened 2026-07-09
fix(mobile): stamp distance_to_target on live-round shot saves

cner-smith:
Mobile live-round shot saves never wrote `distance_to_target` on the `shots` row, so `getExpectedStrokes(category, undefined)` returned null and the SG engine skipped scoring entirely — stamping `sg_off_tee`/`sg_approach`/`sg_around_green` at exactly 0.00 for every live round at completion. `buildPayload` in `useShotActions.ts` now computes it the same way `PastRoundMap.tsx` does: rounded yards from the shot-start (`ball`) to the resolved pin (`roundPin ?? storedPin`), for non-putt shots only. Putts keep `distance_to_target` null (parity with the web save path — a putt's distance lives in `putt_distance_ft`).

How verified: `npm run typecheck` in `apps/mobile` passes clean. Runtime/device verification (confirming SG values populate correctly on a live round at completion) is still pending.

Closes #709

### DIFF
--- a/apps/mobile/components/round/hole/useShotActions.ts
+++ b/apps/mobile/components/round/hole/useShotActions.ts
@@ -131,6 +131,8 @@ export function useShotActions(input: UseShotActionsInput): UseShotActionsResult
 
   function buildPayload(meta: ShotLoggerValue | null): ShotPayload | null {
     if (!user || !currentHoleScore || !ball) return null
+    const isPutt = meta?.club === 'putter' || meta?.lieType === 'green'
+    const pinTarget = roundPin ?? storedPin ?? null
     return {
       hole_score_id: currentHoleScore.id,
       user_id: user.id,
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(mobile): stamp distance_to_target on live-round shot saves

**Concerns**
- `apps/mobile/components/round/hole/useShotActions.ts:134` — `isPutt` is re-derived inline (`meta?.club === 'putter' || meta?.lieType === 'green'`) instead of calling the existing `isPuttShot(lieType, club)` from `@oga/core` (`packages/core/src/round.ts:89`). That helper's own comment is explicit about why it exists: "Single source of truth for putt classification... every surface must agree... which was a drift between hand-copied predicates (#660)." This is currently logically identical to `isPuttShot`, but it's now the second hand-copied instance of the same predicate in this file (the first is `persistShot`, line 197), which is the exact pattern #660 warns against. CONTRIBUTING.md's package-boundary rule points the same way ("Don't add helpers to apps/web/src/lib or apps/mobile/lib that other surfaces would want to reuse — lift them to @oga/core"). Suggest `isPuttShot(meta?.lieType ?? null, meta?.club ?? null)` here instead.

- While touching this area, `persistShot`'s pre-existing hand-copied `isPutt` (useShotActions.ts:197) could also move to `isPuttShot` for the same reason — a separate small chore, not required for this fix.

---
Looked good: the root-cause diagnosis is well-evidenced (`getExpectedStrokes(category, undefined)` returning null and silently zeroing SG for every live round); the fix correctly excludes putts from `distance_to_target`, matching the shape the web save path already uses (`apps/web/src/pages/rounds/hooks/useRoundActions.ts:667`, `isPuttRow ? null : Math.round(...)`); scope stays to the one stated concern; PR targets `dev`; no `@oga/core` changes so no new-export test gap.

---

PR #723, opened 2026-07-09
fix: completeRound no longer finalizes totals/differential from unplayed score-0 holes (#711, #717)

cner-smith:
## Mechanism

Mobile round creation pre-inserts a `hole_scores` row with `score: 0` for every hole, and per-shot score stamping is a direct online write that is never queued — so offline stretches, early End-round, or unlogged holes leave score-0 rows that `completeRound` trusted everywhere: totals, `holedOut` GIR inference, and the WHS differential. `adjustedScore` adds 0 strokes per unplayed hole, producing impossible negative differentials (≈ −23..−65) that permanently dominate the last-20 handicap recompute; past the profiles CHECK floor the profile write then threw **after** the round had already finalized ("End round failed"). Separately, mobile's no-handicap fallback was a hardcoded `18` while web uses `DEFAULT_HANDICAP` (15) — different SG brackets per platform (#717).

## Changes

**`apps/mobile/lib/completeRound.ts`**
- Self-repair stale stamped zeros: a score-0 hole that HAS synced shots is patched in place to `score = shot count` (the exact value live stamping would have written online) before totals/SG/holedOut/differential read it; the existing SG + infer upserts persist the repaired value server-side.
- Differential is computed only when played rows (effective score > 0) cover the round's full hole count (`inferHoleCount` over the course's hole numbers); otherwise `score_differential` stays null — never partially computed.
- Both `?? 18` handicap fallbacks replaced with `DEFAULT_HANDICAP` from `@oga/core` (#717).
- Handicap-index recompute is now non-fatal: the round is already finalized at that point, so a diffs-fetch or profile-update error DEV-warns and continues instead of surfacing "End round failed".

**`apps/web/src/hooks/useCompleteRound.ts`**
- Same differential guard: filter to score > 0 rows and require full coverage of `inferHoleCount(holes)` before computing; null otherwise. (Web creates `hole_scores` lazily, so the partial-round arm shows up as missing rows rather than score-0 rows — the coverage requirement handles both.)

## How verified

- `cd apps/mobile && npm run typecheck` — clean
- `pnpm typecheck` (workspace: core, web) — clean
- No `@oga/core` changes, so no new tests. Device/runtime QA pending (offline stretch + early End-round repro).

Closes #711
Closes #717

### DIFF
--- a/apps/mobile/lib/completeRound.ts
+++ b/apps/mobile/lib/completeRound.ts
@@ -1,8 +1,10 @@
 import {
+  DEFAULT_HANDICAP,
   adjustedScore,
   calculateDifferential,
   calculateHandicapIndex,
   computeRoundSG,
+  inferHoleCount,
   inferHoleStats,
 } from '@oga/core'
 import {
@@ -78,6 +80,21 @@ export async function completeRound({
   const tees: CourseTeeRow[] = teesRes.data ?? []
   const roundTee = roundRes.data ?? null
 
+  // Self-repair stale stamped zeros (#711). Round creation pre-inserts
+  // every hole_scores row with score 0; live play updates it per shot via
+  // a direct online write that is never queued, so an offline stretch
+  // loses the score even though the shots themselves sync later. Live
+  // stamping defines score = shot count, so a score-0 hole that HAS
+  // synced shots recovers exactly the value the online path would have
+  // written. Patched in place so every downstream reader (computeRoundSG
+  // totals, holedOut, the SG upsert's carried score, the differential)
+  // sees the repaired value; the SG/infer upserts below persist it.
+  for (const hs of holeScores) {
+    if (hs.score !== 0) continue
+    const shotCount = shots.filter((s) => s.hole_score_id === hs.id).length
+    if (shotCount > 0) hs.score = shotCount
+  }
+
   // Infer fairway_hit + gir for any hole where the player didn't set
   // them manually. Mobile live mode never writes those columns
   // per-shot, so without this round-level fairwaysHit/gir totals stay
@@ -135,7 +152,7 @@ export async function completeRound({
     holes,
     holeScores,
     shots,
-    handicap: handicap ?? 18,
+    handicap: handicap ?? DEFAULT_HANDICAP,
   })
 
   // Per-hole SG upsert. Mirrors useCompleteRound.ts: carry round_id /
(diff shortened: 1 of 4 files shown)


cner-smith:
## Review — fix: completeRound no longer finalizes totals/differential from unplayed score-0 holes (#711, #717)

**Concerns**
- `apps/mobile/lib/completeRound.ts:218-221` and `apps/web/src/hooks/useCompleteRound.ts:139-142` — the completeness gate (`const playedRows = holeRows.filter((r) => r.score > 0); const holeCount = inferHoleCount(...); if (playedRows.length === holeCount)`) is hand-copied verbatim (module the handicap-fallback arg) into both apps instead of living in `@oga/core`. CONTRIBUTING.md's package-boundary rule is explicit: "Don't add helpers to `apps/web/src/lib` or `apps/mobile/lib` that other surfaces would want to reuse — lift them to `@oga/core`." This is the same shape of drift `isPuttShot`'s comment in `packages/core/src/round.ts:82-88` warns about (#660: "drift between hand-copied predicates"). A shared `hasFullHoleCoverage(rows, holeCount)` (or similar) next to `inferHoleCount` in `packages/core/src/round.ts`, with a test, would remove the duplication and the risk of the two arms silently diverging later.
- `apps/mobile/lib/completeRound.ts:219` / `apps/web/src/hooks/useCompleteRound.ts:140` — `inferHoleCount` is being reused here for a purpose its docstring (`packages/core/src/round.ts:67-79`) doesn't describe: bucketing a course's *own* mapped hole numbers to a canonical 9/18, explicitly to handle "interior gaps (holes 1–4, 7–18 → still an 18)." Using that same bucketed value as the "did the player finish every hole" target only works because round creation (`apps/mobile/app/(app)/round/new.tsx:228-260`) and web's `ensureRealHole` materialize real `holes` rows up to that same inferred count as each hole is scored. I traced through both paths and the invariant seems to hold for a genuinely complete round, but it's an implicit coupling between three separate call sites (round creation, ensureRealHole, and this gate) that isn't documented anywhere as a shared contract — if a future change to either materialization path changes when/whether synthetic holes get created, this gate would silently start leaving `score_differential` null on fully-played rounds with no error surfaced. Might be worth a one-line comment noting the dependency, or asserting `holes.length >= holeCount` isn't required for correctness so much as for it to keep being true.

**Nits**
- This PR bundles two issues (#711 and #717) rather than one. CONTRIBUTING.md asks for one concern per PR. In this case both fixes land on the same lines of the same function (`handicap ?? 18` → `handicap ?? DEFAULT_HANDICAP` sits inside the block already being rewritten for the differential guard), so splitting them would likely just create merge friction for no real benefit — flagging for awareness rather than asking for a split.

---
Looked good: root-cause tracing is thorough and cites the actual failure mode (negative differentials from `adjustedScore` over 0-strokes-per-unplayed-hole, then the CHECK-floor throw happening *after* the round had already finalized); the non-fatal handling of the post-finalize handicap-index recompute is the right call since the round write already succeeded by that point; the self-repair loop correctly recovers a value equal to what live stamping would have written and is the same source live GIR inference already trusts; CI is fully green; no new `@oga/core` exports so no test-coverage gap; PR targets `dev`.

---

PR #724, opened 2026-07-09
fix(mobile): hole shot-count integrity — fetch-error guard, sync dedupe, reset-on-switch

cner-smith:
## What

Three fixes to the counts-load effect in `useHoleData.ts` (~line 179), all inside the same `[currentHoleScore?.id]` effect:

**1. Fetch-error guard (#712).** `shotsRes.error` was never inspected — a transient failure resolved the promise normally (postgrest-js returns failures in-band as `{data: null, error}`), so `const shots = shotsRes.data ?? []` silently zeroed `remoteShotCount`. The next saved shot then computed `shot_number = 1`, collided with the existing shot 1 on `unique(hole_score_id, shot_number)`, and got quarantined `broken` by `lib/sync.ts` — invisible to every read including the end-round unsynced-shots warning.

**Design choice:** retry once (mirrors the existing chunk-upsert retry pattern in `lib/sync.ts:153-160`), then on persistent failure, DEV-warn and leave counts at the neutral reset state (see #3) rather than committing anything. I considered adding a `countsReliable` gate that `persistShot` could check before saving, but that would touch `useShotActions.ts` and the save UI — outside this file and larger than the fix warrants. The retry closes the common transient case; the residual "still failing after retry" window is the same class of problem the issue itself flags as longer-term hardening (server-side shot_number derivation). Noting this explicitly since it's a real, if narrow, trade-off.

**2. Sync-window dedupe (#714).** A shot mid-sync could count in both `remoteShotCount` (server row committed) and `pendingForHole` (not yet `markShotSynced`'d), inflating `shotNumber` and eventually `hole_scores.score` / `rounds.total_score`. Added `id` to the remote `shots` select and now filter pending rows whose `payload.id` (client-generated, always set before insert per the sync idempotency work in `lib/sync.ts`) appears in the remote result.

**3. Reset-on-hole-switch (Part of #720, item 2).** Counts were never reset when `currentHoleScore.id` changed, so during the refetch RTT, `hasPriorShots` (`LiveRoundSession.tsx:143`, `remoteShotCount + localShotCount > 0`) held the *previous* hole's answer — a swipe back from a shotless hole could let the tee-default effect place a ball marker on a hole the player was only reviewing. Now `remoteShotCount`/`remotePuttCount`/`remoteShotStarts`/`pendingForHole` reset to 0/[] synchronously at the top of the effect, before the async fetch — this also matches the true pre-fetch state of a fresh hole, so it's not a behavior change for the common path.

Also wrapped the async IIFE body in try/catch so a SQLite (`pendingShotsForHoleScore`) rejection warns (DEV-gated) and leaves counts at the neutral reset instead of stranding the previous hole's values.

## How verified

- `cd apps/mobile && npm run typecheck` — clean.
- Runtime verification pending (needs a device/simulator repro of a transient fetch failure or mid-sync navigation, per CLAUDE.md's "repro before fix" — these are audit-filed, mechanism-verified-in-code bugs; no live repro attempted this session).

Closes #712
Closes #714
Part of #720 (item 2)

### DIFF
--- a/apps/mobile/components/round/hole/useHoleData.ts
+++ b/apps/mobile/components/round/hole/useHoleData.ts
@@ -198,11 +198,28 @@ export function useHoleData(
             .select('id, club, lie_type, shot_number, start_lat, start_lng')
             .eq('hole_score_id', currentHoleScore.id)
             .order('shot_number')
-        const [shotsResInitial, local] = await Promise.all([
+        const [shotsResInitial, localInitial] = await Promise.all([
           fetchShots(),
-          pendingShotsForHoleScore(currentHoleScore.id),
+          // The SQLite read gets the same one-retry treatment as the remote
+          // fetch below — a transient rejection must not leave the neutral
+          // reset committed as a confident local=0 the save path could
+          // collide on (same class as the remote arm of #712).
+          pendingShotsForHoleScore(currentHoleScore.id).catch(() => null),
         ])
         if (myNonce !== fetchNonceRef.current) return
+        let local = localInitial
+        if (local === null) {
+          local = await pendingShotsForHoleScore(currentHoleScore.id).catch(
+            () => null,
+          )
+          if (myNonce !== fetchNonceRef.current) return
+          if (local === null) {
+            if (__DEV__) {
+              console.warn('[hole/pending-fetch] SQLite read failed twice')
+            }
+            return
+          }
+        }
         let shotsRes = shotsResInitial
         if (shotsRes.error) {
           // postgrest-js returns failures in-band ({data: null, error}) —
--- a/apps/mobile/components/round/hole/useHoleData.ts
+++ b/apps/mobile/components/round/hole/useHoleData.ts
@@ -177,31 +177,89 @@ export function useHoleData(
   // resolve after a newer run started are silently dropped. (#284)
   const fetchNonceRef = useRef(0)
   useEffect(() => {
+    // Reset to the neutral "unknown" state synchronously, before the async
+    // fetch below resolves. Without this, a hole switch during the fetch
+    // RTT leaves the PREVIOUS hole's counts in state — e.g. `hasPriorShots`
+    // (LiveRoundSession) reads a stale true/false and the tee-default effect
+    // can place a ball marker on a hole the player is only reviewing (#720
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(mobile): hole shot-count integrity — fetch-error guard, sync dedupe, reset-on-switch

**Concerns**
- `apps/mobile/components/round/hole/useHoleData.ts:212-283` — the retry-and-fallback added for the remote `shots` fetch has no equivalent for the local `pendingShotsForHoleScore(currentHoleScore.id)` read. Both calls sit in the same `Promise.all` (line 199-202), so if the SQLite read throws, the whole `try` block aborts into the `catch` at line 277 before any of the already-successful remote data is used, and `pendingForHole` stays at the neutral `[]` reset from line 189. That reset feeds directly into `shotNumber = remoteShotCount + localShotCount + 1` (this file, outside the diff, consumed by `useShotActions.ts:137` as `shot_number: shotNumber`) — so a transient SQLite read failure produces the same class of bug this PR's own #712 fix targets: a new shot gets saved with a `shot_number` that's already claimed by an un-synced local pending shot, colliding on `unique(hole_score_id, shot_number)`. It's a narrower window than #712 (local SQLite vs. network), but the PR added a retry specifically because a first-attempt failure shouldn't be trusted as "confidently zero" for the remote side — the same reasoning seems to apply to the local side, which currently has none.

**Nits**
- This PR bundles three closed issues (#712, #714, part of #720). CONTRIBUTING.md's "one concern per PR" rule would normally call for a split, but all three touch the same replaced lines of the same effect, so splitting would mostly create merge friction rather than real review value — flagging for awareness, not asking for a split (same call made on PR #723's differential/handicap bundle).

---
Looked good: the root-cause tracing for both #712 (in-band `{data: null, error}` never checked) and #714 (sync-window double count) is precise and each is independently verified against the existing `lib/sync.ts` idempotency/retry patterns it says it mirrors — `payload.id` is in fact always set before insert (`lib/db.ts:61`), so the dedupe's `!payload.id` fallback is genuinely defensive, not silently wrong; the nonce-gated reset-on-hole-switch fix is a real correctness improvement over the prior stale-previous-hole behavior; CI is fully green; PR targets `dev`; no `@oga/core` changes so no test-coverage gap.

---

PR #725, opened 2026-07-09
fix(db): per-round par override + authorized tee RPC — hole writes silently no-op'd with no UPDATE policy (#710, #721)

cner-smith:
## Mechanism

`holes` has RLS enabled with only SELECT + INSERT policies — no UPDATE policy exists in any migration 0001–0039 (verified against live prod + dev `pg_policy`). An UPDATE under RLS with zero policies filters to 0 rows, and PostgREST reports success with `error: null`. All three client call sites checked only `error`, so par corrections and auto-captured tee coordinates **silently never persisted for any user**, and every rollback/warn branch was structurally unreachable.

## Design (per maintainer decision on the #710 open question)

- **Par is a per-round override, not global crowd curation** — new nullable `hole_scores.par` (null = use the course hole's par). `hole_scores` already carries the owner-scoped FOR ALL policy from 0001, so no policy work needed.
- **Tee auto-capture stays global** (GPS tee coords are factual data) — writes go through a narrow SECURITY DEFINER RPC.

## Migration `0040_round_par_override_and_tee_rpc.sql`

- `alter table hole_scores add column par smallint check (par between 3 and 6)` (nullable).
- New `update_hole_tee(p_hole_id, p_round_id, p_tee_lat, p_tee_lng)` — SECURITY DEFINER, `set search_path = public`, auth check, authorization = caller owns a round whose `course_id` matches the hole's course; updates only `tee_lat`/`tee_lng`. `revoke all from public` + `grant execute to authenticated`.
- **#721:** `create or replace` of `insert_synthetic_hole` adding the one-line `and r.course_id = p_course_id` predicate its own comment always claimed — body otherwise copied verbatim from 0026.

## `types.ts` hand-edit

Added `par` to the `hole_scores` Row/Insert/Update types and an `update_hole_tee` Functions entry in `packages/supabase/src/types.ts`. Types are normally generated — these entries will be reconciled at the next `gen types` run after the migration is applied.

## Call-site changes

- **web `HoleScoreCard.tsx` (setPar):** par now persists through the component's existing `hole_scores` upsert path (`ensureRealHole` + `useUpsertHoleScore`). `effectivePar` chain is `parOverride ?? holeScore?.par ?? hole.par`. Because `hole_scores.score` is NOT NULL, a par tap before any score stays a local override and `persist()` carries it with the first score save. On failure: rolls back the optimistic override AND surfaces via the card's existing `saveError` affordance.
- **mobile `LiveRoundSession.tsx` (onChangePar):** writes `hole_scores.par` by the hole's `hole_scores` row id (rows are batch-created at round start), chaining `.select('id')` and **treating 0 returned rows as failure** — the #710 lesson. The previously-unreachable optimistic-rollback branch is now reachable and rolls back `holeScores` state.
- **mobile `useShotActions.ts` (writeTee):** `.update()` → `update_hole_tee` RPC. Keeps its fire-and-forget background-enrichment character — real errors `console.warn`, never user-blocking.

## Effective-par reads (`hs.par ?? hole.par`, inline — no helper)

- **Completion boundary (both apps):** `apps/mobile/lib/completeRound.ts` and `apps/web/src/hooks/useCompleteRound.ts` patch the fetched `holes` array with per-round overrides before `computeRoundSG` / `inferHoleStats` / the WHS differential — one boundary covers SG categorization, stat inference, and to-par stamped at round completion.
- **Mobile displays:** live HUD par (LiveRoundSession ×2), `Scorecard.tsx` modal (display, running par, diff, cycle base, a11y labels) + `ScorecardPreview` diff, past-round detail scorecard (`round/[id]/index.tsx`: running totals, per-hole diff, FIR/GIR inference, par cell), `ShareableScorecardCard` (total par, grid par sums, par row, score-cell par).
- **Web displays:** `RoundSummary` (total par + best/worst hole diffs), `ScorecardView` (holePar passed to the shot editor), `RoundDetailPage` `HoleReviewSheet` (par + totalPar — `scoresByHoleId` already in scope).

**Deliberately left reading base par** (out of scope, no hole_scores fetched alongside or intentionally course-level): `@oga/core` stats aggregates, courses pages, web `MapView` hole-marker titles, web `useRoundData` internal shot-categorization pars, mobile `PastRoundMap` HUD par, `round/new.tsx` (round creation).

## How verified

- Root `pnpm typecheck` and `apps/mobile` `npm run typecheck` both pass.
- **Migration NOT applied to any live database** — needs `npx supabase db push` on dev, then prod, after merge.

Closes #710
Closes #721

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -545,7 +545,9 @@ export default function RoundIndex() {
     const hs = scoresByHoleId.get(h.id)
     if (hs?.score != null && hs.score > 0) {
       runningScore += hs.score
-      runningPar += h.par
+      // Per-round par override (#710) — hole_scores.par wins over the
+      // course hole's par when the player corrected it.
+      runningPar += hs.par ?? h.par
     }
   }
   const diff = runningScore - runningPar
@@ -796,14 +798,17 @@ export default function RoundIndex() {
             const hs = scoresByHoleId.get(h.id)
             const score = hs?.score ?? 0
             const putts = hs?.putts ?? null
-            const d = score > 0 ? score - h.par : null
+            // Per-round par override (#710) — hole_scores.par wins over
+            // the course hole's par when the player corrected it.
+            const par = hs?.par ?? h.par
+            const d = score > 0 ? score - par : null
             const shotCount = hs ? holeScoreShotCount.get(hs.id) ?? 0 : 0
             // FIR/GIR: prefer the persisted hole_scores columns — web parity,
             // a manual/web-set value wins (apps/web useRoundActions) — else
             // infer from the hole's placed shots. par-3 fairway → null (blank).
             const inferred = inferHoleStats(
               hs ? shotsByHoleScoreId.get(hs.id) ?? [] : [],
-              h.par,
+              par,
             )
             const fairway = hs?.fairway_hit ?? inferred.fairway
             const gir = hs?.gir ?? inferred.gir
@@ -838,7 +843,7 @@ export default function RoundIndex() {
                     fontVariant: ['tabular-nums'],
                   }]}
                 >
-                  {h.par}
+                  {par}
                 </Text>
                 <ScoreCell
                   value={score}
--- a/apps/mobile/components/round/LiveRoundSession.tsx
+++ b/apps/mobile/components/round/LiveRoundSession.tsx
@@ -661,16 +661,21 @@ export default function LiveRoundSession({
         routerReplace={(href) => router.replace(href as Parameters<typeof router.replace>[0])}
         id={roundId}
         onChangePar={async (holeId, newPar) => {
+          if (!roundId) return
           // Optimistic update so the cell reflects the tap immediately.
           // Roll back if the DB write fails so the UI doesn't lie.
+          // `holes` has no UPDATE RLS policy — a direct .update() filters
+          // to 0 rows and reports success (#710) — so the write goes
+          // through the authorized RPC, scoped to this round.
           const prev = data.holes.find((h) => h.id === holeId)?.par ?? 4
           data.setHoles((cur) =>
             cur.map((h) => (h.id === holeId ? { ...h, par: newPar } : h)),
           )
-          const { error: parErr } = await supabase
-            .from('holes')
-            .update({ par: newPar })
-            .eq('id', holeId)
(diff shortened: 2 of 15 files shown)


cner-smith:
## Review — fix(db): hole par/tee curation writes go through an authorized RPC (#710, #721)

**Concerns**
- `supabase/migrations/0040_hole_curation_rpc.sql:34-38` — `p_par` is bounds-checked (3–6) before the update, but `p_tee_lat`/`p_tee_lng` have no validation at all — no `-90..90`/`-180..180` range check, not even a `Number.isFinite`-equivalent. `holes` is shared crowd-curated data (not user-owned), and the authorization model here is intentionally loose by design — anyone who has ever started a round on a course can call this RPC for any hole on that course (matching the existing `insert_synthetic_hole` pattern). The client-side guard (`useShotActions.ts:317`, `Number.isFinite`) doesn't reach a caller hitting the RPC directly, and the function's own comment on the par check ("so a malformed payload returns a clean error instead of tripping the constraint deep in the function") shows the validation gap for lat/lng wasn't a deliberate omission so much as a gap. Garbage coordinates written here corrupt tee position for every future player at that course, not just the writer. Worth adding the same style of range check used for par.

**Nits**
- This PR bundles two closed issues (#710 and #721) into one migration. Both are real security fixes to the same crowd-curation subsystem and #721 is a two-line addition to a function already being touched conceptually (not literally the same lines as #710, but same file/domain) — flagging for awareness per CONTRIBUTING.md's one-concern-per-PR rule, not asking for a split, since the two are tightly coupled (both are RPC-authorization hardening for hole curation) and splitting would mean re-reviewing the same migration file twice.
- CI was still queued at review time (checks hadn't completed) — the PR is otherwise ready modulo that.

---
Looked good: root cause is well-evidenced (RLS with zero UPDATE policies silently zero-row-filters and PostgREST reports success, confirmed against the description of `pg_policy` state); the new RPC's authorization check correctly joins `rounds`/`holes` on `course_id` and requires `rounds.user_id = auth.uid()`, matching the existing `insert_synthetic_hole` ownership shape; `revoke all from public` + `grant to authenticated` correctly locks out `anon`; the `#721` fix for `insert_synthetic_hole` (adding the missing `r.course_id = p_course_id` predicate) is verified correct against migration `0026`'s original body; all three call sites now handle the error path where they previously couldn't (web's `ensureRealHole` rejection is now caught, mobile's rollback branches are now reachable); migration number `0040` is the correct next-free sequential number; no `@oga/core` changes so no test-coverage gap; PR targets `dev`.

---

PR #726, opened 2026-07-09
fix: display batch — 9-hole share card, sentinel score renders, hole clamp, styledata listener

cner-smith:
- **#715** — Web share card and round header hardcoded 18 holes. `ShareableScorecardCard.tsx` now uses `inferHoleCount` (mirroring the mobile card) to size `holeNumbers`/`totalPar` and only renders the back-nine grid when `holeCount > 9`. `RoundDetailPage.tsx` computes `holeCount` from the round's `holes` array and passes it to `RoundHeader.tsx`, which now shows `{holesPlayed}/{holeCount} holes scored` instead of the hardcoded `/18`.
- **#716** — Mobile home `avgScore` counted `total_score = 0` sentinel rounds (abandoned past-round logs), skewing the average indefinitely; `bestScore` already excluded them. Unified both into a single `realScoreRounds` filter (`total_score != null && total_score > 0`) in `app/(app)/index.tsx`. Also fixed three literal-`0` score renders that `?? '—'` didn't catch: `app/(app)/rounds.tsx`, `components/home/RecentRoundsList.tsx`, and web's `pages/rounds/RoundsPage.tsx`.
- **#718** — Deep-linking `round/[id]?hole=10+` on a 9-hole round hit the "Hole N isn't set up" dead-end. Added a clamp effect in `LiveRoundSession.tsx` right after `data = useHoleData(...)`: once `data.loading` is false, if `holeNumber > data.holeCount` it clamps `holeNumber` down to `data.holeCount`, mirroring the existing `useActiveRound` resume-banner clamp pattern.
- **#719** — `useMapLayers.ts`'s `styledata` wait-for-style branch registered `map.once('styledata', ...)` with no teardown, so repeated effect re-runs during the load window could stack stale-closure listeners that redraw old markers. Now uses a named handler and returns `() => map.off('styledata', handler)`.

**How verified:** `pnpm typecheck` (root) and `cd apps/mobile && npm run typecheck` both pass clean. Visual verification (9-hole share card render, avg-score display, hole-clamp deep link, map marker drag) is pending — not yet manually tested in-app.

Closes #715
Closes #716
Closes #718
Closes #719

### DIFF
--- a/apps/mobile/app/(app)/index.tsx
+++ b/apps/mobile/app/(app)/index.tsx
@@ -142,17 +142,20 @@ export default function Home() {
       const avg = values.length === 0 ? 0 : values.reduce((a, b) => a + b, 0) / values.length
       return { ...c, value: Number(avg.toFixed(2)) }
     })
-    const scoreRounds = rounds.filter((r) => r.total_score !== null)
-    const avgScore = scoreRounds.length > 0
-      ? scoreRounds.reduce((s, r) => s + (r.total_score ?? 0), 0) / scoreRounds.length
-      : null
     // total_score === 0 is the past-round-logger sentinel for "no score
-    // entered" (map-created rounds), so exclude it from the best-round min —
-    // otherwise a single unscored round always reads as a best of 0.
-    const realScores = rounds
-      .map((r) => r.total_score)
-      .filter((s): s is number => s != null && s > 0)
-    const bestScore = realScores.length > 0 ? Math.min(...realScores) : null
+    // entered" (map-created rounds), so exclude it from avg + best-round min —
+    // otherwise an abandoned log skews the average and a single unscored
+    // round always reads as a best of 0.
+    const realScoreRounds = rounds.filter(
+      (r): r is typeof r & { total_score: number } =>
+        r.total_score != null && r.total_score > 0,
+    )
+    const avgScore = realScoreRounds.length > 0
+      ? realScoreRounds.reduce((s, r) => s + r.total_score, 0) / realScoreRounds.length
+      : null
+    const bestScore = realScoreRounds.length > 0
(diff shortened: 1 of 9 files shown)


cner-smith:
## Review — fix: display batch — 9-hole share card, sentinel score renders, hole clamp, styledata listener

**Concerns**
- PR-level — this bundles four independent fixes across four unrelated subsystems (web share card sizing, mobile home avg/best-score math, mobile deep-link hole clamp, web map listener leak). CONTRIBUTING.md's "One concern per PR" rule says "PRs that add scope beyond their stated concern will be asked to split" — these don't share a root cause or touch overlapping code, they're grouped only by being display-adjacent audit findings. Worth splitting into 4 PRs so each can be reviewed/reverted independently (e.g. if the map listener fix needs a follow-up, it shouldn't block the share-card fix from landing).

**Nits**
- `apps/mobile/app/(app)/rounds.tsx:206`, `apps/mobile/components/home/RecentRoundsList.tsx:139`, `apps/web/src/pages/rounds/RoundsPage.tsx:172` — `r.total_score ? r.total_score : '—'` is equivalent to `r.total_score || '—'`; either is fine, just flagging the more idiomatic form exists if you want it (scores are never negative here, so `||` is safe).

---
Looked good: the `total_score === 0` sentinel handling is now consistent between `avgScore`/`bestScore` (already fixed in `index.tsx`) and all three literal-render call sites; `inferHoleCount` is reused from `@oga/core` rather than re-implemented in `apps/web`/`apps/mobile` (correctly respects the pure-logic-lives-in-core rule); the `styledata` listener fix and the hole-count deep-link clamp are both narrow, correctly scoped to the bug they close, and don't touch unrelated code.

---

PR #730, opened 2026-07-09
fix(mobile): Android GPS permission gate via PermissionsAndroid — expo-location request hangs on-device

cner-smith:
The #704 fix moved position reads to Mapbox's locationManager but left the permission gate on expo-location. On a Galaxy S23 production build, `requestForegroundPermissionsAsync()` hangs the same way the module's position APIs do — the 10 s timeout fires on every run (confirmed via release logcat: `[useHoleState perm-timeout]` at exactly +10 s, twice), the gate treats it as not-granted, and the Mapbox chain below never executes. Symptom: recenter grayed + no auto-place while the (native, expo-location-free) LocationPuck renders fine.

Fix: Android checks/requests `ACCESS_FINE_LOCATION` through RN core's `PermissionsAndroid` (own native module; `check` is instant when already granted). iOS keeps expo-location, which works there. The 10 s defensive race is retained on both platforms.

How verified: mechanism confirmed on-device via logcat (see #729 for the capture); fix is typecheck-clean; on-device confirmation happens via OTA to the production channel after merge (pure-JS change).

Closes #729

### DIFF
--- a/apps/mobile/components/round/hole/useHoleState.ts
+++ b/apps/mobile/components/round/hole/useHoleState.ts
@@ -1,5 +1,5 @@
 import { useEffect, useMemo, useRef, useState, type Dispatch, type SetStateAction, type MutableRefObject } from 'react'
-import { AppState } from 'react-native'
+import { AppState, PermissionsAndroid, Platform } from 'react-native'
 import * as Location from 'expo-location'
 import { locationManager } from '@rnmapbox/maps'
 import AsyncStorage from '@react-native-async-storage/async-storage'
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(mobile): Android GPS permission gate via PermissionsAndroid — expo-location request hangs on-device

**Concerns**
- PR description — the new Android path (`PermissionsAndroid.check`/`request`) is described as typecheck-clean with on-device confirmation deferred to OTA in production after merge. The logcat evidence cited (#729) confirms the *old* hang, not that this specific fix resolves it on-device. Since this gates the core GPS auto-place flow, a manual smoke test on real Android hardware before merge (an emulator won't reproduce the hang) would de-risk shipping straight to production via OTA — may not be feasible depending on your setup, flagging as a risk note rather than a blocker.

**Nits**
- `apps/mobile/components/round/hole/useHoleState.ts:245` — the trailing comment "expo-location stays for the permission prompt above only" is now stale: as of this diff, expo-location is only used for the permission prompt on iOS; Android goes through `PermissionsAndroid` (lines 216–221). Worth a one-line correction so the next person debugging a permission hang isn't pointed at the wrong module.

---
Looked good: scope held to the single stated concern (Android permission gate), no `any`, the updated `eslint-disable-next-line` carries an explanation per CONTRIBUTING.md's rule, and the fix stays in `apps/mobile` platform-wiring code rather than misplacing logic — this isn't shared business logic that belongs in `@oga/core`.

---

PR #731, opened 2026-07-09
feat(mobile): recenter button snaps the ball back to GPS during placement

cner-smith:
Device-QA feedback: after manually dragging the ball marker there was no way to put it back on the player — the manual-placement freeze (correctly) blocks silent GPS clobbering, but nothing deliberately lifted it, and the recenter button was camera-only.

Now, during PLACE_BALL, tapping the recenter button recenters the camera AND resumes GPS-driven ball tracking: freeze lifted, Kalman restarted from the next fresh fix, ball snapped to the current fix immediately, `ballMoved` reset so the HUD's ball-from-GPS labeling returns. Outside placement (and in past-round review, which passes no handler) the button stays camera-only. A deliberate tap is explicit intent, so this doesn't conflict with the #713 anti-clobber protection. Accessibility label reflects the dual action in placement phase.

How verified: mobile typecheck clean; on-device via production-channel OTA after merge (pure JS).

### DIFF
--- a/apps/mobile/components/round/HoleMap.tsx
+++ b/apps/mobile/components/round/HoleMap.tsx
@@ -111,6 +111,14 @@ interface HoleMapProps {
   aimCommitted?: boolean
   onSetAim: (loc: LatLng) => void
   onSetBall: (loc: LatLng) => void
+  /**
+   * Called with the current GPS fix when the recenter button is tapped
+   * during PLACE_BALL — a deliberate tap is explicit intent to put the
+   * ball back on the player, unlike the silent GPS clobber the manual-
+   * placement freeze exists to prevent (#713). Parent resumes GPS-driven
+   * ball tracking. Optional: past-round review passes nothing.
+   */
+  onRecenterBall?: (loc: LatLng) => void
   onPlacePin?: (loc: LatLng) => void
   /**
    * Whether to mount the Mapbox LocationPuck. The puck owns its own
@@ -203,6 +211,7 @@ export function HoleMap({
   holeNumber,
   onSetAim,
   onSetBall,
+  onRecenterBall,
   onPlacePin,
   showLocationPuck,
   tapToPlaceBall = true,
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — feat(mobile): recenter button snaps the ball back to GPS during placement

**Concerns**
- `apps/mobile/components/round/LiveRoundSession.tsx:562` — the new `onRecenterBall` handler calls `finalState.setBall(loc)` unconditionally except for `isPastMode`; there's no check for `isRevisitingPlayedHole`. `useHoleState.ts`'s own GPS listener explicitly gates ball updates on that flag ("keep the GPS chip/recenter live... but don't auto-drive the BALL marker" — the #484 review posture), and `roundState` stays `PLACE_BALL` while reviewing a previously-played hole, so the recenter button is rendered and tappable there too. A tap would now silently move the ball off its historical breadcrumb position onto live GPS, bypassing the `onAddShot` opt-in the rest of the review flow uses to re-arm live aids. This may be intentional — a deliberate tap is arguably explicit intent regardless of revisit state — but the PR description only reasons about the interaction with #713's anti-clobber freeze, not this one. Worth confirming it was considered.

**Nits**
- `apps/mobile/components/round/HoleMap.tsx:114` — the `onRecenterBall` doc comment says "past-round review, which passes no handler," but the only caller (`LiveRoundSession.tsx`) always passes the prop and instead gates on `isPastMode` inside the callback body. The comment doesn't match the actual call site.
- `apps/mobile/components/round/HoleMap.tsx:936` — `isPlaceBallPhase && onRecenterBall ? '...' : '...'` — `isPlaceBallPhase` is already guaranteed true here (the whole block is wrapped in `{isPlaceBallPhase && (...)}`), so that half of the condition is dead; `onRecenterBall ? '...' : '...'` reads the same.

---
Looked good: single concern held to the recenter button, reuses the existing #713 manual-freeze/Kalman-reset primitives instead of adding new state, and the accessibility label change reflects the new dual behavior.

---

Issue #601, opened 2026-06-13
Lift shot "is-putt" + shot-summary formatting into @oga/core (cross-surface dupes)


cner-smith:
Surfaced during #596 review. Two related cross-surface duplications that the package-boundary rule says belong in `@oga/core`:

1. **"Is this a putt" heuristic** — `lie_type === 'green' || club === 'putter'` is hand-inlined ~10× across web + mobile (e.g. `apps/mobile/components/round/PastRoundMap.tsx`, `PastHoleShotsSheet.tsx`, web shot components). No canonical helper exists in core (verified — core only has scattered `lieType === 'green'` checks). Add `isPutt(shot)` / `isOnGreen(shot)` to `@oga/core` and replace the inlined copies.
2. **Shot-summary formatting** — mobile has `summarizeShot` (`PastRoundMap.tsx`) and web has `formatShotSummary` (`apps/web/src/components/rounds/ShotEntryModal.tsx`), independently built with different shapes. Reconcile into one core formatter (display-string output) the apps wire.

Both are pure domain/format logic → `packages/core/src/` with tests (`@oga/core` exports need tests). Low urgency, no behavior change — a consolidation pass, do deliberately.

cner-smith:
Partial progress: **part 1 (is-putt helper) shipped** — `isPuttShot(lieType, club)` now lives in `@oga/core` (`round.ts`) with tests, and all web call sites use it (PR #683). **Still open:** (a) migrate mobile's hand-rolled predicates (`PastRoundMap.tsx`, `PastHoleShotsSheet.tsx`) to the helper, and (b) lift shot-summary formatting (`summarizeShot`) into core. The mobile-migration + a related putter-classification wrinkle are tracked in #691. Keeping this open for the shot-summary lift + mobile migration.

PR #732, opened 2026-07-10
fix: putt classification is green-lie only — putter off the green is a normal shot

cner-smith:
Closes #691.

## Problem
Write paths classified putts by `lie === 'green'` only, but read paths (core `isPuttShot`, mobile putt counts, past-round display) used `lie === 'green' || club === 'putter'`. A putter played from off the green (Texas wedge) was therefore **written** as a normal shot (`distance_to_target` in yards, no putt fields) but **read back** as a putt — rendering the yards value as feet (~3× wrong) and inflating putt counts.

## Decision (product)
**Option B from #691: putt = green lie only.** A putter off the green is a normal shot (yards, SG around-green) — matches what the save paths already write and how the SG engine (`sg-calculator`, lie-only since inception) already scores it. No UI changes needed.

## Changes
- `@oga/core` `isPuttShot(lieType)` — drops the `club === 'putter'` OR term and the club param; comment records the #691 decision. Tests updated (Texas wedge case pinned).
- Migrates every remaining hand-rolled copy of the predicate onto `isPuttShot` (the is-putt half of #601):
  - mobile: `useHoleData` remote/local putt counts, `useShotActions` persist path, `PastHoleShotsSheet` row + editor, `PastRoundMap.summarizeShot`
  - web: `ShotEntryModal` `isPuttSave`/`isPutt` (were inline lie checks)
- Intentionally untouched: `stats.ts` putter exclusions in club-distance/accuracy tables (club-based exclusion, not putt classification).

#601's other half (reconciling `summarizeShot`/`formatShotSummary` into one core formatter) is a separate no-behavior consolidation PR.

## Verification
- `pnpm test` green (isPuttShot suite re-pinned for lie-only).
- Direct `tsc --noEmit` clean on web and mobile (not turbo-cached).

### DIFF
--- a/apps/mobile/components/round/PastHoleShotsSheet.tsx
+++ b/apps/mobile/components/round/PastHoleShotsSheet.tsx
@@ -18,6 +18,7 @@ import {
   combinedPuttResult,
   formatClubLabel,
   formatDistance,
+  isPuttShot,
   type BreakDirectionHorizontal,
   type BreakDirectionVertical,
   type DistanceUnit,
@@ -276,7 +277,7 @@ function ShotRowView({
   unit: DistanceUnit
   onPress: () => void
 }) {
-  const isPutt = shot.lie_type === 'green' || shot.club === 'putter'
+  const isPutt = isPuttShot(shot.lie_type)
   const clubLabel = isPutt
     ? 'Putt'
     : shot.club
@@ -426,7 +427,7 @@ function EditShotSheet({
     (shot.break_direction_horizontal as BreakDirectionHorizontal | null) ?? null,
   )
 
-  const isPutt = lieType === 'green'
+  const isPutt = isPuttShot(lieType)
 
   const idx = allShots.findIndex((s) => s.id === shot.id)
   const prevShot = idx > 0 ? allShots[idx - 1]! : null
--- a/apps/mobile/components/round/PastRoundMap.tsx
+++ b/apps/mobile/components/round/PastRoundMap.tsx
@@ -8,6 +8,7 @@ import {
   destinationYards,
   formatClubLabel,
   formatDistance,
+  isPuttShot,
   type DistanceUnit,
   type LieType,
   type ShotResult,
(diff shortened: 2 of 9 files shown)


cner-smith:
## Review — fix: putt classification is green-lie only — putter off the green is a normal shot

**Concerns**
- `packages/core/src/round.ts:110` — the reclassification is retroactive on already-persisted data. Before this PR, mobile's `buildPayload` (`apps/mobile/components/round/hole/useShotActions.ts`) called the old, club-aware `isPuttShot(lieType, club)`, and `ShotLogger`'s "Club" chip row lets a player pick `putter` independently of the "Lie type" row (they're two separate `ChipRow`s, not gated on each other) — so a Texas wedge logged through the standard (non-`PuttingSheet`) mobile flow would have hit `isPutt = true` on the old OR logic and gotten `distance_to_target: null`. That branch of `ShotLogger` also never collects putt fields (those only render when `lieType === 'green'`), so the row would have *no* distance data at all. After this fix ships, that same historical row reads as a normal shot but still has no yardage to show — a blank/broken row in `PastHoleShotsSheet`/`PastRoundMap` for any pre-fix mobile-logged Texas wedges. Unclear from the diff whether this path was ever actually exercised in production or whether it's already an accepted, narrow edge case — if it's more than theoretical, worth a one-line note on the known gap (a backfill isn't obviously warranted for what should be a rare row shape).

---
Looked good: core `isPuttShot` is the single place the rule changed, tests are re-pinned for the new signature (`packages/core/src/__tests__/round.test.ts`), and the PR migrates every remaining hand-rolled `lie==='green'||club==='putter'` copy onto the core function rather than leaving them to drift again — exactly the "pure logic lives in `@oga/core`" rule, and the right amount of blast radius for what's still one concern (this classification, not a broader `summarizeShot`/`formatShotSummary` consolidation, which is correctly called out as a separate PR). No secrets, no migrations, targets `dev`.

---

cner-smith:
Review concern verified — the affected window is narrower than it reads: mobile's buildPayload only began writing distance_to_target in #722 (2026-07-09); before that every mobile live-logged shot had a null yardage (bug #709). So a 'putter + non-green lie via the standard logger' row with nulled distance could only be created between the #722 OTA and this fix (~1 day, pre-tester). Accepted as a known narrow gap, no backfill — and the follow-up shot-summary PR (#601) adds a haversine-to-next-shot fallback on mobile's stepper, which recovers a displayed yardage for exactly this row shape.

PR #733, opened 2026-07-10
chore: lift shot-summary formatting into @oga/core

### LINKED ISSUE #601: Lift shot "is-putt" + shot-summary formatting into @oga/core (cross-surface dupes)

cner-smith:
Closes #601 (second half; the is-putt half shipped in #732).

## What
Web `formatShotSummary` (ShotEntryModal) and mobile `summarizeShot` (PastRoundMap) were independently built formatters for the same shot-summary line. This lifts the shared derivation into `@oga/core`:

- **`summarizePuttParts(shot, unit)`** — unit-aware putt distance (`formatPuttDistance`) + result. Miss derives **axes-first** (`putt_distance_result`/`putt_direction_result` are the canonical columns) with legacy `putt_result` as the back-compat fallback.
- **`summarizeShotParts(shot, next, unit)`** — distance + full result label, with a haversine fallback to the next shot's start when `distance_to_target` is null.

Apps keep thin wrappers for what genuinely differs: mobile prepends `Putt`/club·lie (its stepper is a single line; web shows club·lie on its own row line) and each picks its empty-state text.

## Visible diffs (canonicalization — all richer, per #691/#601 decision)
- mobile putt line: `12 ft` → unit-aware (`3.7 m` in meters mode); raw lowercase axis values → `Short Left`
- mobile shot line: gains a yardage via haversine fallback where `distance_to_target` is null (also recovers a distance for the pre-#732 Texas-wedge rows flagged in #732's review)
- mobile result labels: `Push R` → `Push right` (app-local `SHOT_RESULT_SHORT` deleted)
- web putt line: axes-first miss (`Short Left`) instead of the flattened legacy label (`Short`)

## Notes
- Mobile's next-shot lookup matches on `hole_score_id` **and** `shot_number + 1` — the `shots` prop is round-wide and `shot_number` restarts per hole.
- 11 new core tests (589 total); direct `tsc --noEmit` clean on web and mobile.

### DIFF
--- a/apps/mobile/components/round/PastRoundMap.tsx
+++ b/apps/mobile/components/round/PastRoundMap.tsx
@@ -7,11 +7,11 @@ import {
   bearingDegrees,
   destinationYards,
   formatClubLabel,
-  formatDistance,
   isPuttShot,
+  summarizePuttParts,
+  summarizeShotParts,
   type DistanceUnit,
   type LieType,
-  type ShotResult,
 } from '@oga/core'
 import type { Database } from '@oga/supabase'
 import { supabase } from '../../lib/supabase'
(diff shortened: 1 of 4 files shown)


cner-smith:
## Review — chore: lift shot-summary formatting into @oga/core

**Concerns**
- `packages/core/src/round.ts:154` (`summarizePuttParts`) — `feet` now falls back to `shot.distance_to_target` for putts (`shot.putt_distance_ft ?? shot.distance_to_target`), same as the old web behavior, but the PR's "Visible diffs" list doesn't call this out as a change for mobile. Mobile's pre-PR `summarizeShot` never fell back to `distance_to_target` for putt rows — only `putt_distance_ft`. If any putt rows have `distance_to_target` set but `putt_distance_ft` null, mobile's stepper will now show a distance where it previously showed none. Worth confirming this is intentional (it's plausible putt rows never populate `distance_to_target`, in which case this is a no-op) — if so, might be worth a line in the PR body since the rest of the diffs are enumerated carefully.

**Nits**
- `packages/core/src/round.ts:117-125` — `ShotSummaryFields` types `shot_result` and `putt_result` as plain `string | null` rather than `ShotResult | null` / the legacy putt-result union, pushing the `as keyof typeof …` casts into the formatter bodies (`round.ts:167`, `round.ts:205`). This mirrors the pre-existing web code's casting style, so not a new pattern, but now that the logic is centralized in `@oga/core` it might be worth tightening the field types so the cast can move to the (thinner) call sites or disappear.

---
Looked good: this is the right fix for the actual problem — `SHOT_RESULT_SHORT` in `PastRoundMap.tsx` was diverging from web's `SHOT_RESULT_LABELS` (the `Push R` vs `Push right` drift called out in the PR body), and CONTRIBUTING.md's "Pure logic lives in `@oga/core`" rule is unambiguous that shared format helpers belong there regardless of caller count — the two-caller extraction here is justified by that rule, not the 3-caller helper-extraction guideline, and the old app-local comment explaining the prior "too view-specific to lift yet" call is cleanly removed along with the code it was justifying. `summarizePuttParts`/`summarizeShotParts` are new `@oga/core` exports with tests in the same package (`packages/core/src/__tests__/round.test.ts`) covering made/miss/single-axis/legacy-fallback/unit-conversion/haversine-fallback/empty-signal cases. Single concern, targets `dev`, no migrations or secrets touched.

---

cner-smith:
Concern confirmed intentional (canonicalize-on-web-behavior): all three current write paths null distance_to_target on putt rows (web ShotEntryModal save, web review-save in useRoundActions, mobile buildPayload), so the fallback is a no-op for well-formed rows — it only fires on legacy/edge rows (e.g. a non-putt edited into a putt with stale columns, #663), where showing the stored distance beats showing nothing. That's the same reason web carried the fallback.

Type-tightening nit declined for now: callers pass raw DB Row types whose columns are `string | null`, so union-typing ShotSummaryFields would just move the casts onto every caller instead of one formatter body.

PR #734, opened 2026-07-10
fix(seed): demo practice plan uses real per-drill rationale

### LINKED ISSUE #606: Demo practice plan shows 'Drill' placeholders — seeded drill_ids don't resolve on prod
The demo/seed account (`TerryMindle`) Practice screen renders every drill as a generic **"Drill" / "Targets the lowest SG category."** on live web.

## Root cause (not the real generator)
The plan is `scripts/insertPracticePlan` seed data: each block stores `drill_id` + a hardcoded `rationale: 'Targets the lowest SG category.'` (`seed-demo.ts:719`). The drill **name** is resolved at render via `useDrillsByIds(drill_id)` and falls back to `'Drill'` (`PracticePlanPage.tsx:480`) — so the seeded `drill_id`s no longer resolve against prod's `drills` table. The seed fetched real drills at seed time (`category=approach limit 3`), so the IDs were valid then; they're stale now (seeded against a different/older drills set, or `drills` re-inserted with new IDs).

## Scope
- This is the **demo/seed account**, not the real generator (a generated plan has real per-drill rationale + current IDs). Real users are likely fine — **spot-check one real generated plan to confirm**.

## Fix
1. Re-seed the demo account's practice plan against **prod** so `drill_id`s resolve.
2. Improve `insertPracticePlan` to use a real per-drill rationale instead of the placeholder, so the demo reads like a genuine plan (important for the App Store review/demo account).
3. Optional hardening: `PracticePlanPage` could show a softer fallback than bare "Drill" when a drill_id doesn't resolve, but the real fix is valid IDs.

Touches prod data → do deliberately.

cner-smith:
Closes #606.

## Investigation (prod, read-only first)
- The demo account (`TerryMindle`) had TWO plans: the stale 2026-06-10 seeded plan (3 blocks, **0/3 drill_ids resolved** — the 'Drill' placeholders) and a genuinely generated 2026-07-05 plan (**5/5 drill_ids resolve**, real per-drill rationales).
- `getLatestPracticePlan` fetches newest-only, so the page has shown the good plan since 07-05 — the reported symptom self-healed when that plan was generated. Expiry is soft (`isExpired` relabels + offers Generate but still renders the full plan), so the demo tab stays populated past its 07-12 `valid_until`.
- Real generator confirmed fine — the resolving plan IS an engine-generated one.

## Changes
1. **Prod data (already applied, deliberate):** deleted the stale seeded plan row (`1f1132d7…`, scoped to the demo user, `RETURNING` confirmed exactly 1 row) so it can never resurface as 'latest'.
2. **Seed script:** `insertPracticePlan` block `rationale` now uses the drill's `description` (the same copy a generated plan carries) instead of the flat placeholder, with the old string as fallback. Future re-seeds read like genuine plans.

Skipped the issue's optional softer 'Drill' fallback in `PracticePlanPage` — the real fix is valid IDs, and the demo account now has them.

### DIFF
--- a/scripts/seed-demo.ts
+++ b/scripts/seed-demo.ts
@@ -716,7 +716,9 @@ async function insertPracticePlan(userId: string): Promise<void> {
             order: i,
             type: blockTypes[Math.min(i, blockTypes.length - 1)],
             minutes: d.duration_min ?? 15,
-            rationale: 'Targets the lowest SG category.',
+            // Real per-drill copy so a seeded plan reads like a generated
+            // one — the flat placeholder here was #606's App Review risk.
+            rationale: d.description ?? 'Targets the lowest SG category.',
             drill_id: d.id,
             target: null,
           })),


cner-smith (commented):
## Review — fix(seed): demo practice plan uses real per-drill rationale

**Nits**
- `scripts/seed-demo.ts:721` — this reuses the drill's static `description` column (documented in CONTRIBUTING.md's drill schema as "one-line summary" of what the drill *is*) as the block's `rationale`. The real generator's `rationale` is a distinct field — "[o]ne sentence on why this drill is here" (`packages/core/src/practice-plan/prompt.ts:129-133`), i.e. why it was picked for this plan, not what it does. For most drills the description will read fine as a stand-in, but it's not quite the same thing the PR title claims ("real per-drill rationale") — it's real per-drill *copy*, borrowed from a different field. Not a blocker for a seed script; flagging in case a reader expects this to match the generator's rationale semantics.

---
Looked good: minimal, single-concern fix scoped to the seed script; comment explains the non-obvious *why* (#606 App Review risk) per CONTRIBUTING's comment guidance; fallback to the old placeholder string preserves behavior when `description` is null; no `@oga/core` exports, migrations, or app-layer logic touched, so no test-coverage or package-boundary concerns.

---

cner-smith:

---

PR #735, opened 2026-07-10
perf(mobile): screen cache kills refetch-on-revisit spinners

### LINKED ISSUE #599: Perf: cache screen data to kill refetch-on-revisit (screen-to-screen loading)

cner-smith:
Closes #599.

## Design: stale-while-revalidate, always-revalidate
New `lib/screenCache.ts` — a session-lifetime in-memory `Map`. Screens seed their state from it so a revisit renders instantly instead of spinning; every consumer is a stack screen whose existing fetch re-runs on mount (or focus), so the network response always races behind and replaces the cached render. **Staleness is bounded by one fetch latency and self-corrects** — no TTLs, no persistence.

## Wired screens
- **Home** — profile + recent rounds seed from cache; the existing focus-refetch (#705/#707) is the revalidate.
- **All-rounds list** — same, on its mount fetch.
- **Round detail** — `{round, courseName, courseCenter, holes, holeScores, shots}` bundled per `round:${id}`. Seeded inside the fetch effect (works across param changes between rounds), and **mirror-written from state** so in-screen scorecard/sheet edits re-cache — a revisit seeds post-edit data. Unfinished rounds are never cached (they redirect to the live session before the summary fetch).
- **Stats** — also fixes the issue's second item: the L5/L10/L20 chips used to refire the whole nested rounds query. Now fetch L20 once + `slice(0, n)` client-side (query orders `played_at` desc, so the slice is exactly what a limit-n fetch returned). Toggle is instant.

## Invalidation
Coarse `clearScreenCache()` at the four places where even a one-fetch stale flash is wrong:
- round **delete** — home, list, and detail handlers (a stale cache would resurrect the deleted round as a ghost, #705 redux)
- round **finalize** — `completeRound` after the successful rounds write (totals/SG/handicap all change)
- **sign-out** — both profile paths (account switching must not leak cached screens)

Everything else rides the always-revalidate. Per-key invalidation deliberately skipped until coarse-clear demonstrably thrashes.

## Verification
- Mobile `tsc --noEmit` clean (direct, not turbo).
- Needs on-device QA for feel: revisit home/stats/round detail (no spinner, instant paint), L-window toggle (instant), delete → home (no ghost), finalize → home (fresh totals), sign-out → sign-in (no stale screens).

### DIFF
--- a/apps/mobile/app/(app)/index.tsx
+++ b/apps/mobile/app/(app)/index.tsx
@@ -9,6 +9,7 @@ import { useAuth } from '../../hooks/useAuth'
 import { useActiveRound } from '../../hooks/useActiveRound'
 import { syncPendingShots } from '../../lib/sync'
 import { pendingCount } from '../../lib/db'
+import { clearScreenCache, getCached, setCached } from '../../lib/screenCache'
 import { AppBar } from '../../components/ui/AppBar'
 import { ConfirmDialog } from '../../components/ui/ConfirmDialog'
 import { Entrance } from '../../components/ui/Entrance'
@@ -52,8 +53,14 @@ const KICKER: import('react-native').TextStyle = {
 
 export default function Home() {
   const { user } = useAuth()
-  const [profile, setProfile] = useState<Profile | null>(null)
-  const [rounds, setRounds] = useState<RecentRound[]>([])
+  // Seed from the screen cache so a revisit renders instantly; the focus
+  // fetch below always re-runs and replaces this (#599).
+  const [profile, setProfile] = useState<Profile | null>(
+    () => getCached<Profile>('home:profile') ?? null,
+  )
+  const [rounds, setRounds] = useState<RecentRound[]>(
+    () => getCached<RecentRound[]>('home:rounds') ?? [],
+  )
   const activeRound = useActiveRound()
   const [pending, setPending] = useState(0)
   const [pendingDelete, setPendingDelete] = useState<{
@@ -78,6 +85,9 @@ export default function Home() {
       try {
         const { error } = await deleteRound(supabase, id, user.id)
         if (error) throw error
+        // Wipe ALL cached screens — a stale rounds-list or round-detail
+        // cache would resurrect the deleted round as a ghost (#705 redux).
+        clearScreenCache()
         setRounds((prev) => prev.filter((r) => r.id !== id))
       } finally {
         setDeleting(false)
@@ -103,7 +113,10 @@ export default function Home() {
           console.error('[home/getProfile]', error.message)
           return
         }
-        if (data) setProfile(data as unknown as Profile)
+        if (data) {
+          setProfile(data as unknown as Profile)
+          setCached('home:profile', data)
+        }
       })
       getRecentRounds(supabase, user.id, 20).then(({ data, error }) => {
         if (!active) return
@@ -112,7 +125,10 @@ export default function Home() {
           console.error('[home/getRecentRounds]', error.message)
           return
         }
-        if (data) setRounds(data as RecentRound[])
+        if (data) {
+          setRounds(data as RecentRound[])
+          setCached('home:rounds', data)
+        }
       })
       return () => {
         active = false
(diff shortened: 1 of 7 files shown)


cner-smith:
## Review — perf(mobile): screen cache kills refetch-on-revisit spinners

**Concerns**
- `apps/mobile/lib/screenCache.ts:1-24` — This is a generic, React-Native-agnostic keyed `Map` cache with no mobile-specific dependencies. CONTRIBUTING.md's package-boundaries section says: "Don't add helpers to `apps/web/src/lib` or `apps/mobile/lib` that other surfaces would want to reuse — lift them to `@oga/core`." A session-lifetime stale-while-revalidate store isn't quite the kind of thing the doc's examples cover (SG calc, units, format helpers, post-round inference), so I'm not fully sure this rule is meant to catch it — but if web ever wants the same instant-revisit pattern on its own screens, this would need to move or get duplicated rather than reused as-is. Worth a line on why this is mobile-only, if that's deliberate.

**Nits**
- `apps/mobile/lib/screenCache.ts:10` — "ponytail: coarse full-clear; per-key invalidation only if this thrashes." The word `ponytail:` doesn't parse as English and looks like a leftover editing artifact — probably meant to be "Note:" or similar.

---
Looked good: invalidation (`clearScreenCache()`) is wired into every mutation that would actually go stale-wrong — delete, finalize, sign-out — not just the happy path. The round-detail seed/mirror-write effect (`round/[id]/index.tsx`) correctly gates the cache write on `!loading && !error && !redirectToLive`, so a mid-navigation transition between two rounds can't write round A's data under round B's cache key. The stats L5/L10/L20 change depends on `getRoundsWithDetails` ordering by `played_at desc` for the client-side `slice(0, n)` to be correct — confirmed that in `packages/supabase/src/queries/rounds.ts:127`, so that part checks out.

---

cner-smith:
Both addressed in the follow-up commit:

- **Mobile-only is deliberate, now documented in the module comment:** the cache's safety depends on every consumer being a stack screen that re-fetches on mount/focus (the always-revalidate contract). Web's screens don't share that lifecycle, so lifting the bare Map to @oga/core would export the mechanism without the invariant that makes serving cached data safe. If web ever wants the pattern, the lift should carry the contract with it — deferred until that caller exists.
- **Comment marker reworded** to plain English.

PR #738, opened 2026-07-10
fix(mobile): email confirmation round-trip — PKCE, never-hanging callback, self-clearing check-email screen

### LINKED ISSUE #509: [mobile] Email confirmation redirects to localhost — never returns to the app

cner-smith:
Fixes the #509 field failures (two real iPhone users, production build). Closes #509.

## The two reports
1. Signed up → confirmed → **'Check your email' screen never cleared** (account created fine).
2. Signed up → tapped the link → **stuck at 'Confirming your account…'** — this one pinned the root cause: the deep link FIRES (the callback screen mounted), but the implicit-flow **token fragment didn't survive** the mail-app → browser → `oga://` handoff. The callback's `if (!fragment) return` then sat on the spinner forever.

## Fix, three layers
**1. PKCE flow (root cause).** Mobile client switches to `flowType: 'pkce'` (new `createOgaClient` option; web untouched — omitted = implicit default). The confirm redirect now carries `?code=` in the **query string**, which survives redirect chains that strip fragments. `exchangeCodeForSession` uses the `code_verifier` in SecureStore — the exchange only works on the signup device, which is also the Android scheme-hijack hardening #509 tracked as pre-launch work.

**2. `auth-callback` can no longer hang.** Handles `?code=` (PKCE), keeps the fragment path for legacy links minted pre-switch, surfaces `error_description` from query OR fragment, and adds a no-payload fallback + 8s watchdog: GoTrue verifies the email server-side BEFORE redirecting, so the fallback says so and offers 'Back to sign in' instead of spinning.

**3. Self-clearing 'Check your email' screen.** Covers the deep-link-never-fires modes (confirmed on another device, in-app mail browsers, dismissed open-in-app prompt): the screen still holds the just-entered credentials, so it silently attempts `signInWithPassword` on every app-foreground return (the natural just-confirmed-in-Mail moment) plus a manual 'I've confirmed it' button. Fails 'Email not confirmed' until the link is tapped; succeeds right after → same welcome-splash route as login. Turnstile-aware (prod captcha ON): the card mounts its own widget, tokens consumed per attempt + widget remounted for the next mint, queued attempts fire on token arrival, stale signup token dropped on submit.

## For currently-stuck users
Their accounts ARE confirmed — force-quit, reopen, sign in normally. This ships OTA (pure JS).

## Verification
- Mobile + web `tsc --noEmit` clean (direct), 589 core tests pass.
- Needs on-device iOS verification: same-device confirm (PKCE round-trip), other-device confirm (button path), foreground auto-clear, and a legacy pre-PKCE link (fragment path).

Related: #509's dashboard config (done) and PR #510 (the original deep-link wiring).

### DIFF
--- a/apps/mobile/app/(auth)/signup.tsx
+++ b/apps/mobile/app/(auth)/signup.tsx
@@ -45,19 +45,29 @@ export default function Signup() {
   const captchaTokenRef = useRef<string | null>(null)
   const attemptInFlightRef = useRef(false)
   const wantAttemptRef = useRef(false)
+  // Auto attempts (foreground returns) keep a minimum spacing so a user
+  // fidgeting between apps can't burn GoTrue's auth rate budget on
+  // speculative pre-confirmation attempts — getting rate-limited on the
+  // REAL attempt would reproduce the stuck screen this exists to fix.
+  // The manual button is exempt (explicit user intent).
+  const lastAutoAttemptAtRef = useRef(0)
 
   const captchaEnabled = Boolean(TURNSTILE_SITE_KEY)
   const canSubmit = !loading && (!captchaEnabled || captchaToken !== null)
 
   async function tryConfirmSignIn(manual: boolean) {
     if (attemptInFlightRef.current) return
+    if (!manual && Date.now() - lastAutoAttemptAtRef.current < 20_000) return
     const token = captchaTokenRef.current
     if (captchaEnabled && !token) {
       // No token yet (widget still minting after the last attempt consumed
       // one) — remember the intent; the token-arrival effect fires us.
+      // Deliberately NOT stamped as an attempt — the queued retry must not
+      // be suppressed by the auto-attempt interval.
       wantAttemptRef.current = true
       return
     }
+    if (!manual) lastAutoAttemptAtRef.current = Date.now()
     attemptInFlightRef.current = true
     wantAttemptRef.current = false
     if (manual) setChecking(true)
@@ -81,10 +91,14 @@ export default function Signup() {
       return
     }
     // "Email not confirmed" is the expected pre-confirmation result; stay
-    // quiet on auto attempts, give feedback on the button.
+    // quiet on auto attempts, give feedback on the button. Other failures
+    // (network, captcha, rate limit) show their real message — masking
+    // them as "not confirmed" would misdirect the user.
     if (manual) {
       setConfirmHint(
-        'Not confirmed yet — tap the link in your email, then try again.',
+        signInError.message.toLowerCase().includes('not confirmed')
+          ? 'Not confirmed yet — tap the link in your email, then try again.'
+          : signInError.message,
       )
     }
   }
@@ -97,6 +111,10 @@ export default function Signup() {
       if (s === 'active') void tryConfirmSignIn(false)
     })
     return () => sub.remove()
+    // tryConfirmSignIn is redefined every render; listing it would
+    // re-subscribe the listener per render. Its mutable inputs are refs;
+    // the state it reads (email/password) can't change once submitted —
+    // the form is unmounted — so the captured closure stays correct.
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [submitted])
 
@@ -106,6 +124,9 @@ export default function Signup() {
     if (submitted && captchaToken && wantAttemptRef.current) {
       void tryConfirmSignIn(false)
     }
+    // tryConfirmSignIn omitted for the same reason as the AppState effect
+    // above — per-render identity, and its closure inputs are stable
+    // after submit.
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [submitted, captchaToken])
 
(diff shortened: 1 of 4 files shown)


cner-smith:
## Review — fix(mobile): 'Check your email' screen self-clears after confirmation

**Concerns**
- `apps/mobile/app/(auth)/signup.tsx:94-101` — The `AppState` listener fires `tryConfirmSignIn(false)` on every foreground return with no debounce or minimum interval between attempts. Each attempt calls `signInWithPassword` against GoTrue and consumes/re-mints a Turnstile token. A user who backgrounds/foregrounds the app a few times while waiting on the confirmation email (checking notifications, taking a call) will generate several failed sign-in attempts in a row. Supabase/GoTrue rate-limits auth requests — if that budget gets used up on these speculative pre-confirmation attempts, the real post-confirmation attempt could get rate-limited too, which would reproduce the exact "stuck screen" symptom this PR is fixing. Worth considering a minimum interval between attempts (e.g. ignore foreground events within N seconds of the last one).
- `apps/mobile/app/(auth)/signup.tsx:83-89` — Any `signInWithPassword` failure — not just the expected "email not confirmed" — is surfaced to the user as "Not confirmed yet — tap the link in your email, then try again." That's misleading if the actual cause is a network error, a rejected/expired captcha token, or a GoTrue rate limit, and there's no way for the user (or telemetry) to distinguish those cases from a genuinely-unconfirmed account.
- `apps/mobile/app/(auth)/signup.tsx:100` and `:109` — Bare `// eslint-disable-next-line react-hooks/exhaustive-deps` with no reasoning attached. CONTRIBUTING.md: "No `// eslint-disable-next-line` without a comment explaining why." The omission of `tryConfirmSignIn` from the deps looks intentional (it's redefined every render, so including it would re-run the effect on every render) — that reasoning should be written on the line rather than left implicit.

**Nits**
- `apps/mobile/app/(auth)/signup.tsx:166-187` — This Turnstile `WebView` block is a near-verbatim duplicate of the one further down in the file (used for the initial signup submission). Only two instances so this doesn't hit the 3-caller extraction rule yet, just flagging in case a third copy shows up later.

---
Looked good: refs (`captchaTokenRef`, `attemptInFlightRef`, `wantAttemptRef`) are used correctly to keep the `AppState` listener and queued-attempt effect from reading stale closure state; token consume/remint bookkeeping around the single-use Turnstile token is careful; scope stays to the one stuck-screen concern; no `@oga/core` or migration surface touched.

---

cner-smith:
All three concerns addressed in the follow-up commit: 20s minimum spacing on foreground auto-attempts (stamped only when an attempt proceeds, so the queued no-token retry can't suppress itself; manual button exempt), non-'not confirmed' failures now surface their real message, and both exhaustive-deps disables carry the reasoning. WebView-duplication nit left as flagged (2 instances, extraction on the 3rd).

PR #739, opened 2026-07-10
fix(core): formatSG rounds before picking sign — no more signed zero

cner-smith:
Closes #672.

`formatSG(-0.004)` rendered **'-0.00'** (and +0.004 → '+0.00') because the sign was picked from the raw value before `toFixed` rounded it to zero. A round with a near-zero SG showed a signed zero anywhere SG is displayed (home, stats, round detail — both platforms).

Fix: round to 2dp first (`Number(n.toFixed(2))` — collapses the rounded −0.00 to 0), then pick the sign, so anything that rounds to zero shows a clean '0.00'. Non-zero values are unchanged.

Test updated: the case that pinned the old '+0.00'/'-0.00' now asserts '0.00' for ±0.001/±0.004. 589 core tests pass.

### DIFF
--- a/packages/core/src/__tests__/units.test.ts
+++ b/packages/core/src/__tests__/units.test.ts
@@ -111,11 +111,13 @@ describe('formatSG', () => {
     expect(formatSG(-1.23999)).toBe('-1.24')
   })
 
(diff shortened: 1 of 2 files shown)


cner-smith:

---

Issue #284, opened 2026-05-13
Mobile: ShotLogger silently resets mid-entry when background sync changes shotNumber


cner-smith:
**Severity:** medium (data loss — player loses partially filled shot metadata)
**Audit source:** mobile-app-developer agent, 2026-05-13

## Mechanism

\`apps/mobile/app/(app)/round/[id]/hole/_components/HoleModals.tsx:109\`:

\`\`\`tsx
<ShotLogger key={shotNumber} ... />
\`\`\`

Keying by \`shotNumber\` forces a full remount on each new shot, which resets state. **Intentional for clearing stale form data.**

But \`shotNumber\` is derived from \`remoteShotCount + localShotCount + 1\`, and \`remoteShotCount\` is fetched async from Supabase. If a background sync (e.g. \`syncPendingShots\` in \`lib/sync.ts\`) completes while the player has the logger modal open:

1. \`remoteShotCount\` increments
2. \`shotNumber\` increments
3. The key changes
4. ShotLogger remounts with empty state
5. **Player loses any partially filled metadata** (club, lie, slope, result)

## Proposed fix

Key on \`currentHoleScore?.id\` plus a local sequence counter that only increments after \`persistShot\` succeeds, not on every remote count change.

\`\`\`tsx
const [shotEntrySeq, setShotEntrySeq] = useState(0)
// in persistShot success: setShotEntrySeq(s => s + 1)
<ShotLogger key={\`\${currentHoleScore?.id}-\${shotEntrySeq}\`} ... />
\`\`\`

## Files

- \`apps/mobile/app/(app)/round/[id]/hole/_components/HoleModals.tsx:109\`
- \`apps/mobile/app/(app)/round/[id]/hole/_hooks/useShotActions.ts\` (success path that should bump the local sequence)

cner-smith:
Done in PR #337 — ShotLogger now keyed by stable `shotEntrySeq` (monotonic, bumps only on persistShot success). useHoleData fetch race hardened with nonce-based cancellation pattern replacing the boolean `active` flag. Merged to `dev`; keeping open until `dev → main`.

PR #744, opened 2026-07-10
fix(mobile): ShotLogger re-seeds lie from initial on open

cner-smith:
Closes #654.

## Bug
Answering 'No' to the on-green prompt (or 'Not a putt? Chip / bunker →') opened the shot logger with **no lie preselected**, so a player who skipped the picker stored `lie_type: null` where the flow intended `rough` — silently degrading SG categorization.

## Mechanism (regression from the #284 key change)
`ShotLogger` reads `initial` only in its `useState` initializer, and only remounts when `shotEntryKey` changes (save / hole change). But `handleOnGreenNo` (`setLoggerInitial({ lieType: 'rough' })`) and `swapPuttingToShot` set `loggerInitial` **mid-cycle**, after this instance already mounted — so the seeded lie was set on a prop the mounted instance never re-reads.

## Fix
An effect keyed on `visible` re-seeds `value` from `initial` each time the sheet opens (mirrors the useState initializer exactly). Re-seeds on open only — deps are `[visible]` because `loggerInitial` is set before the sheet opens and isn't mutated while it's open, so in-progress edits are never wiped.

Mobile `tsc --noEmit` clean.

### DIFF
--- a/apps/mobile/components/round/ShotLogger.tsx
+++ b/apps/mobile/components/round/ShotLogger.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useState } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 import { Modal, Pressable, ScrollView, Text, TextInput, View } from 'react-native'
 import Svg, { Circle, Line as SvgLine } from 'react-native-svg'
 import {
@@ -112,6 +112,25 @@ export function ShotLogger({
   })
   const set = <K extends keyof ShotLoggerValue>(key: K, v: ShotLoggerValue[K]) =>
     setValue((prev) => ({ ...prev, [key]: prev[key] === v ? undefined : v }))
+
+  // Re-seed `value` from `initial` each time the sheet opens. ShotLogger only
+  // remounts when shotEntryKey changes (save / hole change), but the "not
+  // putting" flows seed loggerInitial mid-cycle AFTER this instance mounted —
+  // handleOnGreenNo sets { lieType: 'rough' }, swapPuttingToShot sets the
+  // chosen lie. Without this re-seed the mounted instance never reads that
+  // prop and the shot stores lie_type null, silently degrading SG (#654).
+  useEffect(() => {
+    if (!visible) return
+    setValue({
+      ...initial,
+      lieType: isPutt ? 'green' : initial?.lieType,
+      club: isPutt ? 'putter' : initial?.club,
+    })
+    // Re-seed on open only. initial/isPutt are stable while the sheet is open
+    // (loggerInitial is set before it opens and not mutated during), and
+    // listing them would wipe in-progress edits on any identity change.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [visible])
 
   // Source the club picker from the user's bag. Fall back to DEFAULT_BAG
   // while the bag is loading or if the user trimmed it to nothing — never


cner-smith (commented):
## Review — fix(mobile): ShotLogger re-seeds lie from initial on open

**Nits**
- `apps/mobile/components/round/ShotLogger.tsx:108-112` vs `apps/mobile/components/round/ShotLogger.tsx:124-128` — the seed object (`{...initial, lieType: isPutt ? 'green' : initial?.lieType, club: isPutt ? 'putter' : initial?.club}`) is now built twice, once in the `useState` initializer and once in the effect. They have to stay byte-for-byte in sync or the "mirrors the initializer exactly" invariant the comment relies on silently breaks. A tiny `seed(initial, isPutt)` local function would remove that drift risk — optional given CONTRIBUTING's "two similar lines are fine" bar, but this is four.

I traced the bug mechanism through `useShotActions.ts` (`handleOnGreenNo`/`swapPuttingToShot` call `setLoggerInitial` before `setLoggerOpen(true)`, so `visible` does transition false→true after `initial` updates) and it matches what the PR description claims — the `[visible]`-only effect is the right fix given `ShotLogger` is a single persistent instance keyed on `shotEntryKey` rather than remounting per open.

---
Looked good: root-caused correctly against the actual call sites (not just the symptom), the `eslint-disable-next-line` on the effect's deps carries the required explanation per CONTRIBUTING.md's "no `// eslint-disable-next-line` without a comment explaining why", and the fix is scoped to exactly the regression described — no drive-by changes.

---

cner-smith:
## Review — fix(mobile): ShotLogger re-seeds lie from initial on open

**Concerns**
- `apps/mobile/components/round/ShotLogger.tsx:107-110` vs `apps/mobile/components/round/ShotLogger.tsx:122-133` — the re-seed effect duplicates the `useState` initializer's seeding formula in a slightly different form (`isPutt ? 'green' : initial?.lieType` vs `seededLieType ?? initial?.lieType`). They're equivalent today, but there are now two places that decide "what does a freshly-opened ShotLogger start with," and nothing ties them together. If one is edited later without the other (e.g. a third seeded field is added), first-open and re-open state will silently diverge. Might be worth a shared `seedShotValue(initial, isPutt)` used by both the initializer and the effect.
- The `[visible]`-only dependency array is correct given the current call sites, but its correctness rests on an invariant that isn't enforced or documented anywhere near the setters themselves: `loggerInitial` must be committed at or before the same render where `loggerOpen`/`visible` flips true. I traced the three call sites in `apps/mobile/components/round/hole/useShotActions.ts` — `swapPuttingToShot` (lines 491-495) sets both in the same synchronous batch, and `handleOnGreenNo`/`handleOnGreenYes` set `loggerInitial` in an earlier, separate render before `confirmAim`/`skipAim` later flips `loggerOpen(true)` — so the invariant holds today. But a future caller that calls `setLoggerOpen(true)` and then sets `loggerInitial` afterward (e.g. after an `await`) would reintroduce #654 in a new shape, silently, since the effect wouldn't re-fire. This isn't something this PR needs to fix, but a one-line comment at the `setLoggerInitial` call sites (or on `loggerInitial`'s prop declaration in `HoleModals.tsx:31`) noting the ordering requirement would save whoever touches this next from rediscovering it the hard way.

- Consider whether `loggerOpen`/`loggerInitial` in `LiveRoundSession.tsx` could be collapsed into a single state value (e.g. `{ open: boolean; initial: ShotLoggerValue }`) so the ordering invariant above is structurally guaranteed instead of relying on call-site discipline.

---
Looked good: the effect's "re-seed on open only" claim checks out against every current caller (`handleOnGreenNo`, `handleOnGreenYes`→`swapPuttingToShot`, `confirmAim`/`skipAim`) — I traced each through `useShotActions.ts` and didn't find a path where in-progress edits get wiped or the original null-lie regression reappears; the `eslint-disable-next-line` is accompanied by a real explanation per CONTRIBUTING's rule, and the diff stays scoped to the one file/one concern.


---

PR #748, opened 2026-07-10
fix(mobile): Shot Patterns lie-slope filter — two-axis multi-select

cner-smith:
Closes #746.

## Bug
The Shot Patterns lie-slope filter read the legacy `lie_slope` column, which the app stopped writing at the two-axis migration — it's null on every shot logged since. So selecting any slope matched **zero shots**. It was also the old flat single-select chip row, inconsistent with the two-axis slope model used in shot logging.

## Fix
- `ShotRowMin` + `rowToShot` now carry `lie_slope_forward` / `lie_slope_side` onto the `Shot` (the query already selected them; only the mapping was missing). Legacy `lie_slope` mapping kept, harmless.
- Filter state is two axes, each **multi-select**: forward (uphill / level / downhill) + side (ball above / below). Empty selection on an axis = *any*; the two axes AND together. Tap several within an axis to see them together.
- New `MultiChipRow` (co-located — 2 callers here) mirrors `ChipRow`'s wrapping-chip style with checkbox semantics.
- `filterDispersionByLie` (core) already supported the two axes — no core change needed; slope multi-select is filtered inline since the core helper takes single values per axis.

Mobile `tsc --noEmit` clean. Needs an on-device check that selecting slopes now actually narrows the dispersion set.

### DIFF
--- a/apps/mobile/app/(app)/patterns.tsx
+++ b/apps/mobile/app/(app)/patterns.tsx
@@ -5,7 +5,8 @@ import { captureRef } from 'react-native-view-shot'
 import * as Sharing from 'expo-sharing'
 import {
   CLUBS,
-  LIE_SLOPES,
+  LIE_SLOPES_FORWARD,
+  LIE_SLOPES_SIDE,
   LIE_TYPES,
   YARDS_TO_METERS,
   clubDistanceStats,
@@ -19,6 +20,8 @@ import {
   type DispersionStats,
   type DistanceUnit,
   type LieSlope,
+  type LieSlopeForward,
+  type LieSlopeSide,
   type LieType,
   type Shot,
 } from '@oga/core'
@@ -56,6 +59,8 @@ interface ShotRowMin {
   club: string | null
   lie_type: LieType | null
   lie_slope: LieSlope | null
+  lie_slope_forward: LieSlopeForward | null
+  lie_slope_side: LieSlopeSide | null
   shot_result: string | null
   penalty: boolean
   ob: boolean
@@ -77,6 +82,8 @@ function rowToShot(r: ShotRowMin): Shot {
     club: (r.club as Shot['club']) ?? undefined,
     lieType: r.lie_type ?? undefined,
     lieSlope: r.lie_slope ?? undefined,
+    lieSlopeForward: r.lie_slope_forward ?? undefined,
+    lieSlopeSide: r.lie_slope_side ?? undefined,
     shotResult: (r.shot_result as Shot['shotResult']) ?? undefined,
     penalty: r.penalty,
     ob: r.ob,
@@ -90,7 +97,12 @@ export default function Patterns() {
   const { unit, toDisplay } = useUnits()
   const [club, setClub] = useState<Club>('7i')
   const [lieType, setLieType] = useState<LieType | typeof ANY>(ANY)
-  const [lieSlope, setLieSlope] = useState<LieSlope | typeof ANY>(ANY)
+  // Two-axis slope filter, multi-select. Empty array on an axis = "any"
+  // (no constraint). Replaces the old single-select LIE_SLOPES chip row,
+  // which filtered the legacy lie_slope column (null on every shot logged
+  // since the two-axis migration → matched nothing) (#746).
+  const [forwardSel, setForwardSel] = useState<LieSlopeForward[]>([])
+  const [sideSel, setSideSel] = useState<LieSlopeSide[]>([])
   const [shots, setShots] = useState<ShotRowMin[]>([])
   const [loading, setLoading] = useState(false)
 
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(mobile): Shot Patterns lie-slope filter — two-axis multi-select

**Concerns**
- `apps/mobile/app/(app)/patterns.tsx:216-226` — This PR bundles two things: (1) the actual bug fix — `rowToShot` wasn't carrying `lie_slope_forward`/`lie_slope_side` onto `Shot`, which is a ~2-line fix — and (2) a UI redesign from single-select to a brand-new two-axis multi-select control (`MultiChipRow`, ~60 new lines). CONTRIBUTING.md's "One concern per PR" rule gives "UI restyle, bug fix, and new feature" as the exact example of what shouldn't ship together. The minimal fix could have kept the existing single-select `ChipRow` (using core's legacy `lieSlope` param, which already ORs across both axes) and shipped the data-mapping fix alone; multi-select-per-axis is a UX enhancement on top, not required to make "selecting a slope matches shots" true again. I think this is worth a maintainer call on whether to split, since it doubles the surface area a reviewer has to reason about for what's described as a bug-fix PR.
- `apps/mobile/app/(app)/patterns.tsx:676-708` — `MultiChipRow` is extracted into its own function with exactly 2 callers (line 216, 222). CONTRIBUTING.md's 3-caller rule: "Don't extract a helper, hook, or component into its own function or file unless it has three or more callers, or the inline version is genuinely unreadable." The PR description already flags this ("co-located — 2 callers here"), so the tradeoff was considered — and a ~30-line Pressable/View block duplicated twice arguably clears the "genuinely unreadable inline" bar. Flagging since it's a literal 2-caller extraction, but I don't think this alone should block.

**Nits**
- `apps/mobile/app/(app)/patterns.tsx:240-241` — Empty-state text merges forward and side selections into one unlabeled list, e.g. `(uphill, ball above)`. Readable enough, but doesn't indicate which value belongs to which axis. Minor.
- PR title (`fix(mobile): Shot Patterns lie-slope filter — two-axis multi-select`) is a conventional-commit-style label rather than imperative present tense (CONTRIBUTING.md's example: "Add lie-slope filter to patterns"). Low stakes, but worth aligning if the project wants consistency.

- `packages/core/src/shot-patterns.ts:238-245` — `DispersionFilter` only takes single values per axis, so this PR's multi-select AND/OR logic is reimplemented inline in the mobile component (`patterns.tsx:134-142`) rather than in `@oga/core`. If web ever wants the same multi-select UX, extending `lieSlopeForward`/`lieSlopeSide` to accept arrays in core would avoid a second implementation. Not required now — the inline version is small and correct — just noting it as the pure-logic-in-core rule's natural extension point.

---
Looked good: the core bug diagnosis and fix are correct — confirmed `getShotsByClub`'s `PATTERN_SHOT_COLUMNS` already selects `lie_slope_forward`/`lie_slope_side`, so this was genuinely just a missing mapping in `rowToShot`; no `@oga/core` changes so no new-export-needs-tests trigger; no secrets, migrations, or dependency changes in the diff; PR correctly targets `dev`.

---

cner-smith (commented):
## Review — fix(mobile): Shot Patterns lie-slope filter — two-axis multi-select

**Concerns**
- `apps/mobile/app/(app)/patterns.tsx:138-149` — the multi-select membership filter (`forwardSel.includes(...)` / `sideSel.includes(...)`) is pure filtering logic implemented inline in the mobile screen instead of being folded into `packages/core/src/shot-patterns.ts`'s `DispersionFilter` / `filterDispersionByLie`, which already owns single-value lie-slope filtering and is called from both this file and `apps/web/src/hooks/useShotPatterns.ts`. CONTRIBUTING.md: "Web and mobile apps wire UI; they do not own logic... lift them to `@oga/core`." Web's hook currently only supports single-value `lieSlopeForward`/`lieSlopeSide` filters through that same core function, and it has the identical underlying bug (same null legacy column) — if web later grows the equivalent multi-select chips, this exact array-membership logic would need to be reimplemented rather than reused. I think extending `DispersionFilter` to accept an array (or set) per axis and moving the membership check into `filterDispersionByLie` would keep this in one place for both surfaces.

**Nits**
- `apps/mobile/app/(app)/patterns.tsx:657-716` — `MultiChipRow` has 2 callers, one short of CONTRIBUTING.md's 3-caller threshold for extraction. I think this is a defensible exception — duplicating the `Pressable` + accessibility-state block twice inline would hurt readability more than it helps — but noting it since it's a literal miss on the stated rule.

---
Looked good: the bug diagnosis holds up — `PATTERN_SHOT_COLUMNS` in `packages/supabase/src/queries/shots.ts` already selects `lie_slope_forward`/`lie_slope_side`, so the fix really was the missing row-mapping, not a query change; scope stayed to the one stated concern, and the two-axis AND / multi-select filter logic matches the two-axis slope model used elsewhere in the app.

---

cner-smith:
One-concern note: the multi-select two-axis grid was explicitly requested alongside the bug report (user: 'those should be multi select grid shouldn't they?'), so the data-mapping fix and the UX are one coherent unit of work here rather than a restyle riding along. MultiChipRow stays inline-co-located per the 2-caller/unreadable-inline tradeoff. Core-array extension for shared multi-select noted as a future step if web wants the same UX.

PR #771, opened 2026-07-10
fix(learn): glossary audit corrections

cner-smith:
Verified audit corrections to the Glossary article across all three copies (web, mobile, docs mirror).

- **Burn mark grain read inverted** — rewritten in all three files: the burnt, ragged edge of the cup shows the direction the grain grows toward (down-grain); the sharp, healthy edge is up-grain; putts run faster and break toward the burnt side. The "roots exposed" causal story replaced with the correct mechanism (down-grain blades hang over the hole and die off).
- **Spinner (shot) pin advice** — a spin-back shot suits a front pin, not a back pin: "fly the ball past the hole and let the spin bring it back." All three files.
- **Spinner (putting) deleted** — unsourced as a lip-out term and a second, conflicting definition of Spinner in the same article; Lip out already exists in Slang. All three files.
- **Lede no longer assumes the reader uses OGA** — "are covered in the app's stat reference section" → "are not covered here." All three files.
- **docs mirror drift** (docs/learn/glossary.md only): deleted the docs-only Scalp entry under Divots (contradicted the article's own Scalp/Thin/Blade definition), the duplicate Dormie in Historical, the stray "Flush — See short game section above" cross-ref, and the extra Skins sentence about made-for-TV events.
- **Consistency polish**: web intro now uses the shared `Lede` primitive instead of the local body `P` (mobile primitives have no Lede equivalent — left as-is); mobile Thai Spinner aligned to "2025 Texas Children's Houston Open."

Typechecks pass locally (web `pnpm typecheck`, mobile `npm run typecheck`).

Closes #750
Closes #758
Part of #767, #768, #770 (glossary items)

### DIFF
--- a/apps/mobile/components/learn/articles/Glossary.tsx
+++ b/apps/mobile/components/learn/articles/Glossary.tsx
@@ -9,10 +9,9 @@ export function GlossaryArticle() {
         title="Glossary of golf terms."
       />
       <P>
-        The basics — birdie, bogey, par, handicap, GIR — are covered
-        in the app's stat reference section. This glossary focuses
-        on the terms that come up once you start playing regularly
-        and trying to improve.
+        The basics — birdie, bogey, par, handicap, GIR — are not
+        covered here. This glossary focuses on the terms that come
+        up once you start playing regularly and trying to improve.
       </P>
 
       {GLOSSARY_GROUPS.map((group) => (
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — fix(learn): glossary audit corrections

**Concerns**
- `apps/web/src/pages/learn/articles/GlossaryArticle.tsx:29-33`, `apps/mobile/components/learn/articles/Glossary.tsx:12-15` — the lede change ("are covered in the app's stat reference section" → "are not covered here") is justified in the PR body as fixing an assumption that "the reader uses OGA," which makes sense for `docs/learn/glossary.md` (read outside the app, e.g. on GitHub). But the web/mobile copies *are* the in-app article — if the app genuinely has a stat reference section covering birdie/bogey/par/handicap/GIR elsewhere, this removes a valid in-app navigation pointer without a replacement, for readers who by definition are using the app. Unclear from the diff whether that section actually exists; if it does, the two in-app copies may not need this particular rewrite even though the docs mirror does.

**Nits**
- `apps/web/src/pages/learn/articles/GlossaryArticle.tsx:1,29` — swapping the intro `<P>` for the shared `<Lede>` primitive is a drive-by consistency change riding along with the factual corrections. Low risk since it's the same paragraph already being touched, but it's a distinct kind of change (presentation, not correctness) from the rest of the PR's stated scope ("glossary audit corrections").

---
Looked good: all three copies (web, mobile, docs mirror) were updated in lockstep for the shared facts (burn mark mechanism, spinner pin advice, Thai Spinner event name), the docs-only drift fixes (stray Scalp/Dormie/Flush entries, duplicate Spinner definition) are scoped to `docs/learn/glossary.md` only as described, and this is content/copy-only — no security, package-boundary, or test-coverage concerns apply.

---

cner-smith:
Re the lede concern: the "app's stat reference section" does not exist — repo-wide grep finds no such section in either app (`grep -rn "stat reference" apps/` matches only this article). The old copy was a dead pointer as well as an app-assumption, so the rewrite stands for all three copies. Re the Lede-swap nit: that item is explicitly scoped in #770 (consistency batch), which this PR carries as "Part of #770 (glossary items)" — intentional, not a drive-by.

PR #779, opened 2026-07-10
fix(mobile): shot-pattern slope filter falls back to legacy lie_slope

cner-smith:
**Bug (user-reported, on-device):** On the Shot Patterns page, selecting any lie-slope chip shows "no data".

**Root cause:** The two-axis slope filter from #746 reads only `lie_slope_forward` / `lie_slope_side`. In prod, **530 of 531 shots** have those columns null — their slope lives in the legacy single-select `lie_slope` column (which stored *either* a forward *or* a side value, never both). So every selection filtered to zero.

**Fix:** `rowToShot` now derives the axes from `lie_slope` via core's `legacySlopeToAxes()` when the new columns are null — the same pattern core's own shot mapper (`round.ts:294`) already uses. Two lines + one import.

**Verified:** the shot-save path (`useShotActions.ts:157`, `PastHoleShotsSheet.tsx:462`) already writes both axis columns, so new logs are unaffected; only legacy reads needed the fallback. Mobile typecheck green.

**Known limitation (in a code comment):** a legacy shot recorded only one axis, so old shots can't match a forward *and* a side selection at once — unrecoverable from the old single-select data. New two-axis shots match both.

Closes #746 follow-up.

### DIFF
--- a/apps/mobile/app/(app)/patterns.tsx
+++ b/apps/mobile/app/(app)/patterns.tsx
@@ -15,6 +15,7 @@ import {
   dispersionVerdict,
   filterDispersionByLie,
   getAimCorrection,
+  legacySlopeToAxes,
   type Club,
   type DispersionPoint,
   type DispersionStats,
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(mobile): shot-pattern slope filter falls back to legacy lie_slope

**Nits**
- PR title — CONTRIBUTING.md's "Opening a PR" section asks for the title in "imperative present tense" (e.g. "Add lie-slope filter to patterns"). `fix(mobile): shot-pattern slope filter falls back to legacy lie_slope` reads as a description of current behavior rather than a command. Something like "Fall back to legacy lie_slope when new axis columns are null" would fit the convention better. Not blocking.

---
Looked good: reuses the existing, already-tested `legacySlopeToAxes()` from `@oga/core` (`packages/core/src/round.ts:244`) instead of reimplementing the fallback in `apps/mobile` — matches the exact pattern `shotRowToDraft` already uses at `round.ts:302-303`, so no package-boundary or test-coverage gap. Diff is genuinely two lines + one import, matching the "ship the minimum" bar, and stays to one concern.

---

PR #780, opened 2026-07-10
fix(learn): drop misleading swing-plane figure + correct retention curve

cner-smith:
Second, thorough pass over every Learn figure (13 total, all computed from coordinates). **11 verified correct and kept** — only these two needed action:

**1. Remove the SwingVariations `PlaneSpectrum` diagram (web + mobile).** It tried to place "single plane" on an upright↔flat *steepness* axis, but one-plane-vs-two-plane is an independent build distinction — a category error no arrangement of three fanned lines can fix. It also mislabeled which end Moe Norman's single-plane method sits at. The prose already explains the spectrum in words; the figure only subtracted clarity. Orphaned `react-native-svg` imports removed with it.

**2. Correct the `RetentionDiagram` blocked-practice curve (web + mobile).** It ended at y=68 — below its own practice-start of y=44 — depicting blocked practice as making you *worse than baseline*. The contextual-interference finding is that blocked *retention* is lower than random's, not net-negative. Endpoint moved 68 → 52: still clearly below random's y=30 (random wins retention), no longer collapsing below baseline.

Docs mirrors unaffected (decorative figures weren't mirrored). Both typechecks green. The other 11 figures (ball-flight, alignment, putting-gate, impact-face, swing-plane/over-the-top, tempo 3:1, gap-ladder, wedge-loft, difficulty-band, schedule) verified correct by coordinate computation.

### DIFF
--- a/apps/mobile/components/learn/articles/PracticeModes.tsx
+++ b/apps/mobile/components/learn/articles/PracticeModes.tsx
@@ -289,7 +289,7 @@ function RetentionDiagram() {
       <Line x1={24} y1={14} x2={24} y2={86} stroke={C.mute} strokeWidth={1.5} />
       <Line x1={106} y1={18} x2={106} y2={86} stroke={C.line} strokeWidth={1} strokeDasharray="3 3" />
       {/* blocked: high in practice, drops later */}
-      <Polyline points="30,44 106,30 182,68" fill="none" stroke={C.mute} strokeWidth={2} />
+      <Polyline points="30,44 106,30 182,52" fill="none" stroke={C.mute} strokeWidth={2} />
       {/* random: low in practice, best later (accent) */}
       <Polyline points="30,74 106,60 182,30" fill="none" stroke={C.accent} strokeWidth={2} />
       <SvgText x={54} y={100} fontSize={7} fontFamily={FONT.mono} letterSpacing={1} fill={C.mute}>
--- a/apps/mobile/components/learn/articles/SwingVariations.tsx
+++ b/apps/mobile/components/learn/articles/SwingVariations.tsx
@@ -1,12 +1,10 @@
 import { Text, View } from 'react-native'
-import Svg, { Circle, Line, Text as SvgText } from 'react-native-svg'
 import {
   ArticleHeader,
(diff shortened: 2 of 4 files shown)


cner-smith:
## Review — fix(learn): drop misleading swing-plane figure + correct retention curve

**Nits**
- `apps/web/src/pages/learn/articles/SwingVariationsArticle.tsx:541-542` — the two-line comment ("Editorial line-art: the shaft plane from upright to flat...") describing the removed `PlaneSpectrum` component is left behind, now sitting directly above `function Sources()`. The mobile counterpart (`SwingVariations.tsx`) removed the equivalent comment along with the function; this file didn't. Dead comment describing code that no longer exists — should go with the rest of the deletion.
- `apps/web/src/pages/learn/articles/PracticeModesArticle.tsx:333` / `apps/mobile/components/learn/articles/PracticeModes.tsx:292` — the blocked-practice endpoint moves from y=68 to y=52, but the PR description's own definition of "baseline" is the curve's practice-start point (y=44). 52 is still numerically below that (higher y = lower on the chart), so the line still dips under its own starting point at the "later" mark — just by 8px instead of 24px. That's a much smaller, easy-to-miss visual dip rather than a visible collapse, so the fix likely reads fine in practice, but it doesn't fully eliminate the "ends up worse than baseline" case the PR says it's correcting. Worth a second look at whether the intended fix was an endpoint at or above y=44.

- Bundling the `PlaneSpectrum` removal and the `RetentionDiagram` correction in one PR is a bit of a stretch against the "one concern per PR" rule in CONTRIBUTING.md, even though both came out of the same figure-audit pass and are both framed as bug fixes. Not asking for a split here since they're small, mechanically identical across web/mobile, and reviewing them together is easy — but worth keeping in mind if a future audit pass turns up more than two fixes at once.

---
Looked good: `PlaneSpectrum` removal is clean on both platforms — no orphaned `react-native-svg`/`Figure` imports left in the mobile file, and the change is scoped to exactly the figure plus its defining comment (mobile) or just the figure (web, see nit above). No domain-pillar, security, or package-boundary issues — this is a docs/figure-only change.

---

PR #743, opened 2026-07-10
feat(mobile): surface the web companion (Profile copy + Stats rider)

cner-smith:
Mobile users installing from the App Store had no cue that oga.golf is the same account with a fuller dashboard. Two low-key in-app pointers:

- **Profile 'OGA on the web' card** — reframed copy: *'Your rounds sync to a free web dashboard. Sign in at oga.golf with the same account for bigger stats, strokes gained, and shot-pattern charts.'* (was the open-source line). Feature framing only — stays clear of App Review 3.1.1 (no payment).
- **Stats top bar rider** — a muted `oga.golf ↗` above the L5/L10/L20 round buttons, opens the site. Deliberately subtle (40% opacity).

Also (not in this diff — gitignored): a one-line web-companion sentence added to `store-submission.md` §2 for the next App Store / Play listing edit. The Marketing URL already points to oga.golf; the description just never mentioned it.

Pure JS — rides the next OTA. Mobile `tsc --noEmit` clean.

PR #781, opened 2026-07-10
fix(mobile): web-dashboard CTA on Android profile + stats rider polish

cner-smith:
Follow-up to the #743 web-companion pointers, from on-device QA (Android/S23).

**Profile (Android):** the web CTA was gated `Platform.OS === 'ios'` — on Android the card showed only the Ko-fi / GitHub Sponsors donation links, so Android users never saw a pointer to the web dashboard. Now the Android branch shows an 'OGA on the web' block (full-dashboard copy + oga.golf link) **above** the donation links, which are **kept** (divider between). iOS unchanged (still web-only, per App Review 3.1.1).

**Stats rider:** was the bare, barely-visible 'oga.golf ↗'. Now 'Full dashboard → oga.golf ↗' (lead-in), with `lineHeight: 16` + `gap: 10` so the 'g' descenders no longer clip behind the L5/L10/L20 toggle, and `numberOfLines={1}` so it can't wrap into the buttons on a narrow screen.

Mobile typecheck green. Pure-JS → OTA-eligible.

### DIFF
--- a/apps/mobile/app/(app)/profile.tsx
+++ b/apps/mobile/app/(app)/profile.tsx
@@ -477,6 +477,43 @@ export default function ProfileTab() {
             </>
           ) : (
             <>
+              <Text style={{ ...KICKER, marginBottom: 10 }}>OGA on the web</Text>
+              <Text
+                style={[TYPE.body, {
+                  color: '#1C211C',
+                  fontSize: 14,
+                  lineHeight: 20,
+                  marginBottom: 14,
+                }]}
+              >
+                Your rounds sync to a free web dashboard. Sign in at oga.golf
+                with the same account for bigger stats, strokes gained, and
+                shot-pattern charts.
+              </Text>
+              <Pressable
+                accessibilityRole="link"
+                accessibilityLabel="Open the OGA website"
+                onPress={() => Linking.openURL('https://oga.golf')}
+                style={{
+                  borderWidth: 1,
+                  borderColor: '#1F3D2C',
+                  paddingVertical: 12,
+                  alignItems: 'center',
+                  borderRadius: 2,
+                }}
+              >
+                <Text
+                  style={[TYPE.bodyBold, {
+                    color: '#1F3D2C',
+                    fontSize: 13,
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — fix(mobile): web-dashboard CTA on Android profile + stats rider polish

**Concerns**
- `apps/mobile/app/(app)/profile.tsx:441-476` vs `:480-515` — the new Android branch duplicates the iOS branch's entire "OGA on the web" block verbatim: same kicker, same paragraph copy, same `Pressable`/`onPress`/style, 36 lines byte-for-byte identical. CONTRIBUTING.md's 3-caller rule doesn't force extraction with only two call sites, and "two similar lines are fine" — but this is a full clone, not two similar lines. Since the branches now differ only in what follows (iOS: nothing; Android: divider + donation CTAs), hoisting the shared "OGA on the web" JSX above the `Platform.OS === 'ios'` ternary — so the ternary wraps only the donation section — would halve this diff and remove exactly the kind of drift risk (one platform's copy silently diverging from the other's) that this PR exists to fix.
- One-concern-per-PR (CONTRIBUTING.md: "UI restyle, bug fix, and new feature do not belong in the same PR"): this bundles the Android CTA-gating bug fix (`profile.tsx`) with an unrelated visual-clipping fix on a different screen (`stats.tsx:187-199` — gap/lineHeight/`numberOfLines`). Both trace back to the same on-device QA pass on web-dashboard pointers, so this may be intentional, but they're different root causes on different screens.

**Nits**
- PR title uses conventional-commit style (`fix(mobile): ...`) rather than the imperative-present-tense form CONTRIBUTING.md's example shows ("Add lie-slope filter to patterns"). Flagging in case the project doesn't already tolerate this style elsewhere — not a real issue if it does.

---
Looked good: the iOS/Android gating itself is correct now (no donation CTAs slip onto iOS per App Review 3.1.1, Android gets the web pointer it was missing), the `stats.tsx` clipping fix is minimal and scoped to the reported symptom, and there's nothing here touching security, domain pillars, or package boundaries.

---

PR #782, opened 2026-07-10
fix(web): putt-entry data integrity (#663, #666)

cner-smith:
Two small, verified round-detail putting fixes.

**#663 — stale putt columns on putt→non-putt edit.** `ShotEntryModal` gated `putt_distance_result`/`putt_direction_result`/`putt_result` on `isPuttSave` but left `putt_distance_ft`, `putt_slope_pct`, and `green_speed` writing unconditionally — so editing a putt into a non-putt persisted stale putt data hydrated from the old row. Now all three are gated the same way.

**#666 — putting sheet wipes typed input.** The `WebPuttingSheet` re-seed effect depended on `initialDistanceFt`, which recomputes live as the player drags the putt marker/pin. Any drag while the sheet was open re-ran the effect and reset every field mid-entry. Dropped it from the deps — the distance is snapshotted at open/shot-change (it's only a seed; the player types the real value).

Both verified against current code in the bug-status audit. Web-only, typecheck green.

Closes #663
Closes #666

### DIFF
--- a/apps/web/src/components/round/WebPuttingSheet.tsx
+++ b/apps/web/src/components/round/WebPuttingSheet.tsx
@@ -113,6 +113,12 @@ export function WebPuttingSheet({
   // is the right key — opening the sheet for shot 4 then shot 5 (without
   // closing in between) should restart the form from the new shot's
   // seed values, not retain shot 4's typed distance.
+  //
+  // `initialDistanceFt` is deliberately NOT a dependency: it recomputes
+  // live as the player drags the putt marker or pin while the sheet is
+  // open, and re-running this effect on every drag wipes in-progress
+  // typed input. Snapshotting it at open/shot-change is correct — the
+  // distance is only a seed, and the player is now typing the real one (#666).
   useEffect(() => {
     if (!open) return
     const seed = initial ?? null
@@ -134,7 +140,7 @@ export function WebPuttingSheet({
     // eslint can't see the dependency and would request a useCallback.
     // Keeping it inline keeps the seed logic in one place.
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [open, shotNumber, initialDistanceFt, initial, unit])
+  }, [open, shotNumber, initial, unit])
 
   const parsedDist = Number(distanceText)
   const distanceFt = Number.isFinite(parsedDist)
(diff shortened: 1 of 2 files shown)


cner-smith (commented):
## Review — fix(web): putt-entry data integrity (#663, #666)

**Concerns**
- PR bundles two independent bug fixes (#663 in `ShotEntryModal.tsx`, #666 in `WebPuttingSheet.tsx`) with different root causes in unrelated components. CONTRIBUTING.md's "One concern per PR" rule ("UI restyle, bug fix, and new feature do not belong in the same PR") is framed around type-of-change rather than count-of-bugs, so this may be a reasonable read of "same concern" (both are putt-entry data-integrity bugs) — but if a reviewer wants to revert or bisect one fix independently of the other, they're now stuck together. Worth a call on whether these should've been two PRs.

**Nits**
- `apps/web/src/components/rounds/ShotEntryModal.tsx:189` — the fix correctly gates `putt_distance_ft` behind `isPuttSave`, matching the existing pattern used by `putt_distance_result`/`putt_direction_result`. Confirmed `legacyPuttResult` (line 173) already internally gates on `isPuttSave`, so no double-guard issue there.

---
Looked good: both fixes are minimal and directly address their linked issues — the `ShotEntryModal` gating change mirrors the existing pattern exactly (no new abstraction), and the `WebPuttingSheet` dependency-array fix comes with a clear inline comment explaining why `initialDistanceFt` is intentionally excluded, satisfying the eslint-disable comment requirement. No `@oga/core` changes, no migrations, no mobile deps, targets `dev`.

---

cner-smith:
## Review — fix(web): putt-entry data integrity (#663, #666)

**Concerns**
- `apps/web/src/components/rounds/ShotEntryModal.tsx:189-198` + `apps/web/src/components/round/WebPuttingSheet.tsx:118-142` — this PR bundles two independently-rooted bug fixes: #663 (stale putt columns persisted on a putt→non-putt edit) and #666 (drag-triggered re-seed wiping typed input). CONTRIBUTING.md's rule is "One concern per PR. UI restyle, bug fix, and new feature do not belong in the same PR" — it's written around restyle/fix/feature mixing, and doesn't explicitly rule out two small same-subsystem bug fixes together. Given how contained and low-risk both changes are, I think this is a defensible judgment call rather than a clear violation, but flagging it since a stricter reading of the rule would ask for a split.

---
Looked good: both fixes match their stated diagnoses. In `ShotEntryModal.tsx`, gating `putt_distance_ft`/`putt_slope_pct`/`green_speed` behind `isPuttSave` (previously only `putt_distance_result`/`putt_direction_result`/`putt_result` were gated) correctly stops stale putt data from surviving an edit into a non-putt shot. In `WebPuttingSheet.tsx`, dropping `initialDistanceFt` from the re-seed effect's deps is the right fix for the drag-wipes-input race — the effect still snapshots it correctly at open/shot-change, and `unit` (which `feetToInput` closes over) is correctly retained in the deps array, so no stale-closure bug is introduced.

---

PR #783, opened 2026-07-10
fix(web): infer hole count for unmapped 9-hole courses (#727)

cner-smith:
`useRoundData`'s `expectedHoleCount` fell back to a hardcoded 18 whenever a course had no `course_tees` row — so a 9-hole unmapped course got a phantom back nine (holes 10–18 synthesized). The old comment even flagged it as a known follow-up.

Now, when there's no tees par, it infers the count from the round's own `hole_scores` hole numbers (the authoritative per-round set) via core `inferHoleCount`. That helper returns 18 for an empty set **and** for any hole number > 9, so an 18-hole round can never be trimmed — it only returns 9 when every scored hole is ≤ 9. The length-based synthetic-hole detection (which CLAUDE.md warns against regressing) is untouched — only the count derivation changed.

Verified against current code in the bug-status audit. Web-only, typecheck green.

Closes #727

### DIFF
--- a/apps/web/src/pages/rounds/hooks/useRoundData.ts
+++ b/apps/web/src/pages/rounds/hooks/useRoundData.ts
@@ -1,6 +1,6 @@
 import { useMemo } from 'react'
 import type { Database } from '@oga/supabase'
-import { getShotMarkerCategory, type LieType } from '@oga/core'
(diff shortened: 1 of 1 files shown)


cner-smith:

---

PR #784, opened 2026-07-10
release: dev → main (Learn accuracy wave + deep-audit fixes + web/mobile/core bug batch)

cner-smith:
Ships everything integrated on `dev` since the last release to production — the whole Learn accuracy wave, the 2026-07-09 deep-audit fix wave, and the mobile/web/core bug fixes. Mobile already received these via production-channel OTAs (runtime 1.0.1); this release brings the **web app (oga.golf)** up to the same state and closes the `merged-to-dev` backlog.

**Database:** migration `0040` was already applied to the prod DB via CLI before this PR — verified live (RPCs present, `hole_scores.par` present). This merge only catches the `0040` file up in git; no migration runs on merge.

**Late changes in this release:**
- Course **Planner hidden from nav** (#785) — it's a rough first cut (#385); the `/plan` route stays for dev, but users won't see it until the redesign lands.
- **#727 reverted** (#786) — its hole-count inference collapsed in-progress 18-hole rounds to 9 on unmapped courses; restored the safe hardcoded-18 fallback. #727 stays open.

## What's shipping (by theme)
- **Learn accuracy wave** — 21 article/graphic/citation corrections across web + mobile + docs mirrors (#750–#770), all 21 articles mirrored into `docs/learn/`.
- **Deep-audit fix wave (2026-07-09)** — live-round GPS + finalize integrity, SG distance/handicap correctness, sync dedupe, `holes` UPDATE RLS via RPC (#709–#719, #721, #729).
- **Web fixes** — 9-hole scorecard/share correctness, putt-entry data integrity, map-layer callback safety (#715, #719, #663, #666, #691).
- **Core** — SG/putt formatting + shot-summary consolidation, demo-plan drill resolution, screen cache (#601, #672, #606, #599).
- **Mobile** — offline sync hardening, chip wrap, PressableTouch, logger lie re-seed, web-companion pointers, email-confirm PKCE (#224, #654, #655, #740, #705, #713, #714, #716, #717, #712, #711, #746, #704, #509, #405).

## Excluded from auto-close
- **#555** (Road to 1.0 launch checklist) — tracking issue; stays open.
- **#727** — reverted this release; needs a real per-round hole_count.

## Closing

Closes #224
Closes #405
Closes #509
Closes #599
Closes #601
Closes #606
Closes #654
Closes #655
Closes #663
Closes #666
Closes #672
Closes #691
Closes #704
Closes #705
Closes #709
Closes #710
Closes #711
Closes #712
Closes #713
Closes #714
[...]

PR #785, opened 2026-07-10
chore(web): hide Planner from nav until redesign

cner-smith:
The course planner (#385) is a rough first cut and would ship to oga.golf with the dev→main release (#784) as-is. Comments out the sidebar nav entry so it isn't exposed to users yet; the `/plan` route is untouched for continued development/preview.

Re-enable the one nav line when the planner redesign lands (on-map HUD to match live-round — dispersion toggle, tee-arc/approach-circle mode, styled club picker, prominent SG readout, and a notes box).

Web-only, typecheck green.


cner-smith:

---

PR #786, opened 2026-07-10
fix(web): revert #727 hole-count inference (regression — 18-hole rounds collapse to 9)

cner-smith:
**Regression from #727 (#783), caught before the release shipped.** #727 changed `useRoundData`'s `expectedHoleCount` to infer the hole count from the round's `hole_scores` hole numbers when a course has no `course_tees` row. That assumed the scores always span the full round — false mid-play. An in-progress **18-hole** round on an unmapped course that has only scored holes 1–9 inferred **9** and stranded holes 10–18: the live-round scorecard defaulted to 9 holes with no way to add more.

Reverts to the safe hardcoded-18 fallback (restores current prod behavior). The rare case #727 targeted — a 9-hole *unmapped* course over-padding to 18 — is the lesser evil and has always existed on prod.

#727 needs a **real per-round hole_count captured at round creation** (the round's own intent), not inference from partial in-progress scores. Reopening it with that direction.

Web-only, typecheck green.

### DIFF
--- a/apps/web/src/pages/rounds/hooks/useRoundData.ts
+++ b/apps/web/src/pages/rounds/hooks/useRoundData.ts
@@ -1,6 +1,6 @@
 import { useMemo } from 'react'
 import type { Database } from '@oga/supabase'
-import { getShotMarkerCategory, inferHoleCount, type LieType } from '@oga/core'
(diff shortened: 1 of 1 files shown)


cner-smith:

---

PR #787, opened 2026-07-10
fix(web): scroll to top on navigation

cner-smith:
Learn articles (and any long page) opened mid-scroll instead of at the top when navigated to — the app had **no scroll restoration**, and the scroll container is AppShell's `<main>` (not the window), so a plain window-scroll wouldn't even help.

Resets `main`'s scroll to top on `pathname` change, folded into AppShell's existing on-navigate effect. Keyed on pathname only so query-param changes (e.g. round detail's map/scorecard tab) don't yank the scroll. App-wide fix — the root cause was app-wide, Learn just made it obvious.

Web-only, typecheck green.


cner-smith (commented):

---

PR #790, opened 2026-07-10
fix(core): proximity stat falls back to following putt distance (#575)

### LINKED ISSUE #575: Stats: PROXIMITY empty when no pin coordinate is captured
## Symptom

Mobile **Stats → Ball Striking** shows **PROXIMITY** as an em-dash (`—`) while FAIRWAYS / GIR / DRIVE AVG are populated.

## Mechanism (verified)

`ballStrikingStats` (`packages/core/src/stats.ts:464-543`) only accumulates proximity when `category === 'approach'` AND `distance_to_target > NEAR_GREEN_YARDS (30)` AND end coords exist AND the result is `< 100` yd (`stats.ts:512-524`). It calls `getProximityYards` (`stats.ts:31-41`), which **returns null unless a pin coordinate exists** (`hole_scores.pin_lat/lng` or `holes.pin_lat/lng`, `stats.ts:37-39`).

Most prod courses have **no `holes` rows** (synthetic-holes architecture — only OSM-mapped courses do), and per-round pin placement is optional, so the pin coord is usually null → every approach returns null → `proximityAvg` is null → em-dash. The other three ball-striking stats don't need a pin (round-level columns + tee-shot haversine), so they populate while proximity alone goes dark.

## Options

This is "correct given the data," not a math bug. Fix requires either:
- an **alternate target source** (next shot's start coords, or course green centroid) when no pin is placed — a feature; or
- surfacing **"no pin data yet"** in the UI instead of a bare dash.

Low priority — filing so it persists. Likely v1.1.

cner-smith:
**Ball-striking PROXIMITY showed a bare em-dash** for most users. It required a pin coordinate (`hole_scores`/`holes` `pin_lat/lng`), which is null on the majority of prod courses — only OSM-mapped courses have `holes` rows, and per-round pin placement is optional. The other three ball-striking stats don't need a pin, so they populated while proximity alone went dark.

**Fix:** `ballStrikingStats` now falls back — when an approach has no pin-based proximity but was followed by a putt, that putt's distance from the hole *is* how close the approach finished (`putt_distance_ft / 3` → yards). Pin geometry still wins when a pin exists. Also covers approaches logged without end coords (scorecard-only rounds with putt data).

+3 core tests (fallback fires, pin-geometry precedence, null when neither). All 592 core tests green, workspace typecheck green.

Closes #575

### DIFF
--- a/packages/core/src/__tests__/stats.test.ts
+++ b/packages/core/src/__tests__/stats.test.ts
@@ -618,3 +618,65 @@ describe('ballStrikingStats — girPct', () => {
     expect(ballStrikingStats([r]).girPct).toBeNull()
   })
 })
+
+describe('ballStrikingStats — proximityAvg (#575)', () => {
+  const PIN_LAT = 35.4676
+  const PIN_LNG = -97.5164
+
+  it('falls back to the following putt distance when no pin coordinate exists', () => {
+    // Synthetic-hole course: no pin on the hole or hole_score, and the
+    // approach has no end coords — but it was followed by a 15 ft putt, so
+    // proximity = 15 ft / 3 = 5 yd.
+    const r = makeRound({
+      hole_scores: [
+        makeHoleScore({
+          holes: makeHole({ par: 4, pin_lat: null, pin_lng: null }),
+          shots: [
+            makeShot({ shot_number: 1, lie_type: 'tee', distance_to_target: 400 }),
+            makeShot({ shot_number: 2, lie_type: 'fairway', distance_to_target: 150 }),
+            makeShot({ shot_number: 3, lie_type: 'green', putt_distance_ft: 15 }),
+          ],
+        }),
+      ],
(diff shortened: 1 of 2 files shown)


cner-smith:

Checked: fallback logic is scoped to `hs.shots` for a single hole so `shots[i + 1]` can't leak across holes; `getProximityYards` returning `null` (no pin) correctly falls through to the putt-distance path even when end coords exist, matching the PR description; pin geometry still takes precedence when both exist. Change is confined to `packages/core/src/stats.ts`, reuses the existing `isPuttShot` helper instead of duplicating lie-type logic, and ships 3 new tests in the same package covering the fallback, pin-precedence, and null cases — satisfies CONTRIBUTING.md's "any new export/behavior in `@oga/core` needs a test" bar. Single concern (proximity fallback), targets `dev`, no scope creep.

---

PR #793, opened 2026-07-11
fix(crawl): keep nameless OSM golf courses instead of dropping them

cner-smith:
## Problem

The OSM course-discovery pass (`scripts/crawl/osm-fetcher.ts`) dropped every `leisure=golf_course` way that had no `name` tag via `if (!name) continue`. But **~20% of golf-course ways in OSM are nameless** — a mapper drew the outline (often with full hole/green geometry) but never typed a name. Those are real courses.

Concrete miss: **FireLake Golf Course** (Shawnee, OK — Citizen Potawatomi Nation, 18 holes) is fully mapped in OSM as nameless way `676111573`, so it never entered the `courses` table. Its 16 mapped holes also had no course centroid to attach to in the `osm-holes` pass. Double miss, same root cause.

## Fix

Keep nameless courses with a location fallback name:

```js
const name = tags['name'] || `Golf Course${city ? ` (${city})` : ''}`
```

Upsert-on-`external_id` means a later crawl patches the real name in if OSM ever gets one. App course-selection is GPS-proximity based, so a located-but-generically-named course is far more useful than a missing one.

## Validation (Oklahoma, dry run — no writes)

Ran the real `fetchOsmCoursesInState('OK')`:
- **212 courses captured**, **41 previously dropped** as nameless.
- FireLake way `676111573` now captured.

## Notes / follow-ups
- **Backfill still required** to populate prod — re-run `--source osm --force` (states skip when already `done`), then `--source osm-holes --force` to attach orphaned holes.
- **Name-clobber caveat:** `upsertCourse` overwrites `name` unconditionally, so a backfill would rename any manually-corrected row back to the fallback. FireLake was hand-inserted with its real name + `external_id=osm_way_676111573`; guard the update (don't overwrite a real name with a fallback) before running the national backfill.
- Nameless courses won't get OpenGolfAPI enrichment (tee ratings) since that pass is name-search based; they still get course + hole geometry.

### DIFF
--- a/scripts/crawl/osm-fetcher.ts
+++ b/scripts/crawl/osm-fetcher.ts
@@ -1,16 +1,7 @@
 // OSM Overpass — state-level course discovery (centroid only).
-import {
-  OSM_DELAY_MS,
-  OVERPASS_ENDPOINTS,
-  STATE_BBOX,
-  sleep,
-} from './util'
+import { OSM_DELAY_MS, OVERPASS_ENDPOINTS, STATE_BBOX, sleep } from './util'
 import type { OsmCourseLite, OverpassResponse } from './types'
-import {
-  getCrawlState,
-  setCrawlState,
-  upsertCourse,
-} from './db-writer'
+import { getCrawlState, setCrawlState, upsertCourse } from './db-writer'
 
 export async function fetchOsmCoursesInState(state: string): Promise<OsmCourseLite[]> {
   const bbox = STATE_BBOX[state]
@@ -56,8 +47,6 @@ out center tags;
         const out: OsmCourseLite[] = []
         for (const el of data.elements) {
           const tags = el.tags ?? {}
-          const name = tags['name']
-          if (!name) continue
           let lat: number | undefined
           let lng: number | undefined
           if (el.type === 'node') {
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(crawl): keep nameless OSM golf courses instead of dropping them

**Concerns**
- `scripts/crawl/osm-fetcher.ts:67` — the new fallback name interacts with `upsertCourse` in `scripts/crawl/db-writer.ts:47` and `:57`, which overwrites `name` unconditionally on every re-crawl (`select-then-update`, no diffing of the incoming name against the existing one). The PR description already flags this: FireLake was hand-corrected to its real name at the same `external_id`, and the next `--source osm --force` backfill will regenerate the same fallback (`tags['name']` is still empty in OSM) and stomp the manual correction right back down. Since this is called out as a prerequisite ("guard the update... before running the national backfill") but isn't implemented in this diff, I'd rather see the guard land here — it's a small change (skip the `name` field on update when the incoming name is a fallback and an existing row already has a real one) — or be tracked in an issue that's linked from this PR, rather than relying on the PR description being remembered later.
- `scripts/crawl/osm-fetcher.ts:67` — `` `Golf Course${city ? ` (${city})` : ''}` `` produces identical names for every nameless course sharing a city, and just the literal string `"Golf Course"` for any nameless course with no `addr:city` tag. GPS-proximity selection in the app softens the impact, but any UI that lists/searches courses by name will show indistinguishable duplicates until OpenGolfAPI enrichment or a manual fix backfills a real name. Worth a short note on whether that's accepted as-is.

**Nits**
- `scripts/crawl/osm-fetcher.ts:2,4` and the two `console.log` reformats around lines 103 and 149 — collapsing these to single lines is unrelated to the nameless-course fix and just adds diff noise. Minor, but CONTRIBUTING's "ship the minimum" bar suggests keeping unrelated formatting out of a `fix/` PR unless it's a byproduct of a repo-wide Prettier setting change.

---
Looked good: the fix itself is small and precisely scoped to the actual bug (dropped nameless OSM ways), it's keyed off `external_id` so it can't create duplicate rows, and the PR description is upfront about the clobber risk instead of glossing over it.

---

PR #794, opened 2026-07-11
feature(crawl): reverse-geocode course city from coordinates (OSM Nominatim)

cner-smith:
## Why

Most OSM `golf_course` ways carry no `addr:city`, so the discovery pass can only give nameless courses a bare `Golf Course` fallback name (see #793). This adds a way to relate any coordinate-bearing course to a location automatically, so we don't hand-edit cities/names.

## What

New `--source geocode` pass + `reverseGeocodeCity(lat, lng)` (`scripts/crawl/geocoder.ts`):
- For every course with a centroid but **no city**, reverse-geocode via **OSM Nominatim** and fill `courses.city`.
- Bare `Golf Course` fallback names → `Golf Course (City)`. **Real names are left untouched** (guards manually-corrected rows like the hand-inserted `FireLake Golf Course`).
- Nominatim policy respected: 1 req/sec + identifying User-Agent, 429 backoff, one retry. `--limit N` caps a run.
- Table-wide (not per-state), so it needs no bbox.

```
tsx scripts/crawl-courses.ts --source geocode [--limit N]
```

## Validation (dry run, no DB writes)

Real `reverseGeocodeCity`:
- `35.3726675, -96.9697245` (FireLake) → **Shawnee** ✓
- `36.5674, -121.9487` (Pebble Beach) → **Del Monte Forest** ✓ (correct CDP)

Found during testing: `zoom=10` collapses Nominatim to county-level and drops the city — the query uses default zoom.

Typecheck clean (`tsc --noEmit` on the changed scripts).

## Notes
- Pairs with #793 (keep nameless courses): after that backfill creates `Golf Course` rows, this pass gives them real city labels.
- Only fills `city` (already-known 2-letter `state` is untouched — Nominatim returns full state names).

### DIFF
--- a/scripts/crawl-courses.ts
+++ b/scripts/crawl-courses.ts
@@ -41,6 +41,7 @@ import { countCourses, fetchAllCrawlState } from './crawl/db-writer'
 import { crawlOsm } from './crawl/osm-fetcher'
 import { crawlOsmHoles } from './crawl/holes-fetcher'
 import { crawlEnrich, crawlOpenGolfApi } from './crawl/enricher'
+import { crawlGeocode } from './crawl/geocoder'
 import type { Args, Source } from './crawl/types'
 
 function parseArgs(argv: string[]): Args {
@@ -54,7 +55,7 @@ function parseArgs(argv: string[]): Args {
     const a = argv[i]
     const next = argv[i + 1]
     if (a === '--source' && next) {
-      const allowed = ['opengolfapi', 'osm', 'osm-first', 'enrich', 'osm-holes'] as const
+      const allowed = ['opengolfapi', 'osm', 'osm-first', 'enrich', 'osm-holes', 'geocode'] as const
       if (!allowed.includes(next as Source)) {
         throw new Error(`--source must be one of ${allowed.join(', ')} (got: ${next})`)
       }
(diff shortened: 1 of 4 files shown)


cner-smith:
## Review — feature(crawl): reverse-geocode course city from coordinates (OSM Nominatim)

**Concerns**
- `scripts/crawl/geocoder.ts:70` — `c.name.startsWith('Golf Course')` is a prefix match, not an exact match against the fallback patterns (`Golf Course` or `Golf Course (${city})`) that `osm-fetcher.ts` actually produces. Any course whose *real* name happens to start with "Golf Course" (e.g. "Golf Course at the Bluffs") and has no city yet would get renamed to `Golf Course (City)`, silently dropping the rest of its real name. That contradicts the PR description's "Real names are left untouched" guarantee. Matching the exact fallback string(s) instead of a prefix would close this.

**Nits**
- `scripts/crawl/geocoder.ts:27-31` — the 429 branch retries indefinitely (30s sleep, `continue`, no attempt counter), unlike the `MAX_ATTEMPTS = 2` cap used for non-429 failures a few lines below. The PR description says "429 backoff, one retry," but as written a sustained rate-limit/ban would loop forever rather than bail after a bounded number of attempts. Probably fine for a manually-run backfill script, but worth either capping it to match the description or updating the description.

---
Looked good: scope is held to the one stated concern (reverse-geocode + backfill city), the new `fetchCitylessCourses`/`updateCourseCity` follow the existing `db-writer.ts` conventions, the Nominatim rate limit (1 req/sec, identifying User-Agent, `--limit` cap) is implemented as described, and this is a `scripts/`-only change so it doesn't raise the RLS/package-boundary/`@oga/core`-test concerns that apply to app code.

---

cner-smith:
Superseded — all of this branch's changes (reverse-geocode) are contained in #796, which merged to dev (including the anchored isFallbackName + 429-cap fixes). Closing as merged-via-#796.

PR #795, opened 2026-07-11
feature(web): 'how'd it go?' hole summary — tickers, detail chips, on-demand putt aimer (#791)

cner-smith:
Refines the web `HoleReviewSheet` into the end-of-hole **"how'd it go?"** summary (epic #791), so the web deferred-metadata model becomes the shared target the mobile port follows.

## What's here
- **Summary strip** — three uniform editable tickers (Score · vs-par / Putts / Penalties), all with ± steppers. Penalties is a new hole-level field.
- **Full-height sheet** — fills the map area above the HUD (z-index 40); no overlays bleed through.
- **Shot detail chips (Version A, inline-expand)** — each shot row has club / lie / slope (two-axis) / result chips. Blank chips read `+ field` (dashed); tapping unfolds the options in place. Club chip capitalized to match the Sentence-case lie/slope/result labels.
- **Putt rows** — Made it + short/long · miss-side axes, plus **break** (line + slope) and **green-speed** chips, and a **Read ▸** chip that brings up the full-sheet `GreenDiagram` aimer on demand. Replaces the auto-popping putting sheet.

## Data
- Migration `0041` adds `hole_scores.penalties integer not null default 0` (already applied to dev + prod; idempotent `add column if not exists`, so the release `db push` is a no-op).
- `ReviewedShotRow` gains `shotResult` / `lieSlopeForward` / `lieSlopeSide`; `saveReviewedHole` persists them. Break/speed/aim columns were already wired.

## Verification
- `pnpm typecheck` clean; `pnpm --filter @oga/core test` green (592).
- Every state rendered locally and screenshotted before shipping (summary, chip expand, putt break/speed, Read aimer).

Mobile port to match is the next step, in lock-step.

### DIFF
--- a/apps/web/src/components/round/HoleReviewSheet.tsx
+++ b/apps/web/src/components/round/HoleReviewSheet.tsx
@@ -12,7 +12,7 @@ import {
   formatClubLabel,
   haversineYards,
   horizontalBreakFromAim,
-  isPuttShot,
+  isPuttEntry,
   type BreakDirectionHorizontal,
   type BreakDirectionVertical,
   type Club,
@@ -190,7 +190,7 @@ export function HoleReviewSheet({
       })
     setRows(merged)
     setScore(merged.length)
-    setPutts(merged.filter((r) => isPuttShot(r.lieType)).length)
+    setPutts(merged.filter((r) => isPuttEntry(r.lieType, r.club)).length)
     setPenalties(0)
   }, [open, holeNumber, par, pinLat, pinLng])
 
@@ -520,7 +520,7 @@ function ShotRow({
   // Putt-ness is the green lie only (set when a putt is placed). Raw
   // distance must NOT classify: a chip/bunker inside 30 yd is not a putt,
   // and unmapped rows read distanceToPin 0 (#660).
-  const isPutt = isPuttShot(row.lieType)
+  const isPutt = isPuttEntry(row.lieType, row.club)
   const { toDisplay, toDisplayFt } = useUnits()
   const { bag } = useUserBag()
   // Source the club options from the user's bag, falling back to
@@ -615,6 +615,12 @@ function ShotRow({
       {isPutt ? (
         <>
           <div style={{ display: 'flex', flexWrap: 'wrap', gap: 6 }}>
+            <FieldChip
+              label={clubLabel}
+              filled
+              active={open === 'club'}
+              onClick={() => toggle('club')}
+            />
             <button
               type="button"
               onClick={() =>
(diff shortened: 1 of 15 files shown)


cner-smith:
## Review — feature(web): 'how'd it go?' hole summary — tickers, detail chips, on-demand putt aimer (#791)

**Concerns**
- `apps/web/src/components/round/HoleReviewSheet.tsx:585` — `hasRead = row.aimOffsetInches != null && row.aimOffsetInches !== 0` treats an explicit dead-straight read (offset `0`) the same as "never opened the aimer." A player who drags to the center and hits "Save read" gets a chip that still reads `Read ▸` unfilled, even though `aimOffsetInches: 0` is saved correctly. Data isn't lost, but the UI can't distinguish "explicit straight read" from "no read entered," which is the kind of ambiguity the "aim point is always explicit user input" pillar cares about getting right. A sentinel other than `0` (e.g. tracking "touched" separately) would close the gap.

**Nits**
- `apps/web/src/pages/rounds/hooks/useRoundActions.ts:576-579,622-624` — `saveReviewedHole`'s `summary` param is optional with fallbacks to `rows.length` / `puttCount` / `0`. `HoleReviewSheet` (the only caller, via `RoundDetailPage.tsx:437`) always passes `{score, putts, penalties}` now, so the fallback path is dead code. CONTRIBUTING's "no parameters added for hypothetical future use" would suggest making `summary` required and dropping `puttCount`, unless there's a second caller planned that isn't in this diff.
- `apps/web/src/components/round/HoleReviewSheet.tsx:678` (`SlopeExpand`) and `:723` (`BreakExpand`) — each has exactly one call site, under the letter of the 3-caller rule. Given `ShotRow` is already a very large function, pulling these two-axis grids out for readability seems like the intended exception, but flagging since the rule is explicit about the caller count.

- `slopeLabel` (`HoleReviewSheet.tsx`) is a generic snake_case→Sentence case formatter that would be reusable if/when the mobile port (mentioned as "next step" in the PR description) needs the same chip labels — worth lifting to `@oga/core` at that point rather than duplicating.

---
Looked good: no secrets, no new tables (so no missing-RLS issue — this only adds a defaulted column to `hole_scores` via idempotent `add column if not exists`), migration `0041` is the correct next sequential number, the on-demand aimer stays genuinely on-demand (never auto-opens, aim is user-dragged via the unchanged `GreenDiagram`), and the `ReviewedShotRow`/`SHOT_RESULT_LABELS`/`LIE_SLOPES_*` symbols this PR consumes were already-tested `@oga/core` exports rather than new, untested ones.

---

cner-smith:
Thanks for the review. Triage:

- **Offset-0 `hasRead` ambiguity:** valid — the read *value* is saved correctly (`aimOffsetInches: 0`), only the Read chip's filled/unfilled display can't tell "explicit dead-straight" from "never opened." No data loss; deferring as a small display follow-up (needs a separate touched flag).
- **Optional `summary` param / `puttCount` fallback:** the param is genuinely used (not hypothetical) and the fallback is a sensible default; leaving it optional.
- **`SlopeExpand`/`BreakExpand` single-caller:** the intended readability exception — `ShotRow` is very large (the "genuinely unreadable" carve-out).
- **`slopeLabel` → @oga/core:** good call for the mobile port; folded into the shared-logic follow-up (https://github.com/cner-smith/opengolfapp/issues/798).

No code changes needed on this PR — merging as-is.

PR #796, opened 2026-07-11
feature(crawl): course completion pass — containment holes + confident dedup

cner-smith:
**Stacked on #794** (uses no code from it directly anymore, but shares crawler files — base on #794 to keep the diff clean; retarget to `dev` if #794 merges first).

## What

New `--source complete` pass — a per-OSM-polygon "derive what's missing" checklist:

| step | how |
|---|---|
| **holes** | **polygon containment** (point-in-polygon), each hole assigned to exactly one polygon |
| **par/yards/tee/pin** | from the contained `golf=hole` ways (reuses `buildOrientedHole`) |
| **dedup** | non-OSM centroid inside a polygon → auto-merge **only on close name match** (Jaccard of significant tokens); tees + rounds **moved**, dup holes/row deleted |
| **flag** | every non-matching in-polygon duplicate → printed for human review, **never auto-deleted** |

`--dry-run` reports without writing. Idempotent + resumable. **Zero Nominatim calls** — city-filling stays in `--source geocode` so this scales across all states.

## Why containment

Nearest-centroid (the old `osm-holes` assignment) is a guess that bleeds holes between adjacent courses. Concretely: an OpenGolfAPI phantom 172 m from FireLake stole 5 of its 18 holes. Containment is exact — a hole inside the boundary belongs to that course.

## Why name-gated dedup

Pure containment over-merges. The OK dry-run caught both failure modes:
- **Elks** — same course as FireLake but a totally different name → correctly **flagged**, not blind-merged.
- **Pawhuska → Ponca City** — a real course with wrong OpenGolfAPI coords lands inside an unrelated polygon → correctly **flagged**, not deleted.

So: auto-fix what's exact (holes), surface what's judgment (dedup).

## Validation (OK, prod)

- **FireLake: 13 → 18 holes** with real pars (3/4/5), phantom Elks flagged.
- 186 polygons, **45 confident merges** (tees preserved), **10 flagged** for review.
- Hardened mid-run: a real `course_tees(course_id,tee_color)` unique collision surfaced → merge now drops the duplicate's colliding tee color (real course wins) before moving the rest.

## Run

```
tsx scripts/crawl-courses.ts --source complete [--states OK] [--force] [--dry-run]
```

National flow: `--source complete --force` (all states, fast) → `--source geocode` (cities, rate-limited). Review the accumulated flag list once.

## Notes / follow-ups
- Relation (multipolygon) golf courses are counted + skipped in v1 (26 in OK); they still get holes via the nearest-centroid `osm-holes` pass. Handle if common.
- A 27/36-hole course mapped as ONE polygon keeps holes 1–18 (DB CHECK); multi-course clubs mapped as separate polygons are handled correctly (a containment strength).

### DIFF
--- a/scripts/crawl-courses.ts
+++ b/scripts/crawl-courses.ts
@@ -43,6 +43,7 @@ import { crawlOsmHoles } from './crawl/holes-fetcher'
 import { crawlEnrich, crawlOpenGolfApi } from './crawl/enricher'
 import { crawlComplete } from './crawl/completer'
 import { crawlGeocode } from './crawl/geocoder'
+import { crawlReconcile } from './crawl/reconciler'
 import type { Args, Source } from './crawl/types'
 
 function parseArgs(argv: string[]): Args {
@@ -53,6 +54,7 @@ function parseArgs(argv: string[]): Args {
   let limit: number | null = null
   let maxCourses: number | null = null
   let dryRun = false
+  let apply = false
   for (let i = 0; i < argv.length; i++) {
     const a = argv[i]
     const next = argv[i + 1]
@@ -65,6 +67,7 @@ function parseArgs(argv: string[]): Args {
         'osm-holes',
         'geocode',
         'complete',
+        'reconcile',
       ] as const
       if (!allowed.includes(next as Source)) {
         throw new Error(`--source must be one of ${allowed.join(', ')} (got: ${next})`)
@@ -81,6 +84,8 @@ function parseArgs(argv: string[]): Args {
       force = true
     } else if (a === '--dry-run') {
       dryRun = true
+    } else if (a === '--apply') {
+      apply = true
     } else if (a === '--status') {
       status = true
     } else if (a === '--limit' && next) {
@@ -95,7 +100,7 @@ function parseArgs(argv: string[]): Args {
       i++
     }
   }
-  return { source, states, force, status, limit, maxCourses, dryRun }
+  return { source, states, force, status, limit, maxCourses, dryRun, apply }
 }
 
 async function showStatus(): Promise<void> {
@@ -137,6 +142,7 @@ async function main(): Promise<void> {
         '  tsx scripts/crawl-courses.ts --source enrich [--states TX,OK] [--force]\n' +
         '  tsx scripts/crawl-courses.ts --source opengolfapi [--states TX,OK] [--force]\n' +
         '  tsx scripts/crawl-courses.ts --source geocode [--limit N]\n' +
+        '  tsx scripts/crawl-courses.ts --source reconcile [--states RI] [--apply]\n' +
         '  tsx scripts/crawl-courses.ts --status',
     )
     process.exit(1)
@@ -154,6 +160,16 @@ async function main(): Promise<void> {
   if (args.source === 'geocode') {
     console.log(`Geocode crawl${args.limit != null ? ` (limit ${args.limit})` : ''}`)
     await crawlGeocode(args.limit)
+    return
+  }
+
+  // Reconcile queries the courses table by state string (no bbox needed).
+  if (args.source === 'reconcile') {
+    const states = args.states ?? Object.keys(STATE_BBOX)
+    console.log(
+      `Reconcile (exact-dup): ${states.length} state(s)${args.apply ? ' [APPLY]' : ' [dry-run]'}`,
+    )
+    await crawlReconcile(states, args.apply)
     return
   }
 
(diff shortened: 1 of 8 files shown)


cner-smith (commented):
## Review — feature(crawl): course completion pass — containment holes + confident dedup

**Blockers**
- `scripts/crawl/db-writer.ts:241-276` (`mergeAndDeletePhantom`) — the merge doesn't guard against two `NO ACTION` FKs that this same codebase already hit and documented in `supabase/migrations/0038_dedup_courses.sql:29-39`: `rounds.course_tee_id → course_tees` and `hole_scores.hole_id → holes` are both `NO ACTION`. If the phantom being merged has a round that scored a hole (`hole_scores`) or was played off a tee color that collides with the real course's (the `dropDup` branch at :256-264), the corresponding delete (`course_tees` :257-264 or `holes` :271) throws a raw FK-violation and aborts `crawlComplete` mid-state — *after* `rounds.course_id` was already repointed to the real course (:242-248), leaving that round pointing at a course whose tee/hole rows no longer match it. Migration 0038 solved exactly this by guarding the delete on `rounds.course_tee_id` citations and running a manual preflight specifically for `hole_scores_on_loser_holes`, noting it "can't be inferred from `rounds_on_losers` alone" (0038:129-135). None of that is replicated here, and unlike 0038 (a one-off migration with a manual preflight), this runs unattended inside an automated crawl. The OK prod run in the description probably didn't trip this because those 45 merged phantoms had no play history yet, but the planned national `--force` run is exactly the scale where a course with real rounds is more likely to hit it.

**Concerns**
- `scripts/crawl/completer.ts:246-250` — `existingCourse.name.startsWith('Golf Course')` is a prefix match used to decide whether an existing DB name is "real" (kept) vs. a fallback (replaced with a freshly-generated `Golf Course (City)`). Same pattern already flagged on #794's `geocoder.ts:70`: a genuine name that happens to start with "Golf Course" (e.g. "Golf Course at the Bluffs") on a way with no OSM `name` tag would get silently replaced. Narrower here since `poly.name` wins whenever OSM does have a name, but the path is live for any nameless-OSM-way course this pass touches.
- This PR targets `fix/course-reverse-geocode`, not `dev`. CONTRIBUTING says "Open a PR against `dev` (not `main`)." The description explains the stacking rationale (shared crawler files, retarget once #794 merges), which is reasonable, but CONTRIBUTING doesn't carve out an exception for stacked branches, so flagging per the letter of the rule.

**Nits**
- `scripts/crawl/holes-fetcher.ts:188-210` (`buildOrientedHole`) — now exported for two call sites (`buildHolesForCourses` here, `completer.ts`). Under the letter of the 3-caller rule this doesn't clear the bar, though sharing ~20 lines of tee/pin-orientation math instead of duplicating it across two files seems like the intended kind of exception.
- `scripts/crawl/completer.ts` — `existing` (all courses in the state) is fetched once before the polygon loop and never refreshed as merges happen. If two polygons' `pointInPolygon` checks both match the same phantom centroid (only plausible for adjacent/overlapping boundaries), the second merge would operate on an already-deleted course id. Likely rare given real-world geometry, but worth a one-line comment if it's a known non-issue.

---
Looked good: scope stays to the stated completion pass (containment holes + gated dedup), `--dry-run` reports without writing and is a real no-op path, the Jaccard name-match gate is conservative and flags rather than deletes on any ambiguity, `pointInPolygon` is a standard ray-cast with a correct comment on the edge cases it resolves, and this is a `scripts/`-only change so it doesn't raise the RLS / `@oga/core`-boundary / test-coverage concerns that apply to app code.

---

cner-smith:
## Review — feature(crawl): course completion pass — containment holes + confident dedup

**Concerns**
- `scripts/crawl/completer.ts:100-102` — the comment on `fetchStateGeom` admits this is the "3rd copy of this endpoint loop (osm-fetcher, holes-fetcher, here)" and defers extraction to "if a 4th caller appears." CONTRIBUTING's 3-caller rule is "don't extract... unless it has three or more callers" — three copies is exactly the trigger, not a reason to wait for a fourth. Since this is called out explicitly rather than just missed, it reads like a deliberate deferral; worth pulling the Overpass POST+backoff loop into a shared `overpassPost(query)` in `util.ts` now, or explaining why this instance is the readability exception.

**Nits**
- `scripts/crawl/db-writer.ts:152-163` — the `fetchCitylessCourses` type annotation and error-push lines are reformatted (single-line → multi-line) with no functional change. Unrelated to this PR's stated concern (completion pass); adds diff noise to a `feature/` PR that otherwise touches `db-writer.ts` only to add new exports.

---
Looked good: no new migrations or tables (so no RLS gap), `mergeAndDeletePhantom` orders its writes so a failure can't lose data (rounds → tees, with a real-course-wins collision guard on `course_tees(course_id,tee_color)` → holes → course), dedup is confidence-gated (Jaccard name match) with non-matches only flagged for human review rather than auto-deleted, `pointInPolygon`/`buildOrientedHole` are correctly scoped as script-local (not `@oga/core` — this is crawler tooling, not shared app logic), and `--dry-run` mirrors the real write path closely enough to trust its output.

---

Issue #226, opened 2026-05-07
Launch: hole_scores updates not durable offline on mobile


cner-smith:
hole_scores.score/putts/pin_lat/pin_lng updates during live round go direct to Supabase with no SQLite queue. If player logs entire round in airplane mode and closes without ending, server-side scores sit at last successful update.

completeRound() reconciles on end-of-round so the full flow works — but a player who quits mid-round offline loses partial scores.

Fix: queue hole_scores patches through same SQLite-backed sync pipeline as shots, or show 'X holes not synced' banner.

Found in launch readiness audit, May 2026.

cner-smith:
## Validation 2026-05-22

**Verdict: real bug — confirmed.** Verified against `dev`.

### Where the hole lives

Every `hole_scores` mutation during a live round bypasses the SQLite pending queue and hits Supabase directly:

- `apps/mobile/components/round/hole/useShotActions.ts:192-201` — `score` + `putts` update fires after every shot save.
- `apps/mobile/components/round/hole/useShotActions.ts:250-256` — `pin_lat` + `pin_lng` write on pin placement.
- `apps/mobile/components/round/hole/useShotActions.ts:269-276` — `pin_lat` + `pin_lng` clear on pin removal.

All three patterns:
```ts
supabase.from('hole_scores').update({ ... }).eq('id', payload.hole_score_id).then(...)
```

No queue, no retry. The `then` handler logs and moves on. By contrast, **shots** go through `lib/db.ts → lib/sync.ts` and survive offline (modulo #224). `hole_scores` updates simply vanish on a failed network call — the local React state updates optimistically (e.g. `setHoleScores` at line 202), so the player sees the score change on the screen but the server never receives it.

### The mitigation that already exists isn't enough

`apps/mobile/lib/completeRound.ts:96-101` re-reads shots and **upserts inferred fairway/GIR + SG into `hole_scores`** at end-of-round. That fixes derived columns, but it does NOT recover lost `score` / `putts` / `pin_lat` / `pin_lng` writes — those were never queued anywhere, and `completeRound` reads them straight back from the (stale) Supabase row.

So the abandon-mid-round-offline scenario in the issue body is correct: 8 holes of putts logged in airplane mode → app force-quit before end-of-round → server still has zeros for putts/score on those 8 holes. The data is gone.

### Root-cause reasoning

Two design slips made this happen:

[...]

PR #797, opened 2026-07-11
feature(mobile): end-of-hole review — location-now, details-at-EOH (#791)

cner-smith:
Mobile port of the web "how'd it go?" hole summary (#791), in lock-step with #795. **Stacked on #795** (base = `feature/web-hole-summary-v2`) — merge that first.

## What changes for the player
During a live round, each shot is now logged **location-only** — mark ball → set/skip aim → it saves and loops straight to the next ball. No per-shot detail modal, no on-green "are you putting?" prompt. At **Finish hole**, the ported `HoleReviewSheet` opens: confirm every shot's club / lie / result + putt read (break / speed / on-demand aimer), edit score/putts/penalties, save → advance. Details are captured once, at the end, instead of interrupting pace of play.

## How the save reconciles (the careful part)
Shots are already durably queued per-shot (`pending_shots`, client-generated id, idempotent sync upsert on `id`). The EOH save **attaches metadata to those existing shots** rather than recreating them: each reviewed row pairs back to its live shot by `shot_number`, so the client id + live-captured aim carry through, and the merged payload is re-queued → the next sync **updates the same server row**.

This deliberately differs from web's delete-then-reinsert: on mobile there's no offline delete queue, so delete-then-reinsert would strand **duplicate server rows** whenever the save happens offline. The id-preserving upsert is the offline-safe equivalent. Field-for-field the `shots`/`hole_scores` writes are identical to web's for the same input (verified by a parity audit).

## Review
Ran a 3-agent review (parity / defects+silent-failures / cross-feature seams) + a falsify pass that tried to disprove each finding. Of 11 consolidated claims: 5 phantom (existing guards defeat them), 4 narrowed, 1 real. Fixes shipped in `056d278`:

- **Never fabricate a colliding shot id** — if a reviewed row can't pair to an existing shot (both reads came up empty), the save **aborts loudly and retries** instead of minting a fresh uuid that would 23505 past the sync's `onConflict:'id'` and get silently quarantined. The sheet stays open with edits intact.
- **One-retry on both pairing reads** (mirrors `useHoleData`) so a transient blip doesn't strand a shot on the abort path.
- **`upsertReviewedShot` excludes `broken` rows** from its id lookup.
- **Putt tally counts green-lie shots** (`isPuttShot`) to match the SG putting engine + putt-count readers — web + mobile sheets + web save fallback. The per-row putt-UI gate stays `isPuttEntry` (green + putter), so a bladed-wedge-on-green shows normal-shot UI but still counts as a putt on the scorecard.

## Known limitation (accepted — #226)
The `hole_scores` tally write (incl. **penalties**) is not offline-queued — an offline review save's penalties don't reach the server. Same pre-existing gap as score/putts (score self-repairs at round finalize; penalties don't). Tracked in #226.

## Follow-up (not in this PR)
`ShotLogger` / `PuttingSheet` / the on-green handlers are now unreached in the live flow but left **inert** (minimal, low-risk diff). Removing them is a clean follow-up once this is validated on-device.

## Verification
- mobile typecheck ✓, web typecheck ✓, core 596 tests ✓.
- The sheet's visuals were rendered/screenshotted previously; this PR is state-machine wiring + a native SQLite save path. **Not device-tested** — needs on-device QA.

## On-device QA checklist
- [ ] Play a hole, finish → summary opens with the right shots in order
- [ ] Confirm club/lie/result + a putt's break/speed/aimer → save → advances
- [ ] Reopen the round: stored shots carry the metadata; score/putts/penalties correct
- [ ] Green-lie shot logged with a wedge → shows normal-shot UI, still counts as a putt in the tally
- [ ] Flaky/offline finish → save either succeeds or fails loudly + retries (no silent loss, no duplicate shots)
- [ ] Last hole finish → round finalizes with correct totals/SG (metadata synced before completeRound)
- [ ] "Edit on map" from the summary → returns to the live map, GPS re-arms

### DIFF
--- a/apps/mobile/components/round/HoleReviewSheet.tsx
+++ b/apps/mobile/components/round/HoleReviewSheet.tsx
@@ -258,6 +258,13 @@ export function HoleReviewSheet({
           <PressableTouch
             accessibilityRole="button"
             accessibilityLabel="Edit shots on the map"
+            // Disabled while a save is in flight: onEditOnMap drops roundState
+            // to PLACE_BALL and closes the sheet, but an in-flight saveHoleSummary
+            // (async) still finishes and would then advanceAfterHole() out from
+            // under the player — mirror the Save button's guard so the two can't
+            // race.
+            accessibilityState={{ disabled: saving }}
+            disabled={saving}
             onPress={onEditOnMap}
             style={{
               borderWidth: 1,
@@ -265,6 +272,7 @@ export function HoleReviewSheet({
               borderRadius: 2,
               paddingVertical: 12,
               paddingHorizontal: 16,
+              opacity: saving ? 0.4 : 1,
             }}
           >
             <Text
--- a/apps/mobile/components/round/HoleReviewSheet.tsx
+++ b/apps/mobile/components/round/HoleReviewSheet.tsx
@@ -13,7 +13,7 @@ import {
   combinedBreakDirection,
   formatClubLabel,
   horizontalBreakFromAim,
-  isPuttShot,
+  isPuttEntry,
   type BreakDirectionHorizontal,
   type BreakDirectionVertical,
   type Club,
@@ -138,7 +138,7 @@ export function HoleReviewSheet({
     const next = initialRowsRef.current
     setRows(next)
     setScore(next.length)
-    setPutts(next.filter((r) => isPuttShot(r.lieType)).length)
+    setPutts(next.filter((r) => isPuttEntry(r.lieType, r.club)).length)
     setPenalties(0)
   }, [visible, holeNumber])
 
@@ -409,7 +409,7 @@ function ShotRow({
 }) {
   // Putt-ness is the green lie only (set when a putt is placed). Raw distance
   // must NOT classify — a chip/bunker inside 30 yd is not a putt (#660).
-  const isPutt = isPuttShot(row.lieType)
+  const isPutt = isPuttEntry(row.lieType, row.club)
   const { toDisplay, toDisplayFt } = useUnits()
   const { bag } = useUserBag()
   // Source the club options from the user's bag, falling back to DEFAULT_BAG
@@ -490,6 +490,12 @@ function ShotRow({
       {isPutt ? (
         <>
           <View style={{ flexDirection: 'row', flexWrap: 'wrap', gap: 6 }}>
+            <FieldChip
+              label={clubLabel}
+              filled
+              active={open === 'club'}
+              onPress={() => toggle('club')}
+            />
             <PressableTouch
               accessibilityRole="button"
               accessibilityLabel="Made the putt"
@@ -538,6 +544,17 @@ function ShotRow({
             />
             <FieldChip label="Read ▸" filled={hasRead} onPress={onOpenAimer} />
           </View>
+          {open === 'club' && (
+            <ChipExpand
+              label="Club"
+              options={clubOptions}
+              value={row.club}
+              onSelect={(v) => {
+                onChange({ ...row, club: v as Club })
+                setOpen(null)
+              }}
+            />
+          )}
           {open === 'break' && <BreakExpand row={row} onChange={onChange} />}
           {open === 'speed' && (
             <ChipExpand
--- a/apps/mobile/components/round/HoleReviewSheet.tsx
+++ b/apps/mobile/components/round/HoleReviewSheet.tsx
@@ -14,6 +14,7 @@ import {
   formatClubLabel,
   horizontalBreakFromAim,
   isPuttEntry,
+  isPuttShot,
   type BreakDirectionHorizontal,
   type BreakDirectionVertical,
   type Club,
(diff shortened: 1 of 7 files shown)


cner-smith:
## Review — feature(mobile): end-of-hole review — location-now, details-at-EOH (#791)

**Blockers**
- `apps/mobile/components/round/HoleReviewSheet.tsx:264-284` (Edit on map button) + `apps/mobile/components/round/hole/useShotActions.ts:562-563` — "Edit on map" has no `disabled={saving}` guard, unlike "Save & next hole" right next to it (`disabled={saving || rows.length === 0}`). `saveHoleSummary` is async (paired reads with retries, then a loop of `upsertReviewedShot` awaits, then a `hole_scores` update) and isn't cancelled by the sheet closing — only re-entrancy is guarded via `saveSummaryInFlightRef`. If a player taps "Save & next hole" then, before it resolves, taps "Edit on map" (which drops `roundState` to `PLACE_BALL` and closes the sheet), the in-flight save still completes and unconditionally runs `setRoundState('PLACE_BALL')` + `advanceAfterHole()` — forcing the player onto the next hole out from under whatever they're doing, and a shot placed in that window risks persisting against the wrong `hole_score_id`. Suggest disabling "Edit on map" while `saving`, and/or having `saveHoleSummary`'s success path no-op if `roundState` is no longer `SUMMARY`.

**Concerns**
- `apps/mobile/components/round/hole/useShotActions.ts:654-707` — the row→payload mapping (`distance_to_target`, `lie_slope_forward/side`, `putt_result`, `aim_offset_yards`, and the `isPuttEntry`-driven field nulling) reads like the "shared business logic" CONTRIBUTING.md says belongs in `packages/core/src/` rather than `apps/mobile/lib`-adjacent code ("Web and mobile apps wire UI; they do not own logic"). The PR description says this "mirrors the web review sheet's replace-all write," which suggests the same mapping is duplicated in `apps/web` (not in this diff, so I can't confirm the exact overlap) — if so the two apps can silently drift on which fields get nulled for putts vs. non-putts. Worth checking whether this can be centralized as a pure `@oga/core` function shared by both apps.
- `apps/mobile/components/round/hole/useShotActions.ts` / `apps/mobile/components/round/hole/types.ts` — I don't see any remaining path that sets `roundState` to `SHOT_DETAIL`/`PUTTING` or opens the `onGreen` dialog, matching the PR's own note that `ShotLogger`/`PuttingSheet`/on-green handlers are now unreached but left "inert." `RoundState` still includes those values and the imports/`swapPuttingToShot` remain in `useShotActions.ts`'s return object. I can't see `LiveRoundSession.tsx` in this diff to confirm whether the JSX is still mounted — if so, that's a real chunk of dead UI/state shipping in this PR. The PR description frames removal as a deliberate fast-follow, which seems reasonable, but flagging since CONTRIBUTING's "ship the minimum" bar leans toward not shipping unreachable code even short-term.
- `apps/mobile/components/round/hole/useShotActions.ts:745` — `// eslint-disable-next-line no-console` has no comment explaining *why*. CONTRIBUTING.md: "No `// eslint-disable-next-line` without a comment explaining why."
- `apps/mobile/components/round/hole/useShotActions.ts:634-707` — the reviewed-row upsert loop awaits `upsertReviewedShot` sequentially per row; if a later row fails to pair (`!existing?.id`), the save throws mid-loop after earlier rows in the *same* save have already been written. Recoverable via the idempotent retry the PR describes, but the "abort" isn't atomic — worth being aware some rows can already be committed by the time the sheet reports failure.

**Nits**
- `apps/mobile/components/round/hole/useShotActions.ts:~654` — `putt_slope_pct: isPuttRow ? row.puttSlopePct ?? null : null` reads a `ReviewedShotRow.puttSlopePct` I don't see any control setting in `HoleReviewSheet.tsx` (only `breakDirectionVertical`/`greenSpeed`/`aimOffsetInches` look user-editable there). Possibly intentional (reserved for parity with web's `buildInitialRows`), but worth confirming it isn't a vestigial no-op on mobile.

---
Looked good: the "never fabricate a colliding shot id — abort loudly and retry" pairing logic, the aim-value carry-through (live-captured aim is never re-derived; skip-aim genuinely nulls rather than defaulting to something), and the `isPuttShot` (tally) vs. `isPuttEntry` (per-row UI gate) split are all implemented as the PR description claims and consistent between mobile and web — no auto-aim or synthesized aim values found anywhere in the reconciliation path.

---

cner-smith:
Thanks — addressed in `580719e`:

- **BLOCKER (save/edit race):** ✅ "Edit on map" is now `disabled={saving}` (+ dimmed + a11y state), mirroring the Save button — the in-flight `saveHoleSummary` can no longer be interrupted by an Edit-on-map that would `advanceAfterHole()` out from under the player.
- **eslint-disable rationale (line ~745):** ✅ added the why-comment.

Deferred with rationale:
- **Row→payload mapping into @oga/core:** agreed it's the right home. Deferring to a tracked follow-up (https://github.com/cner-smith/opengolfapp/issues/798) rather than riding this merge — a dedicated parity pass verified web/mobile are byte-for-byte identical today, and it's a cross-app refactor that deserves its own PR + test.
- **Inert ShotLogger/PuttingSheet/on-green:** deliberate, documented — kept the risky wiring diff small; the cleanup is a fast-follow once this is validated on-device.
- **Non-atomic abort loop:** acknowledged; the partial writes are idempotent on the client id, so a retry re-upserts them harmlessly (confirmed in the review's falsify pass) — self-healing, not data loss.
- **`putt_slope_pct`:** intentional parity placeholder with the shared `ReviewedShotRow`/`buildInitialRows`; no mobile control sets it yet so it writes `null`. Harmless, kept for web parity.

PR #800, opened 2026-07-11
feature(#791): pre-round capture-mode picker — just-track vs track-patterns (step 3)

cner-smith:
**#791 live-round redesign — Step 3.** Pre-round capture-mode picker, web + mobile in lock-step.

> [!WARNING]
> **Blocks-prod: apply migration `0042` to prod BEFORE (or atomically with) the web deploy + mobile OTA.**
> `ROUND_COLUMNS` now selects `capture_mode`, so `getRounds` / `getRound` / `getRoundsWithDetails` (rounds list, round detail, stats) reference the column, and the live-round insert paths write it. Against a prod DB without `0042`: those reads 400 ("column does not exist") and live-round creation 400s (past rounds are unaffected — they omit the column). Mobile reads via `select('*')` degrade gracefully to the `track_patterns` fallback, but **writes and explicit-column reads hard-fail** — so the migration must land first. Dev is already consistent (0042 applied via MCP).

New per-round column `rounds.capture_mode` (`'just_track' | 'track_patterns'`, default `'track_patterns'`) gates aim capture in the live flow:

- **`track_patterns`** (default) — unchanged behavior: an aim is spawned/prompted per shot for the dispersion dataset.
- **`just_track`** — records ball locations only (score + GPS distances). No aim step. Putt-ness, club, lie, and result are still filled in at the end-of-hole review.

### Changes
- **DB:** migration `0042_round_capture_mode.sql` — applied to **dev** (MCP); prod push deferred to the dev→main release (see warning above). Column added to `ROUND_COLUMNS` + hand-added to generated types (matches the 0040 precedent).
- **Core:** `CAPTURE_MODES`, `CaptureMode`, `CAPTURE_MODE_LABELS` in `constants.ts`.
- **Web:** Live-only picker in `NewRoundPage` writes `capture_mode`; `RoundDetailPage` reads it and passes to `useRoundActions`, which skips the aim spawn/prompt in `just_track`. `ModeChip` content top-aligned so uneven-length chips line up.
- **Mobile:** Live-only picker in `RoundSetupStep` writes `capture_mode`; `LiveRoundSession` threads it to `useShotActions`, whose `markBallHere` saves the location immediately and skips `SET_AIM` in `just_track`. `persistShot`/`buildPayload` gained an explicit `ball` arg so the immediate save doesn't read the pre-update `ball` state out of the closure.

The picker shows only for **Live** rounds on both platforms; past-round entry leaves the column at its default.

### Seam audit (post-implementation)
Traced every seam the diff opened — all sound: web no-aim shot persists (aim null-coalesced at save), mobile immediate-persist numbers correctly (`persistShotInFlightRef` blocks double-fire), the dropped on-green/aim dialogs are inert so `markBallHere` is the sole SET_AIM entry (no double-persist), mobile reads `capture_mode` via `select('*')`, and both platforms share the `@oga/core` labels + default. Only material item = the release-ordering warning above.

### Verification
- `npm run typecheck` (mobile) ✅
- `pnpm test` — 596/596 ✅
- Web picker render verified (Playwright screenshot, dev fixture)

### Needs on-device / browser QA
- Web: New round → Live → picker renders, both modes; just_track live round places shots with no aim UI; track_patterns unchanged.
- Mobile: same, plus confirm just_track `markBallHere` persists each ball and loops back to PLACE_BALL (no SET_AIM), and the end-of-hole review still works. (No RN render path — needs an EAS build / simulator.)

Deferred to the release PR: prod `db push` of 0042 **first**, then the closing keyword.

### DIFF
--- a/apps/mobile/app/(app)/round/[id]/index.tsx
+++ b/apps/mobile/app/(app)/round/[id]/index.tsx
@@ -18,6 +18,7 @@ import {
   pickRoundFocus,
   roundFocusHeadline,
   selectNudgeDrills,
+  type CaptureMode,
   type RoundFocus,
 } from '@oga/core'
 import {
@@ -364,6 +365,7 @@ export default function RoundIndex() {
         // The LiveRoundSession `mode`/isPastMode plumbing is now dead and can
         // be removed in a follow-up cleanup.
         mode="live"
+        captureMode={(round?.capture_mode ?? 'track_patterns') as CaptureMode}
         onHoleChange={syncHoleToUrl}
       />
     )
--- a/apps/mobile/app/(app)/round/new.tsx
+++ b/apps/mobile/app/(app)/round/new.tsx
@@ -12,11 +12,14 @@ import { useLocalSearchParams, useRouter } from 'expo-router'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import * as Location from 'expo-location'
 import {
+  CAPTURE_MODES,
+  CAPTURE_MODE_LABELS,
   formatLocation,
   getOpenGolfApiCourse,
   inferHoleCount,
   searchOpenGolfApi,
   todayLocalDate,
+  type CaptureMode,
   type OpenGolfApiSearchResult,
 } from '@oga/core'
 import {
@@ -182,6 +185,7 @@ export default function NewRound() {
   async function startWith(
     courseId: string,
     tee?: { courseTeeId: string | null; teeColor: string | null },
+    captureMode?: CaptureMode,
   ) {
     if (!user) return
     if (startInFlightRef.current) return
@@ -205,6 +209,9 @@ export default function NewRound() {
         ...(tee?.courseTeeId
           ? { course_tee_id: tee.courseTeeId, tee_color: tee.teeColor }
           : {}),
+        // Capture mode steers the live flow only; past entry keeps the
+        // 'track_patterns' column default.
+        ...(mode === 'live' && captureMode ? { capture_mode: captureMode } : {}),
       })
       if (roundError || !round) throw roundError ?? new Error('Round insert failed')
       createdRoundId = round.id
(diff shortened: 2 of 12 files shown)


cner-smith:
## Review — feature(#791): pre-round capture-mode picker — just-track vs track-patterns (step 3)

**Concerns**
- `packages/core/src/constants.ts:214,237-247` — the new `@oga/core` exports (`CAPTURE_MODES`, `CaptureMode`, `CAPTURE_MODE_LABELS`) have no corresponding test anywhere in the diff. CONTRIBUTING.md: "Any new export added to `@oga/core` must have a corresponding test." There's direct precedent for exactly this shape of export — `barrel.test.ts` already pins `DEFAULT_BAG` (a plain array) specifically to "catch barrel drift" — but neither `CAPTURE_MODES` nor `CAPTURE_MODE_LABELS` were added there or anywhere else. Given these are trivial literals I could see an argument that this falls under "no tests for trivial logic," but the `@oga/core`-export rule reads as unconditional and the barrel test's stated purpose (catching a renamed/dropped re-export) applies just as much here.
- `apps/mobile/components/round/hole/useShotActions.ts:459-462` / `apps/web/src/pages/rounds/hooks/useRoundActions.ts:243-247` — `just_track` mode never captures an aim for any shot in the round. CONTRIBUTING's design-pillar list includes "aim point is always explicit user input" and asks contributors to "say why up front" when a feature is pillar-adjacent. I don't think this actually breaks the pillar — no aim is auto-computed or suggested, it's simply not asked for, and `track_patterns` stays the unchanged default — but since skipping aim entirely is the whole point of the new mode, it seems worth a line in the PR description addressing why this isn't the case the pillar guards against, rather than leaving it for a reviewer to reason through.

---
Looked good: no new table so no RLS gap (column added to the already-RLS'd `rounds` table), migration numbered correctly with a CHECK constraint mirroring `CAPTURE_MODES`, `capture_mode` typed as `string` in `types.ts` matches the existing loose-enum-column precedent (`tee_color`), the picker constants correctly live in `packages/core/src/constants.ts` rather than an app `lib/`, web reuses the existing `ModeChip` instead of a new component, and the `effectiveBall`/`opts.ball` threading in `buildPayload` is a real, correctly-reasoned fix for reading stale `ball` state out of the closure in the immediate-save path.

---

cner-smith:
**Design-pillar note (re: 'aim point is always explicit user input').** `just_track` mode doesn't violate the pillar — it never auto-computes, infers, or suggests an aim; it simply doesn't *ask* for one. `track_patterns` (the default) keeps the explicit-aim flow unchanged. The pillar guards against *fabricating* aim data (aim = pin / fairway-center), which this doesn't do — a just-track round records zero aims, so nothing false enters the dispersion dataset.

PR #801, opened 2026-07-12
feature(#791): green rework — on-green Made/Missed + make-% readout (step 4)

cner-smith:
**#791 live-round redesign — Step 4.** Green rework: Made/Missed as the action + tour make-% readout, opt-in read, web + mobile.

> [!NOTE]
> **Stacked on #800 (Step 3).** Base is `feature/live-capture-mode-picker` so the diff shows Step 4 only. Merge #800 first, then this retargets to `dev`. Step 4's mobile edits build on #800's `markBallHere` changes.

### Decisions (from scoping)
- **Green trigger:** a quiet, non-blocking **"⛳ On the green"** button in the PLACE_BALL chrome (not an auto modal — honors Step 1). Robust where pins are absent, which is most of our courses.
- **Platform:** both — mobile gets the full live green flow; web's EOH putt rows get the make-% readout.
- **Make-%:** tour rate shown now; the player's own "You" blend is deferred (epic's phased approach — needs enough logged putts per band).
- **Opt-in read:** the green leads with distance + make-% + Made/Missed; the read (aim/break/speed) and miss detail stay collapsed behind "Set my read" until asked.

### Changes
- **Core:** `putt-make.ts` — `TOUR_MAKE_PERCENT` (public ShotLink-derived anchors) + `tourMakePercent(ft)`, interpolated via the existing `interpolateBaseline`. +5 tests.
- **Mobile:**
  - "On the green" button → `markBallHere({ toGreen: true })` resolves the ball and enters the `PUTTING` flow.
  - `PuttingSheet` reworked: leads with the make-% readout + Made/Missed action; **the read (green diagram, break, speed) + miss detail collapse behind an opt-in "Set my read →"** (default hidden — pace of play first). Anything skipped stays fillable in the EOH review.
  - **"Holed it →" ends the hole** (`persistPutt` → `finishHole` → EOH summary); a miss loops back to place the next putt.
  - Reuses the inert `PUTTING` state/sheet rather than a rewrite.
- **Web:** `HoleReviewSheet` putt rows show `tour Y%` next to the putt distance (parity), when a distance is known. (Web's read was already opt-in via per-row chips.)

### Deferred (with reason)
- **Player "You" make-%** — phased per the epic; needs enough logged green-lie putts per band to be honest.

### Degrades gracefully
No pin (most unmapped courses) → no ball→pin distance → make-% readout hidden; Made/Missed + score still work.

### Verification
- `npm run typecheck` (mobile) ✅
- `pnpm test` — 601/601 (incl. 5 new make-% tests) ✅

### Needs on-device / browser QA
- Mobile: "On the green" appears in PLACE_BALL; tapping enters the putt flow with the make-% readout (when a pin is set) and just distance + make-% + Made/Missed by default; "Set my read →" reveals the read/detail; "Holed it →" ends the hole → summary; "Missed →" → next putt.
- Web: EOH review putt row shows `tour Y%` at a known distance.


cner-smith:
## Review — feature(#791): green rework — on-green Made/Missed + make-% readout (step 4)

**Blockers**
- `apps/mobile/components/round/hole/useShotActions.ts:438-469` — "⛳ On the green" can be the *first* tap on a hole, and in that case it silently drops the shot that reached the green. `markBallHere`'s end-patch only runs `if (prevLocalId != null)` (line 439), which is `null` at hole start (`lastSavedShotLocalIdRef` hasn't been set yet). When `opts?.toGreen` is true (line 465), the code skips straight to seeding `lieType: 'green'` and `roundState = 'PUTTING'` without ever calling `persistShot` for the shot that carried the ball to the green. If the player holes the putt, `persistPutt` → `persistShot` (line ~315) writes a single row as **shot 1**, and `hole_scores.score` gets set to `shotNumber` (1) — an ace — even though it actually took a drive/approach plus a putt. The preceding shot's club/lie/location is never captured at all.
  This is directly reachable: `MapBottomChrome.tsx:132-134` shows the "On the green" chip any time `p.ball != null || p.hasGps` — there's no gate on `p.totalShotsThisHole > 0` — so it's available (and plausible to tap) as the very first action on a drivable/short par 3, or whenever a player doesn't open the app until they're already on the green.
  Compounding this: `useShotActions.ts:301-323`, the `finishHole()` call after a made putt (line 322) checks `previousShots.length === 0` (line 571, evaluated against the *pre-putt* snapshot). In the first-action-of-the-hole case that's `0`, so `finishHole` takes the "nothing placed, skip review" branch (`advanceAfterHole`, line ~572) instead of opening `SUMMARY` — the miscounted score is finalized with no review screen to catch it. The inline comment asserting "previousShots is ≥1 here — the tee shot precedes any putt" (lines 320-321) doesn't hold for this path.
  Suggest gating the chip on `totalShotsThisHole > 0`, or having the `toGreen` branch persist the incoming shot's start before flipping to `PUTTING` so the stroke that reached the green is never skipped.

**Concerns**
- This PR bundles two things under one diff: the on-green interaction/flow change (button, skip-aim-into-putting, made-ends-hole) and the tour make-% stat readout on both mobile and web. CONTRIBUTING's "one concern per PR" rule is about not mixing restyle/fix/feature — these are arguably one "green rework" feature per the epic, so I won't block on it, but the make-% readout is independently shippable and would have been easier to review split out from the flow-behavior change.

---
Looked good: `tourMakePercent`/`TOUR_MAKE_PERCENT` correctly landed in `packages/core/src/putt-make.ts` (not app-level `lib/`) with tests in the same package, reusing `interpolateBaseline` rather than duplicating interpolation logic; the web/mobile make-% readouts share the same core helper and yards→feet conversion is consistent with the existing putt-distance display; no secrets, new tables, or `any`/unexplained-eslint-disable in the diff.

---

PR #809, opened 2026-07-12
fix(mobile): cap accessibility font scaling at 1.3x (#808)

cner-smith:
Closes #808.

**Bug (found on-device):** with iOS Larger-Text / Display Zoom turned up, every `<Text>` scaled unbounded — the fixed-size live-round HUD overlays overflowed and overlapped into an unusable interface. The app capped font scaling **nowhere**.

**Fix:** global `maxFontSizeMultiplier = 1.3` on `Text`/`TextInput` `defaultProps` in the root layout (runs once at startup). Text still scales for accessibility, bounded to what the layouts absorb.

- **1.3× is a starting point** — needs on-device tuning on a large-text device; easy one-line change if it wants to be tighter/looser.
- `Text.defaultProps` is the standard RN global for this; deprecation-warns under React 19 but applies.
- **Follow-up (separate):** rework the live overlays from absolute/fixed to flex/wrap so they reflow at any scale — the proper long-term fix.

Mobile typecheck green. Needs the sideload build to verify on your mom's phone.

### DIFF
--- a/apps/mobile/app/_layout.tsx
+++ b/apps/mobile/app/_layout.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react'
-import { StyleSheet } from 'react-native'
+import { StyleSheet, Text, TextInput } from 'react-native'
 import { Stack } from 'expo-router'
 import { StatusBar } from 'expo-status-bar'
 import { GestureHandlerRootView } from 'react-native-gesture-handler'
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(mobile): cap accessibility font scaling at 1.3x (#808)

**Blockers**
- `apps/mobile/app/_layout.tsx:30-36` — this fix's entire mechanism is `Text.defaultProps` / `TextInput.defaultProps` mutation, and the PR's own comment justifies it as "deprecation-warns under React 19 but still applies." I don't think that's accurate for this project's pinned versions (`react@19.0.0`, `react-native@0.79.6`, per `apps/mobile/package.json`). React 19's release notes list a "Removed: propTypes and defaultProps for functions" change — for function components (which includes `forwardRef`-wrapped ones, and RN's `Text`/`TextInput` are implemented via `React.forwardRef`), `defaultProps` resolution isn't just deprecated-with-a-warning anymore, it's gone. If that's right, this diff sets `Comp.defaultProps` but React never reads it when creating `<Text>`/`<TextInput>` elements, so `maxFontSizeMultiplier` never actually reaches the native view and the overflow bug in #808 stays unfixed — silently, with no error to signal it. I'm not 100% certain from the diff alone (I can't run the RN app here to confirm at runtime), and it's possible RN 0.79's `Text`/`TextInput` implementation special-cases this somehow, but given the PR description already flags "needs the sideload build to verify," I'd treat verifying that the cap actually takes effect (not just that text visually looks bounded, but that it survives a build with React 19's actual defaultProps behavior) as a precondition for merging, not a follow-up. If it doesn't work, the reliable fix is a thin wrapper component (`AppText`/`AppTextInput`) that hardcodes the prop via a normal JS default parameter, used at the call sites — not a global mutation.

**Nits**
- `apps/mobile/app/_layout.tsx:33-35` — `[Text, TextInput] as unknown as Array<{ defaultProps?: {...} }>` double-casts through `unknown` to let the assignment through. Not literally `any`, but same effect as the rule CONTRIBUTING calls out ("Don't add `any` to silence an error — fix the type") — it exists because RN/React's current types don't model `defaultProps` on these components anymore, which is itself a signal for the concern above rather than just a typing inconvenience to cast around.

---
Looked good: scope is held to the one stated concern (font-scale cap), it's a single small file, no migrations/RLS/`@oga/core`/mobile-dependency-pin surface touched by this change, and the PR description is upfront that the constant needs on-device tuning and that the underlying overlay-layout fix is a separate follow-up.

---

PR #807, opened 2026-07-12
fix(learn): add missing review footer to Strokes Gained + Benchmarks (both platforms)

cner-smith:
Surfaced during on-device QA of the preview build: the **Strokes Gained** and **Benchmarks / Stats** Learn articles lacked the `Last reviewed July 2026` footer that the other 18 articles carry — on both platforms.

- **Web:** `StrokesGainedArticle`, `ReadingStatsArticle` (id=`benchmarks`) → local `Footer()` matching peers.
- **Mobile:** `StrokesGained`, `Benchmarks` → shared `<ArticleFooter>` from primitives.
- **Docs mirrors** already carry `last_reviewed: 2026-07` frontmatter — no change.

Additive/static; typecheck green both platforms. Ships with the dev→main release.

PR #799, opened 2026-07-11
chore: strip crawler WIP that leaked into dev via #795

cner-smith:
#795's squash accidentally carried 7 in-progress course-crawler files into `dev` — they belong to the separate crawler PRs #793/#794/#796, not the #791 hole-summary work. This reverts **only** those 7 files to their pre-#795 state (`9af123b`); no app code touched, `0041` and all summary changes stay. Clears the way for the crawler PRs to merge their authoritative versions.

Issue #209, opened 2026-05-06
Promote text+CHECK columns to Postgres enums for narrow Supabase types


cner-smith:
The regenerated Supabase types (PostgrestVersion 14.5) emit `string | null` for every column whose narrow vocabulary is enforced via a Postgres `CHECK` constraint rather than a real enum type. The TypeScript types lose the union information.

Affected columns and their CHECK migrations:
- `shots.lie_type` (0001) — LieType
- `shots.lie_slope` (0001) — LieSlope (deprecated path)
- `shots.lie_slope_forward` (0003) — LieSlopeForward
- `shots.lie_slope_side` (0003) — LieSlopeSide
- `shots.green_speed` (0005) — GreenSpeed
- `shots.putt_distance_result` (0006) — PuttDistanceResult
- `shots.putt_direction_result` (0006) — PuttDirectionResult
- `shots.break_direction` (0001 + 0009) — BreakDirection
- `shots.putt_result` (0001) — LegacyPuttResult
- `shots.shot_result` (0021) — ShotResult
- `profiles.distance_unit` (0007) — DistanceUnit
- `profiles.skill_level` (0001) — SkillLevel
- `profiles.goal` (0001) — Goal
- `profiles.play_style` (0001) — currently 'casual'|'mixed'|'competitive'
- `profiles.play_frequency` (0020) — currently 4-value union

For now the DB→app boundary uses `as <NarrowType> | null` casts in `apps/web/src/components/rounds/shots/draft.ts`, `apps/web/src/hooks/useUnits.ts`, `packages/core/src/sg.ts`, and `packages/core/src/stats.ts` (see fix/types-regen-narrowing). Casts are runtime-safe because the CHECK constraints reject any value outside the union — the types are just losing information the DB already enforces.

Long-term fix: convert each text+CHECK column to a `CREATE TYPE … AS ENUM` with a non-null/null column. Supabase gen-types maps real Postgres enums to TS string-literal unions, so the casts disappear and the regen is the authoritative source.

Migration plan per column:
1. `CREATE TYPE foo_enum AS ENUM (…)` matching the CHECK list.
2. `ALTER TABLE … ALTER COLUMN col TYPE foo_enum USING col::foo_enum`.
3. Drop the CHECK constraint (now redundant).
4. Regenerate types, drop the cast at every call site.

[...]

PR #810, opened 2026-07-12
Release: #791 live-round Steps 1–4 + proximity fix + a11y/learn fixes

### LINKED ISSUE #575: Stats: PROXIMITY empty when no pin coordinate is captured

cner-smith:
Production release (dev → main). 8 commits.

## Shipping
- **#791 live-round redesign, Steps 1–4** — one-tap capture, end-of-hole "how'd it go?" summary (#795/#797), pre-round capture-mode picker (#800), green rework: on-green Made/Missed + tour make-% + opt-in read (#801). *(Step 5 — alt scorecards — deferred; epic stays open, decomposed into #802–806.)*
- **Proximity stat fallback** when no pin coordinate (#790).
- **Learn:** missing "Last reviewed" footer on Strokes Gained + Benchmarks, both platforms (#807).
- **a11y:** cap accessibility font scaling at 1.3× so large iOS text no longer breaks the live-round overlays (#809).
- Crawler-WIP strip that had leaked into dev (#799).

Closes #575
Closes #808

## Migrations — ALREADY APPLIED TO PROD ✅
`0041` (hole_scores.penalties) + `0042` (rounds.capture_mode) pushed to prod via CLI `db push` before this PR (the release gate — the web build reads `capture_mode` and would 400 without it). Prod verified at `0042`.

## Post-merge
- Vercel auto-deploys web from main.
- Mobile: cut EAS **production** Android build → Play internal track (+ iOS/TestFlight).

## Verification
- CI green on all merged PRs; `pnpm test` 603 core tests; typecheck both platforms.
- Preview build (build `89aafb20`) out for on-device QA of the capture-mode gate, green flow, and the font-scale cap.


cner-smith:
## Review — Release: #791 live-round Steps 1–4 + proximity fix + a11y/learn fixes

**Blockers**
- `apps/mobile/components/round/hole/useShotActions.ts:707-709`, `apps/web/src/pages/rounds/hooks/useRoundActions.ts:667-669` — both new end-of-hole save paths write `shot_result: row.shotResult` (a real, user-selectable value now that the review sheets expose a "Penalty"/"OB" result chip) but hardcode `penalty: false, ob: false` unconditionally. Every other shot-write path in the codebase derives these booleans from the result instead — e.g. `useShotActions.ts:198-199`: `penalty: meta?.shotResult === 'penalty', ob: meta?.shotResult === 'ob'`. `packages/core/src/sg-calculator.ts:129` (`shot.penalty || shot.ob ? -1 : 0`) and `stats.ts:343` compute the Strokes Gained penalty adjustment from these two booleans, not from `shot_result`. So a shot tagged Penalty or OB in the new review UI will show correctly in the UI but silently compute as a normal shot in SG — understating true stroke cost with no error or signal. Fix: derive `penalty`/`ob` from `row.shotResult` in both save paths the same way `buildPayload` does.

**Concerns**
- `apps/mobile/app/_layout.tsx:32-38` — the a11y font-scale cap (#808) is applied via `Text.defaultProps`/`TextInput.defaultProps`. The comment already flags that this "deprecation-warns under React 19 but still applies" — I'm not fully sure that holds for RN's current `Text`/`TextInput` implementation (forwardRef-wrapped in recent versions), where `defaultProps` support was dropped, and I couldn't verify at runtime. If the cap doesn't actually apply, the a11y overflow bug this PR ships to fix would be a silent no-op in production. Worth confirming on-device at max Dynamic Type before relying on it, given the whole point of the change is defeated if it doesn't work.
- `apps/mobile/components/round/hole/useShotActions.ts:770-787` — on a failed `hole_scores` tally write, the comment says `completeRound` "self-repairs score from the shot count," but `apps/mobile/lib/completeRound.ts:98-110` only recomputes `score`, not `putts`/`penalties`. If this update fails, the user's entered putts/penalties are lost with just a console warning (local state already shows them as saved). Low-probability path, but the comment overstates the safety net — consider narrowing it or adding a retry/queue.
- `packages/supabase/src/types.ts:994` — `capture_mode` is typed as bare `string` (generated-types limitation for a `text` + `CHECK` column), forcing an unchecked `as CaptureMode` cast at two call sites (`apps/mobile/app/(app)/round/[id]/index.tsx:368`, `apps/web/src/pages/rounds/RoundDetailPage.tsx:155`) rather than normalizing once. Minor, and a pattern already used elsewhere per #209.

**Nits**
- `apps/mobile/components/round/hole/useShotActions.ts` (`buildPayload`, near the `forceAim`/`aimTouched` line) — the comment `` `false ?? x === false`, so an explicit false wins `` describes correct `??` behavior but reads confusingly; a future editor could misread it as flagging a bug that isn't there.

---
Looked good: `tourMakePercent`/`isPuttEntry` are correctly placed in `@oga/core` with solid new tests (`putt-make.test.ts`, `round.test.ts`, `barrel.test.ts`, `stats.test.ts`), no RLS gap (migrations 0041/0042 only alter existing RLS-covered `rounds`/`hole_scores`), migration numbering is clean, no secrets in the diff, and the explicit-aim-only persistence discipline (never auto-committing an unconfirmed aim point) is carried through consistently into the new `just_track` mode.

---

PR #817, opened 2026-07-19
feature(course-data): course summary columns + needs_remap flag (migration 0043)

cner-smith:
## What

Migration `0043` adds four derived/status columns to `courses`:

- `hole_count`, `total_par`, `total_yards` — derived from `holes` (total_yards null when any hole lacks a yardage)
- `needs_remap boolean` — flags courses whose stored hole count is below their validated real hole count

## Backfill

- **Deterministic summaries** run inside the migration: 9,801 courses got hole_count/total_par, 8,715 got total_yards.
- **needs_remap** set out-of-band from the course-validation dataset (stored hole_count < validated actual_holes, only for courses ≤18 holes so multi-nine facilities are left for their own design): **1,554 flagged** (531 unmapped, 1,023 partial). None are user-round courses.

## State

Already applied to **prod** via `supabase db push` (ledger at 0043; db push is ledger-idempotent so re-pushing is a no-op). This PR lands the migration file + regenerated `packages/supabase/src/types.ts` in the repo so dev/CI carry it too. Types diff is scoped to the four new columns (plus a benign field reorder).

### DIFF
--- a/packages/supabase/src/types.ts
+++ b/packages/supabase/src/types.ts
@@ -117,33 +117,45 @@ export type Database = {
           created_at: string
           created_by: string | null
           external_id: string | null
+          hole_count: number | null
           id: string
           lat: number | null
           lng: number | null
           name: string
+          needs_remap: boolean
           state: string | null
+          total_par: number | null
+          total_yards: number | null
         }
         Insert: {
           city?: string | null
           created_at?: string
           created_by?: string | null
           external_id?: string | null
+          hole_count?: number | null
           id?: string
           lat?: number | null
           lng?: number | null
           name: string
+          needs_remap?: boolean
           state?: string | null
+          total_par?: number | null
+          total_yards?: number | null
         }
         Update: {
           city?: string | null
           created_at?: string
           created_by?: string | null
           external_id?: string | null
+          hole_count?: number | null
           id?: string
           lat?: number | null
           lng?: number | null
           name?: string
+          needs_remap?: boolean
           state?: string | null
+          total_par?: number | null
+          total_yards?: number | null
         }
         Relationships: [
           {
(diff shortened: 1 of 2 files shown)


cner-smith:
## Review — feature(course-data): course summary columns + needs_remap flag (migration 0043)

**Concerns**
- `supabase/migrations/0043_course_summary_columns.sql:14-27` — `total_yards` is explicitly nulled out when any hole is missing a yardage (`yards_missing` filter), but there's no equivalent guard needed for `total_par` since `holes.par` is `NOT NULL` in the schema (confirmed via `packages/supabase/src/types.ts` — `par: number`, not `number | null`) — so `sum(par)` is safe as-is. Not a bug, just flagging that I checked this since the asymmetry between the two columns looks suspicious at first glance.
- The `needs_remap` flag's actual values (1,554 flagged rows) come from an out-of-band process against an external "course-validation dataset" that isn't in this repo — I couldn't find a corresponding script under `scripts/` (checked `scripts/crawl-courses.ts` and `scripts/crawl/`). This migration only adds the column with `default false`; a fresh `supabase db reset` on a contributor's machine or in CI will never populate it, so the flag is permanently `false` outside of prod. That's probably fine as one-off curation, but it's worth a line in the PR description (or a follow-up issue) so future readers of `0043` don't go looking for the logic and assume it's missing.

**Nits**
- PR title ("course summary columns + needs_remap flag") doesn't read as imperative present tense per CONTRIBUTING.md's "Title in imperative present tense" rule — though this matches the style of recent merged PRs (#793–#796), so may just be a stale doc.

---
Looked good: derived columns are computed deterministically in-migration from `holes` with correct null-handling, `add column if not exists` keeps the already-applied-to-prod migration idempotent, no RLS gap (existing table, not user-owned), migration number is the next free slot, and scope is held to a single schema change.


---

PR #820, opened 2026-07-20
feature(facilities): facility-first course picker (Phase 1)

cner-smith:
## What

Phase 1 of multi-nine / facilities support: group courses under a parent **facility** so the round-start picker is facility-first — search "Lake Hefner Golf Club" → pick North or South, instead of hunting individual course names in a flat list. Single-course play is completely unchanged.

Design + plan: `docs/internal/multi-nine-facility-design.md`, `docs/internal/multi-nine-phase1-plan.md` (gitignored).

## Commits

- **Migration `0044`** — `facilities` table (SELECT-only RLS) + `courses.facility_id/unit_name/unit_order`. Already applied to **prod** via `db push`; types regenerated.
- **Query layer** — `searchFacilities` / `getFacilityUnits` / `getFacilitiesByIds`; `facility_id` added to `COURSE_COLUMNS`.
- **Curated-seed backfill** — grouped 3 facilities on prod (Lake Hefner N/S, Earlywine N/S, Bandon Dunes 3 courses); normalized polluted OKC `state`. `rounds.course_id` never touched; existing Lake Hefner rounds intact.
- **Facility-first picker (web + mobile)** — facility → units drill-down. Units stay findable by their own name (matched units re-anchor to their facility via `getFacilitiesByIds`).
- **Round-detail facility context** — header shows "Facility — Unit" when the course belongs to a facility.
- **Fix** — `useCourseSearch.local` keeps all courses for `CoursePlanPage`; the picker reads a new `standalone` field.

## Notes

- Plan was **adversarially reviewed before execution** (14 confirmed findings): fixed a critical search regression (grouped units unfindable by name), replaced a fuzzy backfill clusterer with the curated seed, and added the required type casts.
- Prod migration + backfill are already live (this PR lands the code + carries them into dev-supabase on release).
- Web + mobile typecheck green. Live QA (search "Lake Hefner" → North/South) recommended.
- **Phase 2 (deferred):** pick-2-loops composition for 27/36-hole munis (gated on remapping under-mapped courses).


cner-smith:
## Review — feature(facilities): facility-first course picker (Phase 1)

**Concerns**
- `apps/web/src/hooks/useCourses.ts:68-85` and `apps/mobile/app/(app)/round/new.tsx:207-224` — the facility merge/dedup algorithm (split local rows into standalone vs. facility units, re-anchor units to their facility via `getFacilitiesByIds`, dedup the combined list by `id`) is duplicated verbatim across web and mobile. CONTRIBUTING.md's "Pure logic lives in `@oga/core`" section says web and mobile "wire UI; they do not own logic," and helpers "other surfaces would want to reuse" should be lifted to `@oga/core`. This isn't math/SG/units, so I'm not certain it's a hard violation of that rule's letter, but it's exactly the failure mode the rule exists to prevent: two copies of the same non-trivial logic that will silently drift the next time one side gets a fix the other doesn't. Worth extracting to one shared function (with a test) rather than two hand-synced copies.
- `apps/web/src/components/courses/CourseSearch.tsx:63-66` and `apps/mobile/app/(app)/round/new.tsx:241-245` — `openFacility` discards the error from `getFacilityUnits` (`const { data } = await getFacilityUnits(...)`), unlike the sibling `searchCourses`/`searchFacilities` calls in the same files, which `console.warn` on failure. A failed fetch here just opens the facility onto a silently empty unit list with no error surfaced — may be intentional given errors elsewhere in this flow are already best-effort, but it's inconsistent with the pattern right next to it.

- If the merge helper above does get extracted, `packages/supabase/src/queries/courses.ts` (or `@oga/core` if it should stay UI/DB-agnostic) seems like the natural home given it already owns `searchFacilities`/`getFacilityUnits`/`getFacilitiesByIds`.

---
Looked good: `facilities` RLS is SELECT-only with no write policy (service-role only), consistent with the migration's stated intent; migration `0044` is numbered correctly with an additive-only schema change; the backfill script guards against running against a non-prod project and requires each unit to resolve to exactly one course before writing; and the web/mobile pickers mirror each other closely enough that the UX (and the bug surface) stays consistent between platforms.

---

PR #819, opened 2026-07-20
feature(web): start-round CTAs on the dashboard home

cner-smith:
Fixes a web↔mobile parity gap surfaced testing the mobile shell: the web **Home (`/dashboard`) had no start-round action for returning users** — it only appeared in the zero-rounds empty state, so the only path was the drawer → Rounds → New round. Native Home leads with a prominent CTA pair; this brings web to match.

## What changed
- **Dashboard CTA pair** (`DashboardPage.tsx`), mirroring native `RoundCTAs`: filled forest **"Start live round"** (play icon → `/rounds/new?mode=live`) over an outline **"Log past round"** (plus icon → `/rounds/new?mode=past`), placed right under the masthead. Icons via `@mdi/js` (same glyphs as native).
- **"See all →"** on the Recent rounds section header → `/rounds`, so round history is reachable from Home without opening the menu.
- **`NewRoundPage`** now reads `?mode=live|past` from the URL to seed its initial tab (was hardcoded `past`), so the live CTA actually lands on live.

## Scope notes
- **Resume-round banner deferred:** native shows a resume banner for an in-progress round, but web has no active/in-progress-round query (only `useRecentSG`/`useRounds`). Building that query is out of scope here — filed as a follow-up rather than plumbed for a nice-to-have.
- The **top-bar menu stays for now** — it's still the only home for Learn / My Bag. Removing it belongs to the Phase-2 nav reflow (give those routes native homes first). These CTAs remove the main reason to open it.

## Verification
- `pnpm typecheck` + `pnpm --filter @oga/web build` green.
- **Visual QA = this PR's Vercel preview** (mobile + desktop): CTA pair under the masthead, live CTA opens New Round on the *live* tab, "See all" → /rounds.

Separate from #818 (capture chrome). Do not merge until previewed.

PR #818, opened 2026-07-19
feature(web): live-capture onto the mobile map surface

cner-smith:
Brings the web live-round capture screen onto mobile's on-course **map surface** — one contextual instruction + CTA over the map, mobile HUD-pill radii — without regressing tap-to-place. Base editorial system untouched; the map treatment is codified as a documented per-surface exception in the project design system.

## What changed
- **Mapbox controls → top-left** (were bottom-right; top-right is the expected-strokes HUD) so the chrome can own the bottom.
- **`RoundMapInstructionStrip` → `MapBottomChrome`** (file + `RoundMap` barrel re-export + `MapView` caller).
- **Tap-safe over-map bottom chrome**: instruction card is `pointer-events:none` (taps reach the Mapbox canvas listener); the button row is `pointer-events:auto` and only renders when a button matters, so the live tee-shot tap (bottom-center) is never occluded. Every prior state branch + button preserved.
- **HUD pill radii** bumped to the mobile range (ExpStrokesHud, DotsToggle, OverlayRail container); token de-drift on touched lines.

## Hardening
Plan was put through **two adversarial falsification rounds (10 findings, all fixed)** before implementation — most importantly: tap-to-place is a Mapbox *canvas* click listener, so any `pointer-events:auto` element over it swallows the tap; the pointer-events discipline + gated button row is what keeps the tee tap clear.

## Known, by-design tradeoffs
- Past-round "Set pin" at 0 shots renders over the bottom (single small pill; live path unaffected).
- Once shots exist, the CTA row sits over the lower map — intended map-parity behavior.

## Verification
- `pnpm typecheck` + `pnpm --filter @oga/web build` green.
- Per-task reviews + a final whole-branch review (most-capable model): **READY TO MERGE**, no Critical/Important.
- **Visual QA = this PR's Vercel preview** — check: chrome floats bottom, tee-shot bottom-center tappable, top-left controls + bottom-left logo uncovered, HUD pills rounded, paper pages unchanged, **and at a few viewport heights confirm the wrapped button row doesn't overlap the vertically-centered DotsToggle / OverlayRail on short screens**.

Do not merge until reviewed on the preview.

PR #821, opened 2026-07-20
release: facilities + crawler completion + web map/home + data fixes (dev→main)

cner-smith:
Ships to production:

**Features**
- feature(facilities): facility-first course picker + round-detail context (Phase 1, #820)
- feature(web): dashboard start-round CTAs (#819) + live-capture map parity (#818)
- feature(crawl): course completion pass (#796) + keep nameless OSM courses (#793)

**DB / data (already applied to prod)**
- migration 0044 (facilities) + 0043 (course summary cols) — prod at 0044
- curated-seed facilities (Lake Hefner / Earlywine / Bandon)
- state-mislabel correction: 2,398 courses fixed via point-in-polygon (206 border cases held)

**Known follow-up:** #818 short-viewport button-row/rail overlap (edge case); 206 state border-holds pending review.

Mobile changes (facilities picker) are JS-only → shipped via production OTA.


cner-smith (commented):
## Review — release: facilities + crawler completion + web map/home + data fixes (dev→main)

This is the maintainer-cut `dev → main` release PR, so I'm not flagging "one concern per PR" or "PR targets main" — that's the documented release mechanism (CONTRIBUTING.md: "Maintainers periodically merge `dev → main` via PR to cut a release"). Scoped this review to the net new surface: the facilities feature (migrations 0043/0044 + picker UI) and the crawler-completion scripts.

**Concerns**

- `supabase/migrations/0044_facilities.sql:21-23` — `courses.facility_id` and `courses.unit_name` are added to a table whose existing INSERT policy (`0001_initial_schema.sql`: `"Authenticated users can add courses"`, `with check (auth.uid() is not null)`) has no column restriction, and `facilities_read` (line 17, `using (true)`) makes every facility's id publicly discoverable. Any rate-limited (0027) but otherwise unprivileged authenticated user can insert a `courses` row carrying a real facility's `facility_id` and an arbitrary `unit_name`, and it will render inside that facility's card for every user. `unit_name` also has no length/charset CHECK — `0014_course_security.sql` added exactly this (`courses_name_length`, `courses_name_chars`) for `courses.name` after identifying the same "publicly-readable, user-writable text" exposure. Worth the same treatment here, or restricting `facility_id`/`unit_name` writes to service-role until a moderation path exists (mirrors how `0014` restricted `course_tees` UPDATE to service-role for a comparable reason).

- The PR description claims a production data mutation — "state-mislabel correction: 2,398 courses fixed via point-in-polygon (206 border cases held)" — but I can't find a corresponding script or migration anywhere in this diff (`git diff` against the base shows only `scripts/crawl/util.ts`, `scripts/facilities/backfill-facilities.cjs`, and migrations 0043/0044 as new files; no state-fix script). If this already ran directly against prod outside version control, that may be intentional, but as written this diff doesn't let a reviewer verify a change to 2,398 production rows actually happened the way the description says.

- `apps/mobile/app/(app)/round/new.tsx:234-241` and `apps/web/src/components/courses/CourseSearch.tsx:63-69` — `openFacility` (new in this PR, no prior version to compare against) has no `AbortController`/ordering guard around `getFacilityUnits`. Rapid taps between two facility cards can let a stale response overwrite `units`/`facility` after a newer tap already started, since it's a plain `await` with no signal check. Also, on `error` it only `console.warn`s and still calls `setUnits([])`/`setFacility(f)` — a facility with 0 units due to a fetch failure is visually indistinguishable from one that has no listed units.

**Nits**

- `apps/mobile/app/(app)/round/new.tsx:196,237` — two new `// eslint-disable-next-line no-console` (for `searchFacilities` and the new `openFacility`) copy the file's pre-existing pattern (line 187, which predates this PR) but don't carry their own inline rationale. CONTRIBUTING.md: "No `// eslint-disable-next-line` without a comment explaining why." Low severity since it's following established local convention, not introducing a new pattern.

---
Looked good: `packages/core/src/facilities.ts` (`resolveFacilityResults`) is genuinely pure, correctly placed in `@oga/core`, exported, and covered by `facilities.test.ts`; `0044_facilities.sql` enables RLS with a SELECT-only policy and an explicit comment on why writes are intentionally closed; `scripts/facilities/backfill-facilities.cjs` defaults to dry-run, requires `--apply`, and guards against running against the wrong Supabase project; migration numbering (0043/0044) is sequential with no gaps; no `any` anywhere in the diff; aim-point handling in the map rename (`MapBottomChrome`) stays fully explicit-user-tap-driven with no auto-aim introduced; no paywall/payment gating in the new dashboard CTAs.

---

PR #822, opened 2026-07-20
fix(facilities): lock curated course columns + guard drill-down race (#821 review)

cner-smith:
Two pre-merge hardening fixes surfaced by the #821 release review. Both independently reviewed (security agent on the migration, TypeScript agent on the race fix).

## 1. Migration 0045 — facility write lock (security)

**Hole:** 0044 added `facility_id` / `unit_name` / `unit_order` to `courses`, but the INSERT policy is still `with check (auth.uid() is not null)` with no column restriction, and `facilities` is world-readable — so any authenticated user could insert a `courses` row carrying a real `facility_id` + arbitrary `unit_name` and deface that facility's card app-wide.

**Fix:** `BEFORE INSERT OR UPDATE` trigger blocking non-service-role writes to those three curated columns (mirrors 0027's trigger + service-role bypass), plus `unit_name` length (1–100) and charset (`^[^<>{}]*$`) CHECKs mirroring 0014.

- Client `createCourse` (manual-add + OpenGolfAPI import) never sets these columns → unaffected.
- Backfill script uses service-role → bypass works.
- Forward-only; existing 3 curated facility rows pass the new CHECKs.
- Security review: **GO** dev + prod, no exploitable bypass found.

**Applied to dev + prod after merge, per the migration-channel rule.**

## 2. `openFacility` drill-down race (web + mobile)

`getFacilityUnits` was awaited with no ordering guard — rapid taps between two facility cards could let a stale response overwrite the newer facility's units, and a fetch error looked identical to a facility with zero units.

**Fix:** monotonic `facilityReqRef` token drops stale responses (incl. after Back / query-change, so a late response can't re-open a facility you left); fetch errors now show a message. Web + mobile at parity. Dual typecheck clean.

Closes the concerns raised in the #821 review.

### DIFF
--- a/apps/mobile/app/(app)/round/new.tsx
+++ b/apps/mobile/app/(app)/round/new.tsx
@@ -80,6 +80,10 @@ export default function NewRound() {
   // until the query changes or the user taps Back. Mirrors web CourseSearch.
   const [facility, setFacility] = useState<FacilityRow | null>(null)
   const [units, setUnits] = useState<CourseRow[]>([])
+  const [facilityError, setFacilityError] = useState(false)
+  // Monotonic token so a slow getFacilityUnits response from an earlier
+  // facility tap can't overwrite the units of a facility tapped later.
+  const facilityReqRef = useRef(0)
   const [searching, setSearching] = useState(false)
   const [busy, setBusy] = useState(false)
   const [error, setError] = useState<string | null>(null)
@@ -110,8 +114,12 @@ export default function NewRound() {
   // query could strand the picker inside a stale facility whose units no
   // longer match what was typed. Mirrors web CourseSearch.
   useEffect(() => {
+    // Invalidate any in-flight openFacility so a late response can't re-open
+    // a facility the user has navigated away from.
+    facilityReqRef.current++
     setFacility(null)
     setUnits([])
+    setFacilityError(false)
   }, [query])
 
   // Capture GPS once (best-effort) so manual / API course creation can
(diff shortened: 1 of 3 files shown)


cner-smith (commented):
## Review — fix(facilities): lock curated course columns + guard drill-down race (#821 review)

**Concerns**
- `supabase/migrations/0045_lock_facility_columns.sql:34` — the bypass is `auth.role() = 'service_role' or auth.role() is null`. The `is null` half is new relative to the precedent this migration cites: 0027's `rate_limit_course_inserts` only bypasses on `service_role` and otherwise requires `auth.uid() is not null`. The comment explains the intent (direct Postgres connections / migrations have no JWT context), and that's consistent with how Supabase's `auth.role()` behaves today, but it does mean the entire guard rests on the assumption that every anon/authenticated PostgREST request always carries a role claim. Worth a sentence in the migration comment (or PR description) confirming that assumption was checked against this project's actual PostgREST/GoTrue config, since it's a wider escape hatch than the pattern it says it mirrors.
- This PR bundles two unrelated fixes — the DB security trigger (`supabase/migrations/0045_lock_facility_columns.sql`) and the `openFacility` race/error-handling fix (`apps/web/src/components/courses/CourseSearch.tsx`, `apps/mobile/app/(app)/round/new.tsx`). CONTRIBUTING.md's workflow section: "One concern per PR. UI restyle, bug fix, and new feature do not belong in the same PR." Both are bug fixes surfaced by the same release review, and the PR body says each was independently reviewed, so this may be a deliberate call given the #821 release timing — flagging in case a split is still cheap at this point, since either fix might need to be reverted independently of the other.

**Nits**
- `apps/web/src/components/courses/CourseSearch.tsx:41-85` and `apps/mobile/app/(app)/round/new.tsx:83-258` — the `facilityReqRef` monotonic-token pattern (bump counter, capture reqId, drop stale response, reset on Back/query-change) is duplicated near-verbatim between web and mobile. Doesn't hit the 3-caller threshold from CONTRIBUTING.md yet, so inline is correct per that rule — just flagging for when a third call site shows up.

- The facility view still only renders after `getFacilityUnits` resolves, so there's no loading state between tapping a facility and the units (or error) appearing — pre-existing behavior, not introduced here, but worth a follow-up if it feels unresponsive on slow connections.

---
Looked good: the trigger is the right mechanism (one enforcement point instead of re-deriving the same predicate across the INSERT policy and any future UPDATE policy), it correctly distinguishes INSERT-must-be-null from UPDATE-must-be-unchanged, the `unit_name` length/charset constraints mirror 0014's existing pattern, migration numbering is sequential (0045 after 0044), and the race-guard logic in both `openFacility` implementations correctly handles out-of-order responses regardless of which request resolves first. PR targets `dev` as required.

---

PR #824, opened 2026-07-22
feat(picker): dedup OpenGolfAPI courses against local + merge picker list

cner-smith:
Reworks the course picker so importing from OpenGolfAPI stops re-inserting duplicates of courses we already have, and drops the dev-jargon UI — while keeping live global search for genuinely-new courses. Built from the approved design spec (Cluster A, spec parts 1–3).

## What changed

- **`@oga/core` `course-match.ts`** (new, tested): `normalizeCourseName` + `isProbableSameCourse` — fuzzy name+state match. OpenGolfAPI returns no coordinates, so matching is name+state only, by design.
- **Merge-time dedup** (`useCourseSearch` web + mobile search effect): drop API hits that duplicate a local course (incl. facility units) or facility before rendering. No more "same course twice" in results.
- **Pre-insert backstop** (`useImportApiCourse` web + `startWithApiCourse` mobile): before the raw `createCourse`, fuzzy-match the name against local; on a confident match, reuse the existing row instead of inserting a duplicate. The `external_id` check only caught prior imports of the *same* API id; crawled courses have OSM ids and slipped through.
- **UI**: one merged "Courses" list; the "OpenGolfAPI" and "Already imported" headings are gone.

## Why

Picking the OpenGolfAPI twin of a course we already crawled + cleaned was raw-inserting an uncleaned duplicate, bypassing dedup/state-correction/facility grouping — re-polluting the curated DB. Now dups are caught on both platforms; genuinely-new courses still import (self-identifying via `external_id` + `created_by` both set, for later reconciliation).

## Scope / follow-ups

- **Deferred (separate plan):** the GitHub-Action reconciliation cron + Resend email that sweeps net-new imports. Not in this PR.
- **Known minor (best-effort by design):** if OpenGolfAPI returns a full state name ("Oklahoma") while our DB stores the 2-letter form ("OK"), the states "disagree" and a dup can slip past both guards. Only bites when the API supplies a differently-formatted state (often it's absent → name-only match works); the deferred reconciliation cron is the intended sweeper. Optional later hardening: normalize state to a 2-letter code before comparison.

## Verification

- `pnpm test` — 614 core tests pass (incl. new `course-match` suite).
- `pnpm typecheck` + `pnpm --filter @oga/web build` clean; `apps/mobile` `npm run typecheck` clean.
- Reviewed per-task + a whole-branch review (verdict: ready).

Mobile changes are JS-only → OTA-eligible after merge.


cner-smith (commented):
## Review — feat(picker): dedup OpenGolfAPI courses against local + merge picker list

**Concerns**
- `packages/core/src/course-match.ts:21-33` — `isProbableSameCourse` matches on normalized-name + state only; it never looks at `city`. Golf courses with generic names ("Riverside Golf Club", "Pine Valley", "Lake View") are common enough that two unrelated courses in the same state (different cities) could collide. For the search-time dedup (`apps/web/src/hooks/useCourses.ts:83-89`, `apps/mobile/.../round/new.tsx:225-231`) a false positive just hides a result — low cost. But the same function backstops the **import path** (`apps/web/src/hooks/useCourses.ts:199-205`, `apps/mobile/.../round/new.tsx:403-415`): on a match, the mutation returns the *existing* row instead of creating a new one, so a user importing course B could get silently redirected to unrelated course A's holes/pars/yardages/tees. `city` is already available on both sides (`OpenGolfApiSearchResult.city`, `CourseRow.city`) and would be a cheap additional disambiguator. I don't have visibility into how often same-name/same-state collisions actually occur in the curated dataset, so this may be rarer in practice than it sounds — but the failure mode (silent data misattribution on a user's round) seems worth guarding against given the PR's own stated goal is protecting the "curated DB" from pollution.
- One-concern-per-PR: the diff bundles a bug fix (duplicate-import backstop in `useCourses.ts` / `new.tsx`) with a UI restyle (collapsing the "Already imported" / "OpenGolfAPI" sections into one "Courses" list — `apps/web/src/components/courses/CourseSearch.tsx:205-238`, `apps/mobile/.../round/new.tsx:675-677`). CONTRIBUTING.md is explicit: *"UI restyle, bug fix, and new feature do not belong in the same PR."* The description notes this comes from an approved design spec covering both changes together, so this may already be sanctioned outside the PR itself — flagging in case that context isn't visible to other reviewers.

- Given the concern above, consider requiring `city` agreement (when both sides have one) before treating an import-path match as safe to auto-reuse, even if the search-time dedup stays name+state only.

---
Looked good: `course-match.ts` is correctly placed in `packages/core` (not `apps/web/lib` or `apps/mobile/lib`) with its own test file per the "any new `@oga/core` export needs a test" rule; PR targets `dev`; no secrets, new migrations, or RLS-relevant changes; the state-format mismatch (full name vs. 2-letter code) is already called out honestly in the PR description as a known limitation.

---

PR #826, opened 2026-08-07
fix(mobile): green rework — on-green Made/Missed overlays + auto-detect, capture-mode aware

cner-smith:
## Green rework (live round, mobile)

Reworks the on-green flow so putting is one-tap overlays instead of a blocking sheet:

- **38cfd37** — on-green Made/Missed + make-% as bottom overlays; "Not on the green" escape; `PuttingSheet` modal retired; putt read moves to the end-of-hole summary. Auto-detect restored inside `markBallHere` via `PUTTING_RADIUS_YARDS` (30yd — the real pre-#791 mechanism).
- **62722e5** — phantom-ace guard: never auto-enter putting on a hole's first shot (a par-3 tee shot near the pin no longer pops the overlay).
- **70a59da** — from the adversarial review pass:
  - auto-detect + the escape respect capture mode — `just_track` rounds no longer get Made/Missed forced on them; the escape places the ball instead of stranding the player in an aim flow.
  - Made/Missed disabled during the in-flight save (double-tap can't fire `finishHole` twice).

Review-verified: no double-persist (in-flight ref), make-% distance reads the marked ball, missed→loop→re-enter works, ball survives the escape into aiming, no dangling refs, `hole_scores` tally correct. `ShotLogger` + `swapPuttingToShot` + `closePuttingSheet` are now inert — follow-up cleanup ticket, not this PR.

JS-only → OTA-eligible after merge.

## QA needed before merge (no RN render path in CI; live-round behavior)
- Auto-detect feel at 30yd
- Made → summary / Missed → next shot
- "Not on the green" escape (ball placed, not stranded in aim)
- Par-3 tee shot near the pin does NOT pop the overlay
- `just_track` round: no Made/Missed forced

### DIFF
--- a/apps/mobile/components/round/LiveRoundSession.tsx
+++ b/apps/mobile/components/round/LiveRoundSession.tsx
@@ -248,6 +248,11 @@ export default function LiveRoundSession({
     data.storedPin?.lat,
     data.storedPin?.lng,
   ])
+  // Putt distance in feet, mirrored from HoleModals' ShotLogger puttDistanceFt
+  // calc (yards × 3 = feet). Feeds both the on-green make-% pill and the
+  // persisted putt_distance_ft on Made/Missed.
+  const puttDistanceFt =
+    ballToPinYards != null ? Math.round(ballToPinYards * 3) : null
   // Single-color dispersion dots for the selected club (left-toolbar toggle).
   // Computed only when the dots are shown; sparse clubs → null (no dots).
   const dispersionPoints = useMemo(() => {
@@ -682,8 +687,27 @@ export default function LiveRoundSession({
             finalState.setRoundState('PLACE_BALL')
           }}
           onSkipAim={actions.skipAim}
+          // Auto-detect on-green entry lives inside markBallHere itself
+          // (restored from pre-#791 — see useShotActions comment); this
+          // stays a plain pass-through. The chip below still forces toGreen
+          // explicitly as the fallback.
           onMarkBallHere={actions.markBallHere}
           onOnGreen={() => actions.markBallHere({ toGreen: true })}
+          onGreenActive={finalState.roundState === 'PUTTING'}
+          puttDistanceFt={puttDistanceFt}
+          onPuttMade={() =>
+            actions.persistPutt({
+              puttMade: true,
+              puttDistanceFt: puttDistanceFt ?? undefined,
+            })
+          }
+          onPuttMissed={() =>
+            actions.persistPutt({
+              puttMade: false,
+              puttDistanceFt: puttDistanceFt ?? undefined,
+            })
+          }
+          onNotOnGreen={actions.notOnGreen}
           onAddShot={() => {
             // Opt back into the live append flow on a revisited played hole:
             // re-arm the GPS ball + auto-aim and enter PLACE_BALL (#484).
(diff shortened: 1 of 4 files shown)


cner-smith (commented):
## Review — fix(mobile): green rework — on-green Made/Missed overlays + auto-detect, capture-mode aware

**Concerns**
- `apps/mobile/components/round/MapBottomChrome.tsx:214` — the "Not on the green" `TextChip` has no `disabled={p.saving}` guard, unlike the Made/Missed buttons right above it (`:211-212`). If it's tapped while a Made/Missed save is in flight, `notOnGreen()` (`apps/mobile/components/round/hole/useShotActions.ts:543`) synchronously sets `roundState` to `SET_AIM`, but the in-flight `persistPutt` → `persistShot` success path (and `finishHole()` on a made putt) sets it to `PLACE_BALL`/`SUMMARY` right after — whichever resolves last wins, so the "not on the green" tap can be silently discarded. This isn't a data-corruption risk (the putt itself already persisted with the correct made/missed value), but it's a UI-state race in the same class the file is otherwise careful to guard against (see the `persistShotInFlightRef` / `endInFlightRef` comments a few hundred lines up). Might be worth passing `disabled={p.saving}` here too, unless I'm missing why this one path doesn't need it.
- `apps/mobile/components/round/hole/useShotActions.ts:589-591` (`closePuttingSheet`) and `:598-602` (`swapPuttingToShot`) are now unreachable: their only callers were the `onClosePuttingSheet` / `onSwapPuttingToShot` props into `HoleModals`, both removed in this PR. As a consequence, `loggerOpen` (`apps/mobile/components/round/LiveRoundSession.tsx:92`) can never become `true` again in the live-round flow, so the `<ShotLogger visible={loggerOpen} .../>` mount at `apps/mobile/components/round/hole/HoleModals.tsx:101-103` is dead too. The PR's own comment at `HoleModals.tsx:4-9` calls out the `ShotLogger` half of this and defers it to a follow-up cleanup ticket, which is a reasonable call — but `closePuttingSheet` is dead by the same change and isn't mentioned there, and both functions are still exported from the hook's public return type (`useShotActions.ts:79-80`). Given CONTRIBUTING.md's "ship the minimum, then stop," worth folding into the same follow-up ticket rather than leaving it unflagged.

**Nits**
- `apps/mobile/components/round/LiveRoundSession.tsx:254-255` — `puttDistanceFt = Math.round(ballToPinYards * 3)` is a third copy of the same yards→feet conversion already inline in `HoleModals.tsx:106-108` (ShotLogger's prop) and `useShotActions.ts:752` (`saveHoleSummary`). CONTRIBUTING.md's package-boundaries section puts "units conversions" in `packages/core/src/` — each copy is trivial on its own, but a third instance (the comment at `LiveRoundSession.tsx:251-253` even says it's "mirrored from" the existing one) is a reasonable trigger to lift it into a shared helper.

---
Looked good: no security surface touched (no migrations/edge functions in this diff); the phantom-ace guard (`previousShots.length > 0`) and the `just_track` vs `track_patterns` gating on auto-detect both read correctly against their stated intent; Made/Missed double-tap is genuinely guarded via `persistShotInFlightRef`; PR targets `dev`; no new `@oga/core` exports needing tests, no `any`/undocumented `eslint-disable` in the diff.

---

PR #835, opened 2026-08-09
release: v1.2.0 — Expo SDK 54 (Android 16 / API 36), green rework, learnability layer, auth + picker fixes

cner-smith:
## v1.2.0 — the Play-production release

Ships the SDK 54 platform upgrade (satisfies Google Play's targetSdk-36 requirement, deadline Aug 30) plus the learnability layer, green rework, and a wave of QA fixes. **Native release — new EAS production builds required; OTA runtime moves to 1.2.0.**

### What's shipping (dev → main)
- **#827** — Expo SDK 54 / RN 0.81.5 / React 19.1 / **New Architecture ON** / Reanimated 4 + worklets / Mapbox 10.3.5 / view-shot 5.1.1 / victory-native → hand-rolled RN-SVG charts / NativeWind dropped / origin-scoped Metro resolver / **targetSdk 36**. Built + reviewed via SDD; device-QA'd on Android.
- **#826** — green rework: on-green Made/Missed overlays + auto-detect, capture-mode aware.
- **#830** — live-round bottom chrome consolidated (5 stacked pills → 3 layers, wraps at large font scale).
- **#828** — re-mint Turnstile widget after failed auth (login **and** signup form) — fixes a permanent button-disable.
- **#825** — learnability layer Phase 1: first-run intro tour + contextual "?" help on Stats.
- **#829** — intro-tour card 3 copy corrected to the post-#791 end-of-hole-summary flow.
- **#831** — Stats "?" inline with the web link (header keeps its height).
- **#824** — course picker: dedup OpenGolfAPI vs local + one merged list.

### Release checklist (post-merge)
- [ ] EAS production builds off main — Android AAB + iOS (versionCode auto-increments to 10)
- [ ] **RC-build device QA**: green flow + captcha fail→remint (the two paths never run on-device with SDK 54 together)
- [ ] `eas submit -p android --latest` → Play internal track → **confirm "target API 36" warning clears**
- [ ] `eas submit -p ios --latest` → TestFlight
- [ ] Promote to production (Play + App Store) — the launch
- [ ] Post-launch: lift OTA freeze, delete unused MAPBOX_DOWNLOADS_TOKEN secret

⚠️ **OTA freeze holds** until 1.2.0 is live on both stores — no `eas update` on any channel.


cner-smith (commented):
## Review — release: v1.2.0 — Expo SDK 54 (Android 16 / API 36), green rework, learnability layer, auth + picker fixes

This supersedes my earlier "skipping review" comment — since this PR is now the explicit target of the review trigger, here's a full pass. Treating it as the maintainer release-cut aggregation it says it is (per CONTRIBUTING.md's "Maintainers periodically merge `dev → main` via PR to cut a release"), so I'm not flagging mixed concerns across the bundled sub-PRs as a violation.

**Concerns**
- `packages/core/src/course-match.ts` (`isProbableSameCourse`), called from `apps/web/src/hooks/useCourses.ts:199-205` and the dupe-check block in `apps/mobile/app/(app)/round/new.tsx` — city is treated as non-disqualifying whenever either side is missing it, and `OpenGolfApiSearchResult.city` / `OpenGolfApiCourse.city` (`packages/core/src/opengolfapi.ts:11,35`) are optional in practice. The function's own comment says the city check exists because "generic names (`Riverside Golf Club`) repeat within a state" — that exact scenario reopens whenever OpenGolfAPI returns no city, and `useImportApiCourse` would then silently route a new round into an unrelated existing course's holes/tees instead of creating a new one. Worth a fallback (don't treat missing-city as a match for short/generic names) or at least a log when a fuzzy match fires with no city on either side.
- `apps/mobile/components/round/hole/HoleModals.tsx:4-9` and `useShotActions.ts:598-602` — the `ShotLogger` render path is now dead code. `swapPuttingToShot` (the only caller that sets `loggerOpen(true)`) was reachable solely through `PuttingSheet`'s "Not a putt?" escape, and that live-round `PuttingSheet` modal was deleted in this same PR. No other caller sets `loggerOpen` true, and `PuttingSheet.tsx` (665 lines) is now only reachable transitively through the now-unreachable `ShotLogger.tsx` (657 lines) — roughly 1,300 lines shipping in this release that can never render. The comment acknowledges a "follow-up cleanup ticket, not this change" but doesn't link one — worth filing/linking before merge so it doesn't rot silently.
- `apps/mobile/babel.config.js` — the explicit `react-native-reanimated/plugin` entry (previously commented "must be the LAST plugin entry") was removed with no comment explaining why it's now safe. Likely fine under SDK 54's `babel-preset-expo` auto-adding the Reanimated/worklets transform, but that's exactly the kind of change `tsc --noEmit` (the only mobile CI check, `.github/workflows/ci.yml:48-59`) won't catch — flagging so it's confirmed in RC-build device QA before the EAS production build, per the PR's own checklist.

**Nits**
- `apps/web/src/hooks/useCourses.ts:205` (`return dupe as unknown as CourseRow`) — `searchCourses`/`getFacilitiesByIds` now return a narrower column subset than `CourseRow`/`FacilityRow`, cast away via `as unknown as`. Harmless today since the only consumer (`CourseSearch.tsx:119`) reads just `.id`/`.name`, but the cast promises a full row shape that isn't there — a future caller reading e.g. `.created_by` off this value gets `undefined` at runtime despite TS treating it as guaranteed.
- The API-hit dedup filter (`!localRows.some(isProbableSameCourse) && !facilities.some(...)`) is duplicated verbatim between `useCourses.ts` (web) and `round/new.tsx` (mobile). Only 2 call sites, so it's under CONTRIBUTING.md's 3-caller threshold, but since `isProbableSameCourse` itself was correctly lifted to `@oga/core`, the thin wrapper (e.g. `dedupeApiHits(...)`) would be a natural fit there too instead of hand-copying it per platform.

- `apps/mobile/metro.config.js`'s new `resolveRequest` override (scoping `disableHierarchicalLookup` by `context.originModulePath`) is a reasonable, well-commented fix, but like the babel change it only fails at bundle time, not typecheck — worth explicit attention in RC-build QA.
- Pre-existing `^` ranges on `@supabase/supabase-js` and `react-native-url-polyfill` in `apps/mobile/package.json` weren't touched by this diff, so not a new violation of the exact-pin rule, but stick out now that the rest of the file was converted to exact pins.

---
Looked good: NativeWind was fully removed rather than bumped past the locked `4.1.23` (no leftover `nativewind`/`className=` references anywhere in `apps/mobile`), `react-native-worklets` was added as a deliberate Reanimated-4 peer without resurrecting the already-dropped `react-native-worklets-core`, every previously tilde/caret mobile dependency is now an exact pin, and both new `@oga/core` exports (`course-match.ts`, `helpContent.ts`) ship with real same-package test coverage.

---

PR #836, opened 2026-08-09
feat(rounds): add provider-agnostic round import from JSON

Spedge:
Lets a user paste/upload a JSON payload (e.g. from a hand-written Garmin-to-OGA converter) to pre-populate a round, matching provider data straight onto this app's own schema so the payload doubles as a preview of what gets inserted.

- packages/core/src/import.ts: zod schema + validateImportPayload, plus summarizeImportDataQuality — a non-blocking advisory scan that flags shots missing club/lie_type/distance_to_target and putts whose distance landed in distance_to_target instead of putt_distance_ft (the latter silently zeroes out putting Strokes Gained with no error).
- ImportDataPage.tsx: paste/upload -> validate -> pick course/tee ->
  import, with rollback on partial failure and a warning when holes can't be matched to the selected course instead of silently dropping them.
- Migration 0046: rounds.import_id (scoped per-user) for duplicate-import detection.
- Migration 0047: fixes a from-scratch `supabase db reset` gap where PostgREST roles never got table grants on anything created by the `postgres` role, breaking local dev for every table, not just this feature's.
- Hardening: capped payload array sizes, rejected non-finite numbers, and fixed toUserMessage silently showing "[object Object]" for every Postgres error app-wide (postgrest-js errors aren't real Error instances unless .throwOnError() is used).
- vite.config.ts: dev-only CSP relaxation so local Supabase is reachable under `vite dev` (production CSP is untouched).

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Design/UI change
- [ ] Refactor
- [ ] Docs

## Testing
- [x] pnpm typecheck passes (3/3 workspace packages)
- [x] pnpm test passes (all)
- [x] pnpm --filter web build passes
- [x] Tested in browser at localhost:5173
- [ ] Mobile typecheck passes

## Notes for reviewer
I've got a private repo for the garmin JSON creator - I'm not sure how Garmin feel about me getting _my own data_ but I didn't want to link it whatsoever to this project. Happy to share if you need it, but it didn't take long to generate. Only oddity that caught me was the different measurements between yards and feet for putting, but I can see why we need that for SG. 

Only other thing is - you can access the Import from the "Rounds" page but actually, I wasn't sure if you wanted to adjust the "New Round+" page instead? Not too fussed, open to changing it. 

image

### DIFF
--- a/apps/web/e2e/import-data.flow.spec.ts
+++ b/apps/web/e2e/import-data.flow.spec.ts
@@ -1,0 +1,68 @@
+import { test, expect } from '@playwright/test'
+import { createClient } from '@supabase/supabase-js'
+
+const admin = createClient(
+  process.env.SUPABASE_URL!,
+  process.env.SUPABASE_SERVICE_ROLE_KEY!,
+  { auth: { persistSession: false, autoRefreshToken: false } },
+)
+
+// Same seeded, fully-loaded course new-round-past.flow.spec.ts relies on.
+const COURSE_NAME = 'Pine Ridge Golf Club'
+
+function payload() {
+  return JSON.stringify({
+    course_name: COURSE_NAME,
+    played_at: '2026-06-01',
+    total_score: 9,
+    total_putts: 4,
+    holes: [
+      { number: 1, score: 4, putts: 2 },
+      { number: 2, score: 5, putts: 2 },
+    ],
+  })
+}
+
+let createdRoundId: string | null = null
+
+test.describe('Import from data (signed in, mutating)', () => {
+  test.afterEach(async () => {
+    if (createdRoundId) {
+      // Cascades through hole_scores + shots via FK.
+      await admin.from('rounds').delete().eq('id', createdRoundId)
+      createdRoundId = null
+    }
+  })
+
+  test('paste a valid payload → pick course → import → land on round detail', async ({
+    page,
+  }) => {
+    await page.goto('/rounds/import')
+    await expect(
+      page.getByRole('heading', { name: /Import from data/i }),
+    ).toBeVisible()
+
+    await page.getByPlaceholder('{"played_at"').fill(payload())
+    await expect(page.getByText(/Looks good/i)).toBeVisible()
+
+    await page.getByRole('button', { name: /Continue/i }).click()
+
+    // course_name hint pre-seeds CourseSearch verbatim (ImportDataPage
+    // passes initialQuery={payload.course_name}).
+    await expect(page.getByPlaceholder('Search courses')).toHaveValue(COURSE_NAME)
+    await page
+      .getByRole('button', { name: new RegExp(COURSE_NAME, 'i') })
+      .first()
+      .click()
+    await expect(
+      page.getByText(new RegExp(`Selected: ${COURSE_NAME}`, 'i')),
+    ).toBeVisible()
+
+    await page.getByRole('button', { name: /^Import round$/i }).click()
+
+    await page.waitForURL(/\/rounds\/[0-9a-f-]{36}/i, { timeout: 15_000 })
+    const match = page.url().match(/\/rounds\/([0-9a-f-]{36})/i)
+    expect(match).toBeTruthy()
+    createdRoundId = match![1]
+  })
+})
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -7,7 +7,8 @@
     "dev": "vite",
     "build": "tsc -b && vite build",
     "typecheck": "tsc -b --noEmit",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "test": "vitest run"
   },
   "dependencies": {
     "@dnd-kit/core": "^6.3.1",
@@ -24,7 +25,8 @@
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-router-dom": "^6.26.0",
-    "recharts": "^2.12.7"
+    "recharts": "^2.12.7",
+    "zod": "^3.23.8"
   },
   "devDependencies": {
     "@playwright/test": "^1.60.0",
@@ -36,6 +38,7 @@
     "postcss": "^8.4.39",
     "tailwindcss": "^3.4.7",
     "typescript": "^5.4.0",
-    "vite": "^6.4.2"
+    "vite": "^6.4.2",
+    "vitest": "^1.6.0"
   }
 }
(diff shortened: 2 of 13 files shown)


cner-smith:
## Review — feat(rounds): add provider-agnostic round import from JSON

**Blockers**
- PR base — this PR targets `main`, but CONTRIBUTING.md is explicit: "Open a PR against `dev` (not `main`)" and "Branch from `dev`, never from `main` ... Maintainers periodically merge `dev → main` via PR to cut a release; contributors don't open PRs against `main`." This should be retargeted to `dev` before merge.

**Concerns**
- One-concern-per-PR — this PR bundles the import feature with two changes that aren't scoped to it: migration `supabase/migrations/0047_postgrest_role_default_grants.sql` (grants fixing a from-scratch `db reset` gap that, per the PR description, breaks local dev "for every table, not just this feature's"), and `apps/web/src/lib/errors.ts:117-131` (the `toUserMessage` "[object Object]" fix, which is app-wide, not import-specific). CONTRIBUTING.md: "UI restyle, bug fix, and new feature do not belong in the same PR... PRs that add scope beyond their stated concern will be asked to split." The 0047 grants fix in particular reads as a standalone `fix/` PR — it'd also be independently revertable that way if it regresses something, which the mobile-pinning section of CONTRIBUTING calls out as a reason to isolate infra changes.
- `apps/web/src/pages/rounds/ImportDataPage.tsx:274-326` — `handleImport` writes one hole then its shots sequentially, awaiting each `mutateAsync` in a `for` loop (no batching/`Promise.all`). Bounded by the schema's 36-hole/20-shot caps so it can't run away, and there's a progress indicator, so this looks like a deliberate tradeoff rather than an oversight — flagging in case a bulk-insert path was intended instead.

- The `toUserMessage` fix in `errors.ts` (postgrest-js errors not being real `Error` instances) is web-only; worth checking whether `apps/mobile` has the same bug in its own error-message path.
- `handleImport`'s sequential per-shot writes could become a single batched insert per hole (or an RPC) if import volume ever becomes a complaint.

---
Looked good: `import_id` uniqueness in migration 0046 is correctly scoped to `(user_id, import_id)` rather than global, avoiding cross-account collisions; the 0047 grants are safe in that every existing `public` table already has `enable row level security` set (checked the full migration history), so the blanket grant doesn't open anything RLS wasn't already gating; `packages/core/src/import.ts`'s new exports land in the right package and ship with thorough tests in `import.test.ts`; payload caps (max 20 shots/hole, max 36 holes, `.finite()` on numeric fields) bound a malformed payload; `summarizeImportDataQuality` is correctly non-blocking/advisory and doesn't gate the import; and `aim_lat`/`aim_lng`/`aim_offset_yards` are pass-through fields from the external payload, not app-generated aim suggestions, consistent with "aim point is always explicit user input."

---

cner-smith (commented):
First off — this is really solid for a first pass at a codebase you've only just met, and the core of it is right. I gave it a proper going-over (pulled the schema + RLS to check the scary parts myself), so here's the full picture. None of this is a rejection; it's close.

**What's already good (verified, not just eyeballed):**
- No security holes — every insert is RLS-scoped to the signed-in user, there's no way to write onto someone else's round, no IDOR (good call keeping ids out of the payload), no XSS.
- Every zod enum matches the DB CHECK vocabulary exactly.
- The hardening you mentioned all holds — array caps are enforced before any DB call, non-finite numbers rejected.
- The 0046 `(user_id, import_id)` unique index is a real DB-enforced dedup, and scoping it per-user instead of global was exactly the right instinct.
- `summarizeImportDataQuality` is genuinely useful — the misplaced-putt-distance detector is correct and well tested. You clearly got the yards-vs-feet SG subtlety.

**Blockers — and both are project-specific seams you had no way of knowing about, not anything sloppy:**

1. **Synthetic holes.** Most courses in prod have *no* rows in the `holes` table — only OSM-mapped ones do. `ImportDataPage` reads holes from `useHolesForCourse` and skips any hole with no match, so for a typical course it'll skip all 18 and create an empty round. The rule here is: before any `hole_scores.hole_id` write, call `ensureRealHole(hole)` to materialize the row on demand (that's what `RoundDetailPage` does). The e2e passed only because the seeded course happens to have real holes. This is the one I'd genuinely give you a hand wiring up — it's non-obvious, and it's what makes import actually work everywhere. `docs/internal/synthetic-holes.md` has the full story.

2. **Mobile build.** Adding `zod` to `@oga/core` pulls it into the mobile bundle too (index.ts barrel-exports import.ts), but it's not in `apps/mobile`'s package.json/lock — mobile installs separately from the workspace. CI's mobile typecheck passes because it installs core's deps first, but an EAS build only installs inside `apps/mobile`, so it'd fail on "can't resolve zod." That's exactly why your "mobile typecheck passes" box should stay unchecked. The clean fix, though, isn't to add zod to mobile — mobile has zero references to the import feature, so it'd be carrying a runtime dep for code it never runs. Better to keep the import module in `apps/web`, since it's got a single web-only consumer: move `import.ts` into the web app, import core's *types* type-only, and let zod be a plain web dependency. Core and mobile both stay untouched. (If a mobile import path ever lands, that's the moment to promote it into `@oga/core` and add zod to mobile — not before.)

3. **Rollback.** If an insert fails partway, the compensating `deleteRound(...).catch(() => {})` is best-effort and swallows its own failure — and by then the `import_id` is already consumed by the 0046 index, so a retry just warns "already imported" with no way back. Worth making the round-create the atomic guard and cleaning up harder on failure (or considering whether the whole insert can move server-side into one call).

**Smaller stuff, cheap to fix alongside:**
- The e2e `toHaveValue` asserts `'Pine Ridge'` but the box pre-seeds to the full `'Pine Ridge Golf Club'`, so the spec fails as written.
- zod lets two shots share a `shot_number` on the same hole, but the DB has a unique constraint on `(hole_score_id, shot_number)` — so that payload passes validation and then trips the DB mid-insert (→ the rollback path above).
- The schema allows up to 36 holes for double rounds, but the write path upserts on `(round_id, hole_id)`, so a second entry for hole N overwrites the first rather than being a second nine. Either drop the 36 allowance for now or let's talk about modelling it properly.

**One to pull out of this PR:** migration 0047. I checked prod and it's actually a no-op there (those grants already exist), so it's not dangerous — but it's really a local `supabase db reset` fix, and shipping it as a numbered prod migration that grants `anon` write-by-default on all *future* tables removes a nice fail-closed default. Could you split it into its own PR (or move it to seed)? Happy to take it on its own merits.

**Fast-follows, not blocking:** imported putts don't set the legacy `putt_result` (doesn't affect strokes gained — verified — just leaves a blank result in the summary until edited); and the write flow could use a test or two for the failure/duplicate paths.

And on your question — Rounds page vs the New Round screen — the Rounds-page entry is fine for now, no need to move it.

Genuinely good contribution. Ping me on the synthetic-holes piece and I'll point you at exactly where it slots in.

PR #841, opened 2026-08-15
feature(course-editor): dev-only Course Editor + tee-specific hole data

Spedge:
## Summary
- Adds a developer-only Course Editor (`/dev/courses`, gated on `import.meta.env.DEV` — absent from production builds) for editing course/hole/tee/facility data directly, backed by a local Vite dev-server plugin using the Supabase service-role key server-side. `courses`/`holes` have no UPDATE/DELETE RLS policy today, so this is the only way to correct existing course data short of raw SQL.
- Models **tee-specific hole data** (yardage/par/stroke_index/tee location), which previously only existed as one course-wide value per hole regardless of tee:
  - New `hole_tees` table: sparse per-(hole, tee) override, falling back to the base `holes` row when absent (mirrors the existing `hole_scores.par` per-round override pattern).
  - `course_tees.is_primary` designates the tee whose data IS the base `holes` row, replacing an earlier unlabeled "Base" UI concept.
  - `resolveHole`/`resolveCourseTee` (`packages/core`) implement the precedence chain and are shared by both the editor and live/past round scoring on web and mobile.
  - Swapped three previously-duplicated tee-resolution fallbacks onto the shared helper.
- Course Editor UX: "✓ Saved" confirmations on course/hole/tee saves, tee/pin location dots on the holes table (filled = set, colors match the map markers), click-to-place map pins that auto-cycle tee → pin, and a smarter default placing-mode per hole.
- Other additions: `website`/`address`/`country` columns on courses and facilities, stroke-index duplicate/range validation in the holes table, and an extracted `@oga/supabase/osm-import` module shared by the CLI script and the editor's "re-fetch from OSM" button.
- Fixes a React Query cache-key collision where the editor's course fetch could be served a narrower shape (missing website/address/country) if another part of the app had cached it first.
- Carries forward `0047_postgrest_role_default_grants.sql` from `feature/round-import` — none of `main`'s existing migrations grant table-level access to `anon`/`authenticated`/`service_role`, so a from-scratch `supabase db reset` breaks this branch's dev backend too. Same filename/content on both branches, so it merges without conflict whichever lands first.

## Test plan
- [x] `turbo typecheck` clean across `@oga/web`, `@oga/core`, `@oga/supabase`
- [x] `apps/mobile` `tsc --noEmit` clean
- [x] `packages/core` test suite passes (617 tests, incl. new `hole-tees.test.ts`)
- [x] Verified in-browser: Course Editor CRUD (course/holes/tees/facility), tee-override save/clear, primary-tee designation, stroke-index validation blocking bad saves, tee/pin location dots, click-to-place auto-cycle, default placing-mode logic
- [x] Verified in-browser: live round scorecard/map picks up tee-specific overrides; unaffected holes still show base data
- [ ] Manual mobile device/simulator pass (not run — no Expo runtime in this environment)


cner-smith:
## Review — feature(course-editor): dev-only Course Editor + tee-specific hole data

**Blockers**
- Base branch is `main`, not `dev`. CONTRIBUTING.md: "Open a PR against `dev` (not `main`)... Maintainers periodically merge `dev → main`... contributors don't open PRs against `main`." Please retarget.
- This PR bundles at least four distinct concerns: the `hole_tees` schema + `resolveHole`/`resolveCourseTee` in `packages/core/src/hole-tees.ts` + their wiring into production round-viewing code on both platforms; the dev-only Course Editor UI and `apps/web/vite-plugins/dev-course-api.ts`; the unrelated `website`/`address`/`country` column additions (`supabase/migrations/0048_course_website_address.sql`, `0049_course_country.sql`); and a migration carried forward from another branch (`supabase/migrations/0047_postgrest_role_default_grants.sql`, per the PR's own description not authored for this PR). It also folds in a genuine bug fix (`apps/web/src/hooks/useCourseEditor.ts:19-27`, the React Query cache-key collision) inside the editor concern. CONTRIBUTING.md: "One concern per PR... PRs that add scope beyond their stated concern will be asked to split." Given the size (+3805/-557, 46 files), a split into schema+resolver+wiring / editor UI+API / address columns / 0047-standalone seems warranted.

**Concerns**
- `apps/web/vite-plugins/dev-course-api.ts:56` — every `/api/dev/*` route runs against the full service-role client (RLS bypass) with no authentication check while `vite dev` is running. I traced the dev-only gating end to end (`apply: 'serve'` in `vite.config.ts:14,35`, `import.meta.env.DEV` gates in `router.tsx:132-135` and `Sidebar.tsx:29-31`, non-`VITE_`-prefixed service-role key in `vite.config.ts:25-32`) and it does check out — this can't ship to production or the client bundle. Still, it's worth at least a comment acknowledging that anything reachable on the dev server gets unauthenticated write access to the DB via the service-role client.
- `apps/web/vite-plugins/dev-course-api.ts:224` — `readJson` types the request body `any` and passes it straight into Supabase inserts/updates with no runtime shape validation. CONTRIBUTING.md: "Don't add `any` to silence an error — fix the type." A malformed payload here only surfaces as an opaque Postgres/PostgREST error.
- `apps/web/src/pages/dev-editor/CourseEditorIndexPage.tsx:53` (`createNew`) and `:45` (`pickFacility`) — mutation/query errors are silently dropped (no try/catch around `createCourse.mutateAsync`; `getFacilityUnits`'s `error` is destructured away and discarded). On failure these look identical to the legitimate empty/idle state. The rest of the new editor code (`HolesEditor.tsx`, `TeesEditor.tsx`, `FacilityPicker.tsx`, `CourseFieldsForm.tsx`) does render `.error` states, so this looks like an oversight rather than intentional.
- `packages/supabase/src/osm-import.ts` and `packages/supabase/src/queries/courses.ts` add 14+ new exported functions with non-trivial logic (`matchHoles`'s nearest-neighbor dedup, `setPrimaryIfNone`'s race handling) and no tests. This isn't a literal violation — CONTRIBUTING.md's test requirement is scoped to `@oga/core` — but this was previously isolated CLI-script logic and is now shared production logic driving a UI button, so the bar for coverage arguably went up.

**Nits**
- `apps/web/src/components/dev-editor/CourseMapPicker.tsx:117` — `// eslint-disable-next-line react-hooks/exhaustive-deps` with no explanation, unlike the well-commented disable a few lines up in the same file (`:79-80`). CONTRIBUTING.md requires "No `// eslint-disable-next-line` without a comment explaining why."
- `apps/web/src/components/dev-editor/AddressWebsiteHelper.tsx` — extracted into its own file/component with a single caller (`CourseFieldsForm.tsx:144`) and no obvious complexity that would make it unreadable inline. Reads like a 3-caller-rule candidate for inlining.

- `FacilityPicker.tsx`'s `FacilityFields`/`NewFacilityForm` and `HolesEditor.tsx`'s `LocationDots` are also under the 3-caller threshold (1 and 2 callers respectively). More plausible readability justification than `AddressWebsiteHelper` given form complexity, but worth a second look if this file grows further.
- `HoleRoundOverride.tee_lat`/`tee_lng` in `packages/core/src/hole-tees.ts` are declared but no call site in this diff ever populates them (every caller only passes `{ par }`). Not a bug, just unused surface — fine to leave if something upcoming needs it, otherwise consider trimming.

---
Looked good: dev-only gating verified end-to-end (service-role key never reaches the client bundle, `/dev` routes and UI both tree-shake out of prod builds); RLS present and reasoned on the new `hole_tees` table; `packages/core/src/hole-tees.test.ts` thoroughly covers the override precedence chain (round > tee > base for par, live > round > tee > base for tee location); migrations 0047–0051 are sequentially numbered with clear rationale comments; `resolveCourseTee` genuinely deduplicated four previously-duplicated inline tee-resolution snippets across mobile and web.

---

Spedge:
Split into 4 focused PRs per the review above (blockers: retarget to `dev`, one concern per PR). Also excludes the `feature/round-import` merge that landed on this branch after the review — that stays in its own PR (#836), untouched.

1. #842 — grants migration (0047)
2. #843 — website/address/country columns (0048/0049)
3. #844 — hole_tees schema + resolver + production wiring (0050)
4. #845 — dev-only Course Editor UI + backend (0051), plus fixes for all the concerns/nits raised above

Stacked in that order — each depends on the ones before it (noted in each PR body). Closing this one in favor of those.

PR #842, opened 2026-08-16
fix(db): grant table access to anon/authenticated/service_role

Spedge:
Split out of #841 per review feedback — this is the standalone grants
migration, carried over from `feature/round-import` (identical filename/
content, same rationale) since `main`/`dev` are both missing it. No app
code changes.

None of the existing migrations grant table-level SELECT/INSERT/UPDATE/
DELETE to anon/authenticated/service_role on public-schema tables —
Supabase's local bootstrap only auto-grants for tables owned by
supabase_admin, not the postgres role our migrations run as. A
from-scratch `supabase db reset` hits "permission denied" on the first
table PostgREST touches.

First of 4 PRs replacing #841 (see #841 for the full split rationale):
1. **this PR** — grants migration
2. course-address-fields — website/address/country columns
3. hole-tees-resolver — hole_tees schema + resolver + production wiring
4. course-editor-ui — dev-only Course Editor UI + backend

### DIFF
--- a/supabase/migrations/0047_postgrest_role_default_grants.sql
+++ b/supabase/migrations/0047_postgrest_role_default_grants.sql
@@ -1,0 +1,29 @@
+-- `supabase db reset` applies these migrations as the `postgres` role, so
+-- every table here is owned by `postgres`. The Supabase local image's own
+-- bootstrap sets default privileges so future tables owned by
+-- `supabase_admin` auto-grant full CRUD to anon/authenticated/service_role,
+-- but nothing in this repo's migration history ever did the same for the
+-- `postgres` role — so on a from-scratch reset, PostgREST (which runs
+-- queries as anon/authenticated/service_role) gets "permission denied" on
+-- the first table it touches. RLS policies are the real access boundary
+-- throughout this schema; these grants are just the base plumbing PostgREST
+-- needs to attempt a query at all.
+--
+-- Two parts: backfill grants on tables this migration history already
+-- created, and set default privileges so tables created by later
+-- migrations get them automatically.
+--
+-- `anon` only gets `select` (defense-in-depth beyond RLS: a table that ships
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(db): grant table access to anon/authenticated/service_role

**Concerns**
- `supabase/migrations/0047_postgrest_role_default_grants.sql:1` — this is numbered `0047`, but the highest migration on `dev` is `0045` (`supabase/migrations/0045_lock_facility_columns.sql`), so the next free number on this base is `0046`. CONTRIBUTING.md's migration rule is explicit: "The next migration goes in the next free number ... No date prefixes, no skipped numbers." The commit message notes an identical migration exists on `feature/round-import` (#836) — but that PR targets `main`, not `dev`, and isn't merged, so nothing currently claims `0046` on this base. If this lands on `dev` first it leaves a permanent `0046` gap. I'd renumber to `0046`, or confirm the coordinating PR lands `0046` ahead of it.
- `supabase/migrations/0047_postgrest_role_default_grants.sql:16` — this grants full `insert, update, delete` to `anon` on every current and future public table. That is safe only while every table has RLS enabled; the header comment's claim that "RLS ... is the real access boundary" holds only where RLS is actually on. Any table that ships with RLS disabled — or is created before its own `enable row level security` runs — becomes anon-*writable*, not just anon-readable. Granting `anon` `select` only, and reserving DML for `authenticated`/`service_role`, still lets PostgREST attempt queries while shrinking the blast radius. This may be intentional to mirror Supabase's own `supabase_admin` default grants; flagging it as defense-in-depth, not a blocker.

---
Looked good: the header comment accurately explains the GRANT-vs-RLS distinction, and the two-part structure — `grant ... on all tables` to backfill existing tables plus `alter default privileges` for ones created by later migrations in the same reset — correctly covers both cases.

---

PR #843, opened 2026-08-16
feat(db): add website/address/country columns to courses and facilities

Spedge:
Split out of #841 per review feedback — schema-only, no app code
references these columns yet (that comes in course-editor-ui, PR 4/4).

Dev-entered metadata for the upcoming Course Editor. Nullable, no format
validation.

**Depends on #842** (grants migration) — no push access to upstream to
stack branch bases across forks, so this PR's diff currently also
includes #842's commit; it'll shrink to just this PR's change once #842
merges. Please merge #842 first.

2nd of 4 PRs replacing #841:
1. #842 — grants migration
2. **this PR** — website/address/country columns
3. hole-tees-resolver — hole_tees schema + resolver + production wiring
4. course-editor-ui — dev-only Course Editor UI + backend

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #844, opened 2026-08-16
feat(rounds): per-tee hole overrides (hole_tees table + resolveHole/resolveCourseTee)

Spedge:
Split out of #841 per review feedback — the schema+resolver+production-
wiring concern, independent of the dev-only Course Editor that will
write hole_tees rows (course-editor-ui, PR 4/4).

Adds `hole_tees` (sparse per-tee override for yards/par/stroke_index/tee
location) plus `resolveHole`/`resolveCourseTee` in `@oga/core`, and wires
production round-viewing (web + mobile) through the resolver.
`resolveCourseTee` is lifted out of three previously-duplicated inline
copies of the id/tee_color fallback (`useCompleteRound.ts`,
`useRoundActions.ts`, mobile `completeRound.ts`).

**Depends on #842 and #843** — no push access to upstream to stack branch
bases across forks, so this PR's diff currently also includes their
commits; it'll shrink to just this PR's change once they merge. Please
merge #842 and #843 first.

3rd of 4 PRs replacing #841:
1. #842 — grants migration
2. #843 — website/address/country columns
3. **this PR** — hole_tees schema + resolver + production wiring
4. course-editor-ui — dev-only Course Editor UI + backend


cner-smith:
## Review — feat(rounds): per-tee hole overrides (hole_tees table + resolveHole/resolveCourseTee)

**Concerns**
- `apps/web/src/hooks/useCourses.ts:588`, `apps/mobile/components/round/hole/useHoleData.ts:428`, `apps/mobile/app/(app)/round/[id]/index.tsx:59` — the same join-shape strip, `((data ?? []) as Array<HoleTeeRow & { holes?: unknown }>).map(({ holes: _h, ...row }) => row as HoleTeeRow)`, is copy-pasted verbatim (with the same explanatory comment) across three call sites. That's exactly the threshold CONTRIBUTING.md's reuse rule targets ("three or more callers"). Stripping the `holes!inner(course_id)` join shape once — e.g. having `getHoleTeesForCourse` (`packages/supabase/src/queries/courses.ts:153`) return clean `HoleTeeRow[]`, or a small shared `stripHoleJoin` helper — keeps the join shape from leaking to three consumers that each have to know to remove it.

**Nits**
- `packages/core/src/hole-tees.ts:23-24` — `HoleRoundOverride.tee_lat`/`tee_lng` have no production writer. Every call site (`useRoundData.ts:852,861`, `useHoleData.ts:363,376`, `PastRoundMap.tsx:191`) passes only `{ par }`; live tee-drag goes through the separate `liveOverride` slot, and there's no per-round persisted tee column in `hole_scores` for these to read from. So the round-tee-location precedence tier is exercised only by the unit test, not by any caller — the "no params for hypothetical future use" case CONTRIBUTING.md calls out. Worth dropping the two fields until the follow-up Course Editor PR adds a writer.
- `apps/mobile/components/round/hole/useHoleData.ts:361` — `resolvedHole` is recomputed inline (fresh object each render) while `resolvedHoleByNumber` right below it is `useMemo`'d. Harmless in practice (consumers only read primitive `.par`/`.yards`), but the inconsistency stands out next to its memoized sibling.
- Commit message — it lists `useRoundActions.ts` among the three inline `resolveCourseTee` copies being lifted, but that file had no inline copy (it only gained `resolvedPar`). The actual inline copies removed are in `useCompleteRound.ts`, mobile `completeRound.ts`, `RoundDetailPage.tsx`, `RoundHeader.tsx`, and mobile `index.tsx` — five real callers, so the 3-caller rule is comfortably met; only the message's file list is off.

- `apps/web/src/hooks/useRoundActions.ts:611` now feeds `resolvedPar ?? activeHole.par` into `inferHoleStats`, where it previously passed raw `activeHole.par`. Because `resolvedPar` folds in the per-round `hole_scores.par` override, this silently fixes a latent bug (shot inference ignoring a corrected par). Worth calling out in the PR description as an intended behavior change, not just tee wiring.

---
Looked good: `hole_tees` has RLS enabled with public-read only (course reference data, write reserved to service-role) plus `par`/`stroke_index` CHECK constraints, FK cascades, `unique(hole_id, course_tee_id)`, and indexes on both FKs (`supabase/migrations/0050_hole_tees.sql`); `resolveHole`/`resolveCourseTee` live in `@oga/core` with real tests covering sparse partial merge, par precedence, live>round>tee>base tee-location, case-insensitive `tee_color`, and null returns; the resolver uses `??` (not `||`) so a legitimate `0` coordinate survives; no `any`, no secrets, no nativewind/mobile-dep changes.

---

PR #845, opened 2026-08-16
feature(course-editor): dev-only Course Editor UI + backend

Spedge:
Split out of #841 per review feedback — the dev-only editor concern.

Dev-only editor (`apps/web/src/pages/dev-editor`) for courses,
facilities, holes, tees, and per-tee hole overrides, backed by a
service-role Vite middleware (`apps/web/vite-plugins/dev-course-api.ts`)
since courses/holes have no UPDATE/DELETE RLS policy and facilities
require service_role for every write. Includes an OSM re-fetch button
sharing `packages/supabase/src/osm-import.ts` with
`scripts/import-osm-course.ts` (extracted from the CLI script, which now
just calls the shared module).

Also addresses review concerns/nits on this code:
- `dev-course-api.ts`: documented the no-auth/service-role tradeoff, and
  replaced `readJson`'s `any` return with `unknown` + a cast to the
  specific Insert/Update type at each call site.
- `CourseEditorIndexPage.tsx`: surfaced the two previously-swallowed
  errors (`pickFacility`, `createNew`) instead of failing silently.
- Added `packages/supabase` vitest coverage for `matchHoles`
  (osm-import's nearest-neighbor dedup) and `setPrimaryIfNone`
  (course_tees race-safety re-check) — the two pieces of non-trivial
  untested logic called out in review.
- `CourseMapPicker.tsx`: explained the remaining eslint-disable, matching
  the documented one earlier in the same file.
- Inlined `AddressWebsiteHelper.tsx` into its sole caller
  (`CourseFieldsForm.tsx`) per the 3-caller rule.

**Depends on #842, #843, and #844** — no push access to upstream to
stack branch bases across forks, so this PR's diff currently also
includes their commits; it'll shrink to just this PR's change once they
merge. Please merge those first.

Last of 4 PRs replacing #841:
1. #842 — grants migration
2. #843 — website/address/country columns
3. #844 — hole_tees schema + resolver + production wiring
4. **this PR** — dev-only Course Editor UI + backend


cner-smith:
## Review — feature(course-editor): dev-only Course Editor UI + backend

**Concerns**
- `apps/web/src/components/dev-editor/HolesEditor.tsx:138-141` — the base `rows` state is seeded from the `holes` prop on mount and re-seeded only on `[courseId]` change. But the `OsmRefetchButton` on the same page invalidates `['holes', courseId]` on success (`useCourseEditor.ts` `useRefetchOsm`), which wipes+replaces the holes and updates the `holes` prop — yet the grid keeps showing pre-refetch rows because `courseId` didn't change. So "Re-fetch from OSM" updates the summary but the holes table appears frozen until the user navigates away and back. Re-seed when the `holes` prop identity changes and there are no unsaved base edits (or key the editor on a holes version), rather than on `[courseId]` alone.
- `apps/web/vite-plugins/dev-course-api.ts:89-236` — every `/api/dev/*` write casts the raw JSON body straight to a typed `Insert`/`Update` (`(await readJson(req)) as CourseInsert`, etc.) and hands it to a service-role write with no runtime validation and no auth (documented at lines 67-72); Postgres column constraints are the only backstop. This is an acceptable, explicitly-documented tradeoff *because* the plugin is dev-server-only (`apply: 'serve'`) and never ships. Flagging it only to mark the boundary: if these endpoints ever move behind a real server, validation + auth must be added before that move, since service-role bypasses RLS.

**Nits**
- `apps/web/src/components/dev-editor/TeesEditor.tsx:76-88`, `apps/web/src/components/dev-editor/FacilityPicker.tsx:108-113` — these `onSave`/`onSubmit` handlers call `mutateAsync(...)` inside an async callback whose promise floats (invoked from `onClick`, no await/catch). The failure is still surfaced via `upsert.error`/`createFacility.error`, but a rejected `mutateAsync` also emits an unhandled-rejection warning. `createNew` in `CourseEditorIndexPage.tsx:64-72` already shows the try/catch pattern that avoids it.
- `apps/web/src/components/dev-editor/CourseMapPicker.tsx:19-26` — the `clickToPlaceId` JSDoc says "the two-marker hole picker leaves this unset (a stray click shouldn't silently relocate tee or pin…)", but `HolesEditor.tsx` now passes `clickToPlaceId={placing}` and deliberately relies on click-to-place. The doc now contradicts its caller.

- `packages/supabase/src/queries/courses.ts:242-286` — `setPrimaryIfNone`'s comment describes the SELECT re-check as making it "race-safe"; the real guarantee is the `course_tees_one_primary_per_course` partial unique index (a concurrent double-promote fails the second write rather than corrupting state). Fine for single-user dev tooling — just noting the comment credits the wrong mechanism.
- `packages/supabase/src/osm-import.ts` — `haversineMeters`, `centroid`, `metersToYards` are generic geo math with no Supabase dependency and would fit `@oga/core`; reasonable to leave co-located for now, but a candidate to lift if any other surface needs distance math.

---
Looked good: the service-role key is fully contained server-side — read only via `process.env.SUPABASE_SERVICE_ROLE_KEY` in `dev-course-api.ts`, admin client built through `ssrLoadModule('@oga/supabase')` (never a static/browser import), plugin is `apply: 'serve'` so `vite build` drops it, the env var is deliberately not `VITE_`-prefixed so it's never inlined, and both the routes (`router.tsx:60-64`) and nav (`Sidebar.tsx:29-31`) are `import.meta.env.DEV`-gated and tree-shaken from prod. `readJson` returns `unknown` with per-call-site casts (no `any` silencing), every `eslint-disable-next-line` carries an explanation, the `osm-import.ts` extraction is genuine reuse across the CLI script and the dev plugin, `matchHoles`/`setPrimaryIfNone` have real vitest coverage, and writes are batched (no N+1).

---

PR #848, opened 2026-08-20
fix(mobile): live-round on-course fixes — green-slope button, first-hole framing, GPS recenter

cner-smith:
On-course testing fixes (all mobile JS → OTA-eligible):

- **Remove the 'green slope heatmap' toolbar stub** — it was a disabled 'coming soon' button.
- **First hole opens pin-at-top** — the late-pin re-frame no longer bails when the pin resolves before the camera inits, which left the map stuck north-up (common on the first/synthetic hole).
- **GPS recenter targets the freshest fix** — the crosshair button centered on the 5m-displacement-throttled gpsPosition, landing up to ~5m off the live LocationPuck; now reads getLastKnownLocation so camera + ball land on the puck.

Mobile typecheck green. Device-QA notes: green flow untouched; verify first-hole framing on a course with a real pin, and the recenter button on-course.

### DIFF
--- a/apps/mobile/components/round/HoleMap.tsx
+++ b/apps/mobile/components/round/HoleMap.tsx
@@ -244,17 +244,39 @@ export function HoleMap({
     courseCenter,
   })
 
-  const recenterOnGps = useCallback(() => {
-    if (!gpsPosition || !cameraRef.current) return
+  const recenterOnGps = useCallback(async () => {
+    if (!cameraRef.current) return
+    // Recenter on the FRESHEST fix, not the state gpsPosition. That value is
+    // throttled by the listener's 5 m displacement filter (useHoleState), so
+    // "center on me" could land the camera up to ~5 m off the live native
+    // LocationPuck — the puck ended up off-centre and it read as "the camera
+    // moved but the dot didn't." getLastKnownLocation reads the same native
+    // engine the puck renders from. Falls back to the state value if it's
+    // null / throws.
+    let target = gpsPosition
+    try {
+      const last = await Mapbox.locationManager.getLastKnownLocation()
+      if (
+        last &&
+        Number.isFinite(last.coords.latitude) &&
+        Number.isFinite(last.coords.longitude)
+      ) {
+        target = { lat: last.coords.latitude, lng: last.coords.longitude }
+      }
+    } catch {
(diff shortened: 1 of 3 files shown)


cner-smith (commented):
## Review — fix(mobile): live-round on-course fixes — green-slope button, first-hole framing, GPS recenter

**Concerns**
- Scope / one-concern-per-PR — this bundles three independent fixes with distinct root causes: the green-slope toolbar stub removal (`apps/mobile/components/round/HoleMapOverlays.tsx:189`), the first-hole framing race (`apps/mobile/components/round/hooks/useHoleCamera.ts:158`), and the GPS-recenter freshness change (`apps/mobile/components/round/HoleMap.tsx:247`). CONTRIBUTING.md's Opening-a-PR rule is "One concern per PR. UI restyle, bug fix, and new feature do not belong in the same PR," reinforced by "default to no, ship the minimum … PRs that add scope beyond their stated concern will be asked to split." The two camera/GPS fixes at least share the live-round camera subsystem; the stub removal is unrelated and would stand alone as a trivial `chore/`. This may be an intentional "on-course testing" batch, but flagging per the rule — worth confirming the maintainer is fine grouping these rather than splitting the stub removal out.

**Nits**
- `apps/mobile/components/round/HoleMap.tsx:247` — `recenterOnGps` is now `async` and used as a tap handler (fire-and-forget). The single `await` is wrapped in try/catch and the trailing `setCamera` is sync, so there's no unhandled rejection here. Rapid double-taps could fire overlapping invocations that both call `setCamera` at the end, but with near-identical targets that's a visual no-op — not worth guarding unless it shows up in QA. Noting only because the async-onPress pattern is easy to misread later.

---
Looked good: no secrets/migrations/service-role exposure; all changes are UI/camera wiring in `apps/mobile` (no pure logic misplaced from `@oga/core`, no new core exports needing tests); dropping the `!cameraInitialized` gate at `useHoleCamera.ts:158` is correctly latched by `headingAppliedPinRef` and keyed off pin coords, so it won't re-fire on GPS ticks; the recenter fallback chain handles a null/throwing `getLastKnownLocation` and re-checks the camera ref after the await; targets `dev`, no dep or `nativewind` changes.

---

PR #849, opened 2026-08-20
release: live-round on-course fixes (green-slope button, first-hole framing, GPS recenter)

cner-smith:
Mobile on-course fixes from device testing. All mobile JS — ships to existing installs via OTA after this merge.

- Remove green-slope toolbar stub.
- First hole opens pin-at-top (framing race fix).
- GPS recenter targets the freshest fix (camera + ball land on the live puck).

### DIFF
--- a/apps/mobile/components/round/HoleMap.tsx
+++ b/apps/mobile/components/round/HoleMap.tsx
@@ -244,17 +244,39 @@ export function HoleMap({
     courseCenter,
   })
 
-  const recenterOnGps = useCallback(() => {
-    if (!gpsPosition || !cameraRef.current) return
+  const recenterOnGps = useCallback(async () => {
+    if (!cameraRef.current) return
+    // Recenter on the FRESHEST fix, not the state gpsPosition. That value is
+    // throttled by the listener's 5 m displacement filter (useHoleState), so
+    // "center on me" could land the camera up to ~5 m off the live native
+    // LocationPuck — the puck ended up off-centre and it read as "the camera
+    // moved but the dot didn't." getLastKnownLocation reads the same native
+    // engine the puck renders from. Falls back to the state value if it's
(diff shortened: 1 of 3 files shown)


cner-smith:
## Review — release: live-round on-course fixes (green-slope button, first-hole framing, GPS recenter)

No blockers, concerns, or nits.

A few things I checked that could look like flags but aren't:

- **`dev → main` target** is the sanctioned path here, not a mis-targeted PR. `CONTRIBUTING.md`: "Maintainers periodically merge `dev → main` via PR to cut a release." This is that release rollup by the maintainer, so the "branch from `dev`, PR against `dev`" rule doesn't apply.
- **Green-slope removal** (`HoleMapOverlays.tsx:227`) drops a `disabled` "coming soon" toolbar stub for a never-shipped v1.1 heatmap — it doesn't touch lie/slope shot tracking, so the "lie-slope-aware" pillar is intact.
- **`recenterOnGps` going async** (`HoleMap.tsx:247`) re-checks `cameraRef.current` after the `await` for the unmount window, and the only `await` is wrapped in try/catch with a `gpsPosition` fallback — no unhandled rejection, and `setCamera` stays synchronous.
- **Dropping the `cameraInitialized` gate** (`useHoleCamera.ts:160`) is guarded against re-firing by the `headingAppliedPinRef` key latch (`:174`), which resets per hole (`:147`), so there's no camera-fight or repeated re-frame. The load-order reasoning in the comment matches the two prior effects' latches.

All three files are mobile UI/native-camera wiring — no `@oga/core` logic, no migrations, no dep changes, so package-boundary and test-coverage rules don't trigger.

---
Looked good: async recenter handles the unmount window and falls back cleanly; the first-hole re-frame is latched so it can't loop; scope held to on-course fixes.

---

PR #852, opened 2026-08-25
fix(mobile): summary shot-delete disabled on first open (missing client id)

cner-smith:
## Bug (device-repro'd)

On the **first** open of the end-of-hole summary in a live round, every shot row's trash affordance is disabled — tapping does nothing. It only starts working after leaving and re-entering the summary (Edit on map → Finish hole again).

## Root cause

`persistShot` (useShotActions) mints the shot's client `id` **inside** `insertPendingShot` (which stamps it onto the SQLite row) but then optimistically appends the **pre-mint** payload to the in-memory `pendingForHole` list. So the in-memory shot has coords but no `id`.

`useHoleData.previousShotIds` skips id-less pending shots (its `&& p.id` guard) while `previousShots` keeps them → the two arrays diverge in length → the summary's row→shot-id zip (`shotIds[i]`) lands `undefined` on the tail rows → `deleteDisabled = !row._shotId` disables the trash. A refetch rebuilds `pendingForHole` from SQLite (which has the id), which is why the round-trip "fixes" it.

## Fix

Stamp the client id once, up front, and use that stamped payload for **both** the DB insert and the optimistic in-memory append — so they always carry the same id (mirrors the `id ? … : uuid.v4()` idiom already in `db.ts`).

JS-only → OTA-eligible. No migration, no API change.

### DIFF
--- a/apps/mobile/components/round/hole/useShotActions.ts
+++ b/apps/mobile/components/round/hole/useShotActions.ts
@@ -1,6 +1,7 @@
 import { useRef, useState, type Dispatch, type SetStateAction } from 'react'
 import { Alert } from 'react-native'
 import { useRouter } from 'expo-router'
+import { uuid } from 'expo-modules-core'
 import type { User } from '@supabase/supabase-js'
 import {
   combinedBreakDirection,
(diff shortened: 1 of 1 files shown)


cner-smith:

The fix stamps the client `id` once up front and reuses that stamped payload for both the SQLite insert and the optimistic `pendingForHole` append, so `previousShots` and `previousShotIds` (`useHoleData.ts:299-330`) stay length-aligned on first open. It uses the same `uuid` source and `id ? … : uuid.v4()` idiom already in `db.ts` (`db.ts:2,61`), the `insertPendingShot` guard no-ops its own mint when the id is present, and all optimistic paths (`persistPutt`, `markBallHere`) funnel through `persistShot`, so the single-file change covers them. Scope held to one concern; correct package (mobile UI wiring, not `@oga/core` logic); no new deps, migration, or `any`; targets `dev`.

---

PR #853, opened 2026-08-25
release: shot-delete first-open fix (dev→main)

cner-smith:
Promotes #852 (summary shot-delete disabled on first open — optimistic pending shot missing its client id) to main for a production OTA. JS-only, no migration.

### DIFF
--- a/apps/mobile/components/round/hole/useShotActions.ts
+++ b/apps/mobile/components/round/hole/useShotActions.ts
@@ -1,6 +1,7 @@
 import { useRef, useState, type Dispatch, type SetStateAction } from 'react'
 import { Alert } from 'react-native'
 import { useRouter } from 'expo-router'
+import { uuid } from 'expo-modules-core'
 import type { User } from '@supabase/supabase-js'
 import {
   combinedBreakDirection,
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — release: shot-delete first-open fix (dev→main)

Walked the diff (single file, `apps/mobile/components/round/hole/useShotActions.ts`) against the must-check list and gave it a senior pass. No blockers, concerns, or nits.

The fix is sound: `persistShot` now stamps `payload.id` (`useShotActions.ts:247`) before both `insertPendingShot(payload)` and the `setPendingForHole` push, so the in-memory optimistic entry carries the same client id that `insertPendingShot` already writes to SQLite (`lib/db.ts:61`). That closes the mismatch where `previousShots` kept the shot but `previousShotIds` dropped it on its `&& p.id` guard (`useHoleData.ts:322`), which is what desynced the summary's row→shot map. `uuid` from `expo-modules-core` mirrors the existing `db.ts` import, and since `buildPayload` never sets `id`, the `base.id ? base : …` ternary is a harmless defensive mirror of `insertPendingShot`'s own guard.

Notes on the checks that could have flagged:
- Base is `main`, not `dev` — not a violation here. CONTRIBUTING.md: "Maintainers periodically merge `dev → main` via PR to cut a release; contributors don't open PRs against `main`." This is that sanctioned maintainer release PR.
- No new dependency added — `expo-modules-core` is already a dependency (imported in `lib/db.ts`), so the mobile exact-pin rule doesn't apply.
- JS-only, no migration, so no RLS/security surface; scope held to one concern; no `@oga/core` exports, so no test requirement.

---
Looked good: id stamped before both persistence paths so `previousShots`/`previousShotIds` stay 1:1; sanctioned dev→main release PR; single-concern, JS-only diff.

---

PR #855, opened 2026-08-25
feat(mobile): on-map shot editing for played holes (Phase 2A)

cner-smith:
## What

On-map editing of an already-played hole during a live round: **step** through shots (‹ N of M ›), **drag to move** a shot (recomputing its strokes-gained distance), and **delete** a shot. A played hole now opens in this edit mode; the active/live hole stays in capture mode — which also fixes the bug where returning to a played hole showed the live aim-line chrome. Also **unifies `PastRoundMap`** onto the same move-projection helper and the `delete_shot` RPC.

Built subagent-driven from the reviewed spec + plan (`docs/superpowers/{specs,plans}/2026-08-25-live-shot-editing-phase2a*`). Online-only; **no migration**.

## Commits (7)

- **`@oga/core projectShotMove`** — pure, unit-tested move projection: `distance_to_target = round(haversine(start,pin))`; putt → null; no pin → leave unchanged (#662 guard); coord-0 safe.
- **`moveShot`** (useShotActions) — guarded client `shots.update` via `projectShotMove`; flushes pending first + fails (no silent loss) on a 0-row update; online-first.
- **`PastRoundMap` unify** — move via `projectShotMove` (+ the #662 guard it lacked; also fixes dragging a putt writing bogus yardage); **delete via `delete_shot` RPC** — fixing a real bug where its delete never renumbered `shots.shot_number` in the DB or re-tallied putts/penalties.
- **Edit mode UI** — new `ShotStepper`; played hole opens with stepper + drag-move + delete over `HoleMap`, capture chrome suppressed; reached via the hole switcher and "Edit on map".
- **2 fix rounds** on the edit-vs-capture predicate (below) + a final-review fix.

## Review

Each task passed spec + quality review. The whole-branch opus review (SHIP-WITH-NITS) drove:
- **Predicate hardened twice** — the "played vs active hole" gate now ratchets a **finish-frontier** (`furthestHoleReached`, bumped only on a genuine finish-advance), so the actively-logged hole can never flip to edit mode (two earlier variants — a backward peek and a forward peek/scorecard-jump — are both structurally closed).
- **`moveShot` silent-loss** fixed (flush + 0-row guard).
Verified regression-free: the shared `useHoleData` refetch-reset change keeps `previousShots`/`previousShotIds` index-aligned and doesn't regress the shipped Phase 1 summary-delete; new props on `HoleMap`/`MapBottomChrome` default to no-op for `PastRoundMap`.

## Before merge

**Device QA (per the plan):** back to a played hole via the switcher → edit mode (no aim line); step/drag/delete; drag persists + SG reflects it; delete renumbers + re-tallies; airplane-mode guard; the **live hole stays in capture mode**; "Edit on map" opens edit mode; and a **past-round regression pass** (PastRoundMap delete now renumbers+re-tallies, move keeps distance on a pin-less hole).

JS-only → OTA-eligible after QA.


cner-smith:
## Review — feat(mobile): on-map shot editing for played holes (Phase 2A)

**Concerns**
- `apps/mobile/components/round/PastRoundMap.tsx:582` — switching the past-round delete to the `delete_shot` RPC and dropping the trailing `syncScore(...)` means the client's `holeScores` state is now never updated after a delete. `delete_shot` re-tallies `score/putts/penalties/fairway_hit/gir` server-side, but on the client `handleRemoveShot` only calls `onShotRemoved` (which just prunes the `shots` array) — nothing calls `onHoleScoreChanged`. The scorecard/running-score display reads `hs.score` off `holeScores` (`app/(app)/round/[id]/index.tsx:616`, also `:585`/`:869`), so after deleting a shot on the past-round screen the shown hole score and round total stay one (or more) strokes stale until a full refetch/remount. The old `syncScore` path at least pushed the new `score` back via `onHoleScoreChanged`; this path pushes nothing. I think you want `delete_shot` to return the updated `hole_scores` row (or refetch it) and propagate it through `onHoleScoreChanged`. The live-round `deleteShot` looks unaffected since it goes through `data.refreshShots()`.
- Scope / `CONTRIBUTING.md` "One concern per PR" — the PR bundles the new Phase 2A edit feature with a genuine pre-existing bug fix on a different surface: the `PastRoundMap` delete that "never renumbered `shots.shot_number` in the DB or re-tallied putts/penalties" (`PastRoundMap.tsx:567`). CONTRIBUTING says "UI restyle, bug fix, and new feature do not belong in the same PR." The shared `projectShotMove` / `delete_shot` unification is a defensible reason to touch `PastRoundMap` here, so this may be intentional — but the past-round delete-renumber fix reads like a standalone `fix/` that could ship (and be reverted) independently of the feature.

**Nits**
- `apps/mobile/components/round/ShotStepper.tsx:1` — extracted to its own 133-line file with a single caller (`LiveRoundSession.tsx:859`). The file header describes it as "live round + past round," but `PastRoundMap` doesn't consume it in this diff, so it's currently a single-caller component. CONTRIBUTING's 3-caller rule says "Single-caller 'components' stay co-located in the parent file" unless the inline version is "genuinely unreadable" — a 133-line presentational block arguably clears that bar given how large `LiveRoundSession` already is, so this may be fine; flagging only because the second (past-round) caller it's named for isn't wired up yet ("add them when there's a real caller").

---
Looked good: `projectShotMove` correctly placed in `@oga/core` with tests (incl. the #662 no-pin and coord-0 cases); `delete_shot` RPC has `auth.uid()` + owner check and `security definer`/fixed `search_path`; `moveShot` flushes pending and fails loudly on a 0-row update rather than silently losing the move; the `useHoleData` reset now scoped to a real hole_score change so an in-place move/delete doesn't flash edit mode off; no migration, no dep changes, targets `dev`.

---

PR #864, opened 2026-08-26
fix(mobile): numeric break input on putt aimer (#861)

### LINKED ISSUE #861: Putt 'Read' aimer: 5-inch break floor + numeric input
From on-course feedback: the putt Read aimer's break slider floors at ~5 inches (won't go to 'straight'/lower). Add a numeric field beside the slider, editable and two-way-synced with the handle so you can type an exact break amount. First verify whether the 5-inch floor is intentional before removing it.

cner-smith:
## What

Adds a numeric break input + Left / Straight / Right toggle to the putt "Read" aimer (`GreenDiagram`), two-way synced with the drag handle. Closes #861.

## The "5-inch floor" — verified, not a code bug

Investigated first, per the issue's ask. There is **no 5-inch floor anywhere in code**:

- The handle range is ±33 in, rounds to 1-in steps, and **centers to 0 = "Straight"**; the only deadband is ±2 in (`formatAim` + core `horizontalBreakFromAim`).
- The "~5-inch floor, won't reach straight" is a **drag-precision artifact** — a fat handle on a phone can't reliably land on exact center or a small exact value.

So nothing was "removed." The fix is the numeric field the issue asks for, which makes exact/straight reads reachable by typing.

## How it works

- Below the diagram: a magnitude field (inches) + a 3-way **Left / Straight / Right** segmented toggle — matches how a golfer reads a break, mirrors the existing "N in left/right" label.
- **Two-way synced:** dragging the handle updates the number + toggle; typing / tapping moves the handle. Left = negative offset, right = positive, straight = 0 (same sign convention the handle and core already use).
- Typed magnitude clamps to the same ±33 in the handle does, so the two controls can never disagree.
- Lives inside `GreenDiagram`, so **both** aimer surfaces (end-of-hole summary's AimerOverlay + PuttingSheet) get it — no duplication.

## Scope / risk

Mobile, JS-only, one file (+144). No migration, no core/API change, props unchanged. OTA-eligible once QA'd + released. Mobile typecheck green.

## Needs on-device QA (no RN render test path)

- Type a value → handle moves; drag handle → number/toggle update.
- Tap Straight → 0 / handle centers; type 0 or clear → straight.
- Type 40 → clamps to 33 (handle at edge).
- Reachable in both the EOH summary "Read" and the live PuttingSheet.

### DIFF
--- a/apps/mobile/components/round/GreenDiagram.tsx
+++ b/apps/mobile/components/round/GreenDiagram.tsx
@@ -1,5 +1,5 @@
-import { useCallback, useEffect } from 'react'
-import { Text, View } from 'react-native'
+import { useCallback, useEffect, useRef, useState } from 'react'
+import { Text, TextInput, View } from 'react-native'
 import Svg, { Circle, Ellipse, Line, Path } from 'react-native-svg'
 import {
   Gesture,
@@ -13,6 +13,7 @@ import Animated, {
 import type { BreakDirection } from '@oga/core'
 import { useUnits } from '../../hooks/useUnits'
 import { TYPE } from '../../lib/typography'
+import { PressableTouch } from '../ui/PressableTouch'
 
 const AnimatedCircle = Animated.createAnimatedComponent(Circle)
 const AnimatedPath = Animated.createAnimatedComponent(Path)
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — fix(mobile): numeric break input on putt aimer (#861)

**Blockers**
- `apps/mobile/components/round/GreenDiagram.tsx:186` — `onDir` re-derives the magnitude straight from `magText` (`parseInt(magText || '0', 10)`) and passes it to `commitInputs` **without clamping** to `MAX_OFFSET_INCHES`. `onMagText:173` clamps the *committed* value but stores the *raw* digits in state (`setMagText(digits)` at :177), so `magText` can hold `"40"` after typing 40. Sequence: type `40` (field shows "40", commit clamps to 33) → tap **Left/Right** → `mag = parseInt("40") = 40` → `commitInputs` sends `-40`/`40` to `onAimChange`, i.e. an aim offset outside ±33. That value flows to the stored aim and `horizontalBreakFromAim`, and it contradicts the PR's stated invariant ("Typed magnitude clamps to the same ±33 … so the two controls can never disagree"). Cleanest fix is to clamp centrally in `commitInputs` (:161) — `Math.min(MAX_OFFSET_INCHES, Math.abs(mag))` before applying the sign — so every path (type, toggle, drag) shares one clamp.

**Concerns**
- `apps/mobile/components/round/GreenDiagram.tsx:177` — same root cause, display side: after typing `40` the field keeps reading `"40"` while the committed offset is 33. The resync `useEffect` (:154) can't correct it because `aimOffsetInches (33) === lastCommitted.current (33)` early-returns, so the number and the handle/toggle visibly disagree. Setting `setMagText(String(mag))` (the clamped magnitude) fixes this and also removes the stale-text source feeding the blocker above. This narrows the "typing `05` isn't reformatted" nicety only for over-max input, which seems acceptable.

**Nits**
- `apps/mobile/components/round/GreenDiagram.tsx:68,73` — the `±2` deadband in `dirFromOffset`/`magFromOffset` is a third copy of the same magic number already in local `formatAim` (:455) and core's `horizontalBreakFromAim` (`packages/core/src/types.ts:222`). If core's deadband ever changes, these drift silently. Deriving from a single shared constant would be safer, though I'd leave the helpers co-located here — `formatAim` already lives in this file, so this doesn't read as a `@oga/core` boundary violation for a single mobile caller.

---
Looked good: scope held to one concern (one mobile file, JS-only, no migration/core/API change), base correctly targets `dev`, no secrets/RLS/paywall/aim-input pillar issues, two-way sync via `lastCommitted` echo suppression is sound for the drag→input direction.

---

Issue #860, opened 2026-08-26
Admin/ops dashboard


cner-smith:
Internal admin/ops dashboard for app + website health, user metrics, and data-quality. Start with the 3 numbers that drive action, then expand: pending/self-identifying course imports, crawler sweep status, active users/rounds, error/health signals. Build AFTER the current release stack. Context: memory project_admin_dashboard.

PR #868, opened 2026-09-01
feat(db): gate user-submitted courses behind approval (0053)

cner-smith:
Gates user-submitted courses behind approval. Migration only — **no client code changes**, because RLS does the filtering.

## The problem

Verified on prod 2026-09-01, the SELECT policy on `courses` was `using (true)`:

| cmd | policy | predicate |
|---|---|---|
| SELECT | "Anyone can read courses" | `using (true)` |
| INSERT | "Authenticated users can add courses" | `check (auth.uid() is not null)` |
| UPDATE / DELETE | *(none)* | service-role only |

So **any authenticated user could insert a course and it was instantly visible to everyone**, with no review step. 24 such rows exist today (all `created_by` set, `external_id` null) from 20 submitters — 22 of them in the last 30 days, i.e. accelerating.

## What this does

- Adds `courses.approved_at timestamptz` (nullable, no default → no table rewrite on ~20.7k rows).
- Backfills crawler courses (`created_by is null`, 20,733) **plus one carve-out**: a user-submitted course that already has a round logged by a *different* user. Someone else demonstrably played it, so hiding it would break that user's round history. Verified: exactly 1.
- Replaces the SELECT policy with `using (approved_at is not null or created_by = auth.uid())`.

Post-state: **20,734 approved, 23 pending.** The submitter still sees their own pending course and can start a round on it, so "add my home course and play it right now" is preserved.

## Review found the first draft shipped a gate that didn't gate

Three independent reviews (DB, security, client-path audit). Four issues, all fixed here:

1. **Future crawler inserts would be invisible forever.** The backfill only approves rows existing at migration time; `scripts/crawl/db-writer.ts` never sets `approved_at`. Every newly crawled course would have been hidden from everyone — the same "all courses vanish" failure, deferred to the next crawl. Fixed with a `BEFORE INSERT` trigger keyed on `auth.role() = 'service_role'`.
2. **`created_by` was client-writable** — and it is the gate's entire trust signal. A user could post `created_by: null` (forging crawler provenance) or another user's uuid (spoofing attribution). Fixed with the pattern already proven on `course_tees` in `0037`: `default auth.uid()` + `with check (auth.role() = 'service_role' or auth.uid() = created_by)`.
3. **`approved_at` was client-writable** — a user could self-approve on insert. Closed by the `and approved_at is null` conjunct in the same policy.
4. **The backfill had no guard** despite the comment calling it the dangerous step. Now wrapped in a `DO` block that `RAISE EXCEPTION`s if fewer than 20,000 rows are approved, so a mistyped predicate aborts instead of silently hiding the corpus.

The trigger is keyed on `auth.role()`, **never** on `created_by is null` — the latter would let a client forge crawler provenance and be auto-approved, reopening #2.

Also folded in: `if exists` / `if not exists` guards (the convention `0015` established) and `set local lock_timeout = '2s'` before the policy swap, so a DDL wait can't queue behind a long read and stall every course query app-wide.

## Verified, not assumed

- All four existing client insert paths already set `created_by: user.id` and never set `approved_at` — they satisfy the tightened policy unchanged.
- Round detail degrades gracefully (course name falls back to `'Round'`) when a course is hidden; it does not crash. Facility drill-down filters to its existing empty state.
- `search_courses` (`0018`) is `security invoker`, so it re-enforces the new policy per caller.
- No client code touches the service-role key; the only service-role path is the dev-only Vite plugin (`apply: 'serve'`, excluded from production builds).
- BEFORE INSERT triggers fire before RLS `WITH CHECK`, so the trigger and policy compose correctly. `courses` now has three BEFORE INSERT triggers (`0027`, `0045`, `0053`) — disjoint columns, order-independent.

## Known limitation (signed off)

`holes` / `course_tees` / `hole_tees` stay publicly readable and carry coordinates, so this gate hides a pending course's **name and metadata but not its physical location** to anyone holding the `course_id`. Accepted deliberately — a golf course's location isn't secret, and gating those tables would put an `EXISTS` subquery on the two largest tables in the schema (170,829 holes, 39,949 tees). Documented in the migration header.

## ⚠️ Do not merge/apply yet

- **Sequencing:** prod is at `0046`; `0047`–`0052` are on `dev` awaiting the pending release. Applying `0053` first creates a numbering gap — it must land **with or after** that release.
- **`oga-dev` is paused**, so this can't be applied there until it's restored.
- **Known follow-ups, not in this PR:**
  - Import dedup race: with a pending course hidden, `getCourseByExternalId` misses it and the follow-up insert hits a 23505 on `courses_external_id_unique`. Blocks the second user with an error toast. Needs a `SECURITY DEFINER` lookup returning only `id`/`name`.
  - The dev-only Course Editor's search runs as the authenticated user, so it can't find a pending course to approve. Needs a service-role `?pending=1` endpoint, or the ops dashboard (#860) becomes the approval surface.
  - Round-import matcher should be confirmed to match only approved courses.

## Verification after applying

```sql
select count(*) filter (where approved_at is not null) as approved,
       count(*) filter (where approved_at is null)     as pending,
       count(*)                                        as total
  from courses;
-- expected: 20734 approved, 23 pending, 20757 total
```

Relates to #860 (the ops dashboard builds the review queue on top of this).

### DIFF
--- a/supabase/migrations/0053_courses_approved_at.sql
+++ b/supabase/migrations/0053_courses_approved_at.sql
@@ -1,0 +1,184 @@
+-- Gate user-submitted courses behind approval.
+--
+-- Before this migration the SELECT policy on `courses` was `using (true)` —
+-- any authenticated user could INSERT a course and it was immediately visible
+-- to everyone, with no review step. 24 such rows existed at time of writing
+-- (all `created_by` set, `external_id` null), from 20 submitters, 22 of them
+-- in the preceding 30 days.
+--
+-- After this migration a submitted course is visible only to its submitter
+-- until `approved_at` is stamped. Approval is service-role work; rejection is
+-- deleting the row. There is no status enum — a nullable timestamp is the
+-- whole queue state.
+--
+-- No client code changes accompany this. RLS does the filtering, so every
+-- existing course query on web and mobile stops returning unapproved rows for
+-- other users automatically, while the submitter keeps seeing their own via
+-- the `created_by = auth.uid()` branch (preserving "add my home course and
+-- play it right now").
+--
+-- THREE THINGS THIS MIGRATION MUST GET RIGHT, each found in review:
+--
+--   1. Future crawler inserts. The backfill below only approves rows that
+--      exist right now. `scripts/crawl/db-writer.ts` never sets `approved_at`
+--      and inserts with `created_by` null, so without the trigger below every
+--      newly crawled course would be invisible to everyone — the same
+--      "all courses vanish" failure, deferred to the next crawl.
+--
+--   2. `created_by` is the gate's entire trust signal and was client-writable.
+--      The old INSERT policy only checked `auth.uid() is not null`, so a user
+--      could post `created_by: null` (forging crawler provenance) or another
+--      user's uuid (spoofing attribution). Fixed with the `0037` pattern
+--      already proven on `course_tees`: a default plus an equality check.
+--
+--   3. `approved_at` was likewise client-writable — a user could self-approve
+--      by posting a timestamp. The INSERT policy now requires it to be null
+--      for non-service-role callers.
+--
+-- KNOWN, SIGNED-OFF LIMITATION: `holes`, `course_tees` and `hole_tees` remain
+-- publicly readable and carry tee/pin coordinates. So this gate hides a
+-- pending course's NAME and METADATA, but anyone holding its `course_id` can
+-- still read where it physically is. Accepted deliberately — a golf course's
+-- location is not secret and gating those tables would put an EXISTS subquery
+-- on the two largest tables in the schema (170k holes, 40k tees). Revisit if
+-- pending content ever becomes genuinely sensitive.
+
+-- ---------------------------------------------------------------------------
+-- Column
+-- ---------------------------------------------------------------------------
+
+-- Nullable, no default: adding a nullable column without a default does not
+-- rewrite the table, so this is cheap against ~20.7k rows. `if not exists`
+-- keeps the file re-runnable, per the convention 0015 established.
+alter table public.courses add column if not exists approved_at timestamptz;
+
+comment on column public.courses.approved_at is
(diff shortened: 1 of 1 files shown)


cner-smith:
## Review — feat(db): gate user-submitted courses behind approval (0053)

Thorough migration, and the header already pre-empts most of what a review would raise (client-writable `created_by`/`approved_at`, future crawler inserts, the backfill guard, lock timeout). Two things left.

**Concerns**
- `supabase/migrations/0053_courses_approved_at.sql:87` — the backfill's `created_by is null` branch treats a null submitter as "crawler corpus," but `created_by` is `on delete set null` (`0001_initial_schema.sql:31`) and `public.delete_my_account()` (`0036`) deletes `auth.users` → cascades to `profiles` → nulls `created_by` on that submitter's courses without deleting the course row. So `created_by is null` isn't identity-tight with "crawler." Impact today looks nil — the PR body states all 24 current user-submitted rows have `created_by` set, so nothing is misclassified at backfill time. Going forward it's also benign rather than a leak: a pending course whose submitter later deletes their account gets `created_by = null` with `approved_at` still null, so the SELECT policy (`created_by = auth.uid()` → NULL for everyone) makes it invisible cruft, not auto-published (the trigger is INSERT-only and the backfill has already run). I'm flagging it only because the whole migration's mental model is "`created_by is null` == crawler," and this is the one documented path that softens that invariant — worth a line in the header so a future reader doesn't over-trust the signal. May be considered out of scope.

**Nits**
- `supabase/migrations/0053_courses_approved_at.sql:85-87` — the backfill `do $$` block isn't re-run-safe: a second application would overwrite existing `approved_at` values with a fresh `now()`, losing the original approval instant. The file explicitly aims for re-runnability elsewhere (`if not exists` / `drop … if exists`, citing the `0015` convention), so this is the one non-idempotent statement. Impact is nil in practice since tracked migrations run once — noting only because of the stated re-runnable goal. A `where … and approved_at is null` conjunct would make it idempotent.

---
Looked good: RLS-only gate with no client changes, INSERT policy pins both `created_by = auth.uid()` and `approved_at is null` against self-approval/provenance forgery, service-role stamping keyed on `auth.role()` (not `created_by`) so provenance can't be forged, backfill asserts its row count before committing, and scope held to one concern.

---

PR #872, opened 2026-09-08
fix: close the OB review findings blocking dev→main (#839)

cner-smith:
Closes the review findings left open when #871, #868 and #866 merged. These were flagged as blocking the dev→main merge.

## Web and mobile took opposite stances on the same write

The #871 review caught that web kept the `existing.ob ||` fallback that mobile deliberately rejects. The fallback existed because web's review sheet **never hydrated from stored shot data at all** — it rebuilds rows purely from `placedPoints`, so `obCount(merged)` was structurally 0 and `row.shotResult` was never `'ob'` at save time.

Two bugs fell out of that:

- A hole marked OB on mobile opened on web with **no OB chip** and seeded the score **one stroke low**, then persisted that lower value over a correct one.
- The fallback made a stored `ob` **unclearable**. `SHOT_RESULTS` is single-select, so re-tagging that shot `pull` persisted `ob=true` alongside `shot_result='pull'` — SG charges −2 on a row whose label and map badge show a normal shot, with no way back.

Fixed at the root: the sheet now seeds `shotResult: 'ob'` from the hole's stored shots at hydration, so the row is authoritative and `ob` derives from it alone — identical to mobile.

The seed is **keyed by shot number, not position**, and gated on an equal shot count. `activeHoleShots` is not explicitly sorted, and `placedPoints` is a fresh placement with no correspondence to stored shots, so a positional seed would be the same index-alignment bug this feature has produced twice. On a count mismatch the flag is dropped rather than moved to a different shot — identical to the save path's existing snapshot guard, and identical to today's behaviour.

## `ShotLike` now requires an OB representation

Both fields were optional, so a fifth caller supplying neither would silently revert the GIR thresholds to the pre-#839 inflated behaviour **with a clean typecheck** — the exact bug class #839 fixed. Now a union, so the compiler requires one.

This immediately caught two test fixtures relying on the loophole; they now declare `ob: false`, which is also just true of them.

## `0053` backfill is re-runnable, and its central assumption is documented

**Safe to amend in place — `0053` has not been applied to any database.** That window closes the moment it reaches prod.

- The backfill was the one non-idempotent statement in a file that otherwise aims for re-runnability; a second application would overwrite every approval instant with a fresh `now()`. Guarded on `approved_at is null`, with the `OR` **parenthesised** so the guard cannot bind to the first branch only. First-run row count is unchanged, so the `< 20000` assertion still means what it meant.
- Recorded that `created_by is null` is **not** identity-tight with "crawler corpus": `delete_my_account` nulls it via cascade without deleting the course. Benign today — such a row keeps `approved_at = null` and stays invisible rather than auto-publishing — but the signal must not be trusted by a future policy or job.

## Also

- Mobile's score ticker now follows an OB→non-OB chip flip (web already did). Transition-driven, so re-selecting the same result cannot double-count.
- Dropped authoring-process narration from `OAuthButtons`' `busy` comment — "the Apple handler Task 5 adds" doesn't parse for anyone reading merged code, and the handler is already in that file.
- Corrected two comments whose stated mechanism no longer held: web's "the review sheet has no OB control" (it does) and "obCount(merged) is always 0 here" (no longer). A comment asserting a false mechanism is worse than none — the next reader treats the path as adjudicated.

## Verification

`pnpm test` — core 658, web 27, supabase 7, all green. `pnpm typecheck`, `pnpm --filter @oga/web build`, `apps/mobile npm run typecheck` all clean.

**Browser/device behaviour is unverified.** One known limitation worth QA: the seed reads whatever `activeHoleShots` holds when the sheet hydrates, and hydration is gated once per hole. If the shots query hasn't resolved by then the seed no-ops for that session — degrading to exactly today's behaviour, not worse, but the OB wouldn't appear until reopen. I did not add a loading-gate because "no stored shots" is a legitimate state for a first review and I can't distinguish the two without one.


cner-smith:
## Review — fix: close the OB review findings blocking dev→main (#839)

**Concerns**
- `supabase/migrations/0053_courses_approved_at.sql:106` vs `:122` — the new `approved_at is null` guard is documented (lines 97–101) as making this "the one re-runnable statement it was not," but the assertion below appears to defeat that. `get diagnostics approved_count = row_count` reads the row count of *this* UPDATE; on a genuine second application every row already carries `approved_at`, so 0 rows match the guard, `approved_count` is 0, and `approved_count < 20000` raises `'Predicate is wrong — aborting migration.'` So a re-run doesn't no-op cleanly — it aborts with a misleading message. In normal operation this never fires (migrations run once per version), so the guard is purely defensive; but as written the guard and the assertion claim opposite things. If re-runnability is the goal, the assertion likely needs to check total approved rows rather than this statement's `row_count`; if a loud abort on re-run is actually intended, the comment overstates it. I may be missing a reason the second run can't happen — flagging the tension, not asserting a bug.
- `supabase/migrations/0053_courses_approved_at.sql` — amending a committed, numbered migration in place is a deviation from CONTRIBUTING's "The next migration goes in the next free number." The safety of this rests entirely on 0053 never having been applied anywhere; the PR body asserts that, but it's unclear from the diff whether 0053 has run in a shared dev/staging DB or in contributors' local `db reset`s. If it has been applied, the corrected predicate won't take effect there (the version is already recorded) and a follow-up migration would be needed to fix those rows. Worth confirming before this merges toward main.
- `apps/mobile/components/auth/OAuthButtons.tsx:54` — the `busy`-comment cleanup is unrelated to the OB review findings this PR is scoped to (CONTRIBUTING: "One concern per PR"). It's a one-line comment edit, so low-stakes, but it's a separate area from the OB/migration work. May be intentional as part of a "close the open review findings" batch — noting it rather than pressing.

**Nits**
- `apps/web/src/components/round/HoleReviewSheet.tsx:207` and `apps/mobile/components/round/HoleReviewSheet.tsx` (setRow block) — several of the new comment blocks are long. Most explain genuinely non-obvious *why* (keyed-by-shot-number vs. position, the count-guard, the parenthesised `OR`) and earn their place under CONTRIBUTING's "why is non-obvious" bar; a couple lean toward restating the mechanism the code already shows. Minor.

---
Looked good: `ShotLike` as a union correctly forces the compiler to require an OB representation (closes the silent-revert loophole #839 fixed), and the two fixtures were updated in the same package; the web seed keys on `shotNumber` and `activeHoleShots` maps `shot_number → shotNumber` in `useRoundData.ts:324`, so the key matches and the count-guard drop-on-mismatch is the conservative choice; `ob` now deriving from the row alone round-trips with the save path; targets `dev`.

---

PR #846, opened 2026-08-20
feat(web): live store badges on landing + README download links

cner-smith:
OGA is live on both stores — wire up the download links.

- Landing page: App Store + Google Play badges now link to the live listings (open in a new tab).
- README: App Store + Play badges at the top + updated the "Live app" section (was "in pre-launch").

Store URLs:
- App Store: https://apps.apple.com/us/app/oga-open-golf-app/id6785314918
- Google Play: https://play.google.com/store/apps/details?id=app.opengolf.oga

Web-only; no mobile/OTA impact.

PR #873, opened 2026-09-08
fix: OB marker occlusion, shot-list OB, putt gating, 0053 guard (#839)

cner-smith:
Follow-on to #872. Fixes the two OB defects a player would actually hit, plus three smaller ones. Two of these were found by driving the app in a browser against a restored `oga-dev`, not by reading code.

## The OB shot's marker was invisible

An OB shot and its stroke-and-distance re-hit are **by rule** at the same coordinate. Their discs landed on the same pixel and whichever Mapbox added last won — in practice the re-hit. Measured in the browser, discs 2 and 3 were both at `x:883, y:631`. The ring told you a penalty happened somewhere; nothing told you which shot took it.

The OB disc now offsets 20px. The ring stays on the true origin, so the position is still told honestly — only the label moves. Re-measured after the fix: disc 2 at `x:863`, disc 3 at `x:883`, ring still at `x:878`. Both numbers legible.

Mobile needed its own layer pair rather than an expression, because `circle-translate` / `text-translate` are pixel offsets and are **not** data-driven.

## Neither shot list mentioned OB

The web list read `2 Approach · 3 Approach` for a hole with a penalty. Both platforms now show an OB tag and colour the shot number `caddie-neg`. This is the surface that answers "which shot" unambiguously even when discs coincide.

## A putt cannot be out of bounds

`shot_result` was putt-gated; `ob` was not. Mark a shot OB, then change its lie to green, and it persisted `ob=true, shot_result=null`. `sg-calculator`'s OB branch sits **ahead of** `holedOut`, so a made putt on that row booked **−2 putting** instead of roughly +0.1. Gated on both platforms.

## `0053`'s backfill guard could never be rehearsed

It asserted an absolute `approved_count >= 20000`, tuned to prod's corpus. **Dev has 15,321 courses, so the migration aborts there** — meaning prod would have been the first and only place it ever ran. It also counted *rows touched*, which broke once #872 made the backfill idempotent: a correct second run approves 0 and would have read as catastrophic failure.

Now asserts the resulting **state** — at least 90% of courses approved — which still catches what it was written for: a predicate that loses a branch in a rebase leaves the corpus overwhelmingly unapproved and invisible to every user.

## `landing.spec.ts` has been failing silently

Asserted `Free and open source` with `exact: true`; the hero badge has read `Free and open source · MIT` since #846. e2e isn't in CI so nothing caught it. Unrelated to OB, one line.

## Verification

`pnpm test` — core 658, web 27, supabase 7. `pnpm typecheck`, `pnpm --filter @oga/web build`, `apps/mobile npm run typecheck` all clean. Full Playwright suite green (23/23, including the landing spec that was failing before).

**Migrations `0046`–`0052` and `0054` were applied to `oga-dev` and `0054` was behaviourally verified there** against a real Postgres for the first time — OB delete drops the score by 2, non-OB by 1, survivors renumber contiguously. `0053` was not applied to dev; it couldn't be until the guard change in this PR.

**Unverified:** every mobile change here. No device, no emulator — `MAPBOX_DOWNLOADS_TOKEN` is an EAS-only secret, so a local native build can't fetch the Mapbox SDK. The mobile map change in particular splits one layer pair into two and only typecheck backs it.


cner-smith:
## Review — fix: OB marker occlusion, shot-list OB, putt gating, 0053 guard (#839)

**Concerns**
- Scope — this PR bundles the OB-visibility cluster (map discs + shot lists + putt gating, all #839) with two changes that aren't OB at all: the `0053` backfill-guard rewrite (`supabase/migrations/0053_courses_approved_at.sql`) and the `apps/web/e2e/landing.spec.ts` assertion, which the description itself calls "Unrelated to OB, one line." CONTRIBUTING.md is explicit here — "One concern per PR. UI restyle, bug fix, and new feature do not belong in the same PR," and "PRs that add scope beyond their stated concern will be asked to split." Both riders are self-contained and would revert independently of the OB work, so I'd lift at least those two into their own `fix:`/`chore:` PRs. Advisory — flagging, not blocking.
- `apps/mobile/components/round/markers/BreadcrumbLayers.tsx` (the new `prevShotsObDisc` / `prevShotsObNumber` layers) applies `circleTranslate: [-20, 0]` / `textTranslate: [-20, 0]` unconditionally to every `ob == true` row. Web only nudges when something is actually stacked on the coordinate — `s.ob === true && shareCount > 1` in `apps/web/src/components/round/map/useMapLayers.ts`, with the comment "a lone OB shot stays dead centre in its ring." So a lone OB shot with no coincident re-hit (e.g. the in-progress state after tagging OB but before the re-hit is logged) reads centered on web but detached from its ring on mobile — the platforms disagree in exactly the case the web guard was added for. `circle-translate` isn't data-driven, so matching web would mean pushing the coincidence decision into the feature properties (an `obShifted` flag the filter keys on) rather than a static layer offset. The description notes all mobile changes here are unverified, which is where I'd expect this to surface. May be an accepted trade-off given no device access — noting it.

**Nits**
- The "a putt cannot be OB" guard `ob: isPuttRow ? false : row.shotResult === 'ob'` is now duplicated verbatim in `apps/web/src/pages/rounds/hooks/useRoundActions.ts` and `apps/mobile/components/round/hole/useShotActions.ts`. It's a genuine shared domain rule — the kind of thing CONTRIBUTING.md's "pure logic lives in `@oga/core`" points at — but it's a one-line guard inline in each platform's row builder with only two callers, where the 3-caller rule cuts the other way. Keeping it inline may well be right; flagging the drift risk, not asking for a change.

- The `0053` guard rewrite relies on the migration never having committed anywhere (its `raise exception` rolls the whole migration back, so an aborted run leaves no trace — per the description). That reasoning holds only if no environment carries a partially-applied `0053`; worth confirming before relying on the in-place edit, since the new guard won't re-run where the old one already committed.

---
Looked good: web's coincidence key (`${lng},${lat}`) matches the count key, so the offset actually fires rather than silently no-op'ing; the ring stays on the true origin so position is still told honestly; the putt-OB gate correctly lands ahead of sg-calculator's OB branch; base is `dev`, no dep/nativewind changes, no `any`, and each individual change is minimal.

---

cner-smith:
Addressed the blocking concern from the review above in `552c67a`.

**Mobile/web OB offset divergence — fixed.** The coincidence decision now lives in the feature (`obShifted`) and the OB disc/number layers split into shifted/unshifted pairs filtered on it, since `circle-translate`/`text-translate` aren't data-driven. A lone OB shot now stays centred in its ring on mobile, matching web. Keyed on `lng,lat` over the same `previousShots` the layers render — same key order and scope as web's guard.

**On the `0053` suggestion** — confirmable now, and it holds: prod is still at `0046`, so `0053` has never run there; `oga-dev` applied it cleanly during last session's rehearsal. No environment carries a partially-applied `0053`, so the in-place guard edit is safe.

**Scope and the duplicated putt guard** — left as-is, both were flagged advisory. The scope riders would revert independently if needed, and the one-line `ob: isPuttRow ? false : …` guard has two callers, where the 3-caller rule cuts against extraction.

Mobile stays typecheck-only — no device. The counting logic was checked against lone-OB, covered-OB, coincident-non-OB, no-OB and triple-stack cases first, since this is the half that can't be verified on a build.

Issue #859, opened 2026-08-26
Auth: Google Sign-In (sign in + create account)


cner-smith:
## Summary

Let users **sign in and create an account with Google**, alongside the existing email/password flow — on both web and mobile.

## Why

- Faster onboarding (no password to set/confirm) and higher signup conversion.
- Sidesteps the email-confirmation deep-link friction that email/password has hit.

## Scope

- Add Google as a Supabase OAuth provider (web + mobile).
- **New account on first Google sign-in:** the existing `handle_new_user()` trigger already creates the `profiles` row, so a first-time Google user should land in onboarding just like an email signup.
- Route new vs returning Google users correctly (new → onboarding; returning w/ `onboarding_completed` → app).

## Considerations / gotchas

- **Mobile redirect/deep-link:** reuse the auth-callback pattern from the email-confirm work (PKCE + `oga://` deep link); Google OAuth returns via the browser → app handoff.
- **Turnstile:** OAuth doesn't need the captcha widget; make sure the flow doesn't get gated by it.
- **Web:** `signInWithOAuth({ provider: 'google' })` + redirect handling in `AuthContext`.
- Google Cloud OAuth client + Supabase provider config (client id/secret, authorized redirect URIs for web + the Supabase callback).

## Acceptance

- New user can create an account via Google and reach onboarding, then the app.
- Returning Google user signs straight in.
- Works on web and both mobile platforms.

PR #874, opened 2026-09-08
test(web): Google sign-in wiring coverage (#859)

cner-smith:
Adds four browser checks for the Google OAuth wiring, none of which need a Google account.

- `Continue with Google` renders on `/login` and `/signup`
- clicking it reaches `accounts.google.com` carrying our `client_id`, Supabase's `/auth/v1/callback` as `redirect_uri`, and `scope=email+profile`
- the `redirect_to` survives into Google's URL — **this is the allow-list assertion**: Supabase silently substitutes Site URL when a `redirect_to` is not allow-listed, so its survival proves the entry is configured
- `/auth/callback` fails fast on a missing or bogus code instead of sitting on "Signing you in…" — that hang was the #509 field failure

Verified against the dev project with Google + Apple now enabled on both projects. Network-dependent by design: it asserts the live OAuth configuration, not only our code. e2e is local-only today, so no CI flake cost.


cner-smith:
## Review — test(web): Google sign-in wiring coverage (#859)

**Concerns**
- `apps/web/e2e/google-auth.spec.ts:29,30,34` — The assertions hardcode one specific project's OAuth wiring: the maintainer dev project's `client_id` (`594879721243-`), its Supabase URL (`txquvfeyvkaetqlamqlz.supabase.co`), and `localhost:5173/auth/callback`. OGA is documented as self-hostable (README self-hosting, "link a Supabase project"), so anyone running the suite against their own Supabase project will fail all three assertions — the test asserts *the maintainer's* live config, not the code. Since e2e is local-only today there's no CI cost, but as a committed regression guard it's runnable in exactly one environment. To be a shared test, these would need to read the same env/config the app reads (client id, Supabase URL, site URL) rather than literals. To be clear, these are public identifiers (OAuth client_id and the Supabase project URL both ship to the browser), so this is not a secret leak.
- `apps/web/e2e/google-auth.spec.ts:3` — `// TEMPORARY verification of the Google OAuth wiring` lands the file in the permanent suite with no removal trigger. If it's a real regression guard, drop the "TEMPORARY" and address the environment coupling above; if it was a one-time confirmation that #859/#866 wired up, it may not belong in the committed tree at all. "TEMPORARY" tests with no owner tend to become permanent.

**Nits**
- `apps/web/e2e/google-auth.spec.ts:17` — leftover `console.log('LANDED ON:', ...)` debug line.
- `apps/web/e2e/google-auth.spec.ts:19-20` and `:32-33` — the "Supabase silently swaps redirect_to for Site URL when it is NOT allow-listed" comment is duplicated nearly verbatim; one placement (next to the `localhost` assertion it explains) is enough.

- The three network-dependent tests hit live `accounts.google.com` and the live Supabase project on every run. That's acceptable while e2e is local-only, but if these ever move into CI they'll be flaky by construction — worth deciding whether they're kept local-only permanently or reworked to be hermetic before that happens.

---
Looked good: targets `dev`; scope held to one concern (tests only, no app logic or exports touched); no secret material and no `@oga/core` changes; decode-until-stable loop (`:23-28`) is a reasonable way to handle the nested URL encoding.

---

cner-smith:
Addressed the review concerns above in `70545ec`.

**Environment coupling — fixed.** The spec no longer asserts one project's literal wiring. `client_id` is checked for presence and shape (`*.apps.googleusercontent.com`) rather than a specific value; `redirect_uri` is compared against `process.env.VITE_SUPABASE_URL` (which `playwright.config.ts` already loads from `.env.test.local`); the allow-listed return URL comes from Playwright's `baseURL`. That still catches a Supabase project whose Google provider calls back to a *different* project — it just no longer hardcodes which project is correct, so the suite runs against a self-hosted deployment.

**"TEMPORARY" — dropped.** These are real regression guards for the #859/#866 wiring and the #509 hang, so the header now says what they guard instead of implying they should be deleted by someone, sometime.

**Nits — both fixed.** Leftover `console.log` removed; the duplicated `redirect_to` comment collapsed to the single placement next to the assertion it explains.

**On the network-dependence suggestion** — noted in the file header rather than changed: the live-Google test is fine while e2e is local-only, and would need to be made hermetic before it moves into CI.

Verified: 5/5 pass against `oga-dev`, and the `client_id` assertion was deliberately falsified against a wrong pattern to confirm it can actually fail rather than passing vacuously.

PR #815, opened 2026-07-13
feature(web): app-like mobile shell — bottom tab bar + top bar (phase 1)

cner-smith:
**Phase 1** of the mobile-web redesign (app-like responsive web). On **phone viewports** (`md-`), the desktop sidebar is replaced by a native-style shell; **desktop is unchanged** (every new element is `md:hidden`-gated).

### Changes
- **New `MobileTabBar`** — fixed bottom tab bar mirroring the native app's five tabs: **Home** (`/dashboard`), **Stats** (`/stats`), **Patterns** (`/patterns`), **Practice** (`/practice`), **Profile** (`/settings`), with hand-authored inline-SVG icons (no icon-lib dependency) + labels + active state. Safe-area-inset aware.
- **`AppShell`** — adds a slim fixed **mobile top bar** (`oga.` brand + a menu button that opens the *existing* drawer), pads `main` for the two bars, removes the old in-content hamburger.
- Secondary routes (Rounds, Learn, My Bag, Settings) stay reachable via the top-bar menu → drawer for now; **Phase 2** folds them into the Home + Profile pages, **Phase 3** adds GPS to the mobile-web live round.

### Verification
- `pnpm --filter @oga/web typecheck` ✅
- Render-verified at a 390×844 mobile viewport (signed-in): top bar + 5-tab bottom bar, active state tracks route (Home→Stats). Screenshots in the PR thread.

Desktop sidebar untouched.

PR #875, opened 2026-09-08
fix(web): mobile tab bar rendering on desktop (#815 regression)

cner-smith:
Reported from a Vercel preview: the mobile bottom tab bar renders on desktop, across the bottom of the page next to the sidebar.

## Cause

```tsx
className="md:hidden fixed bottom-0 …"
style={{ display: 'flex', … }}
```

`md:hidden` emits `display:none` inside a media query — a stylesheet rule. The inline `display:'flex'` one line below beats any stylesheet rule short of `!important`, so the breakpoint never applied. The gate was written correctly and then silently overridden.

Moving `flex` into the class list puts both display values in the same cascade layer.

## This is live in production

Introduced by #815 and present on `main` — every desktop web user currently sees it.

## About the test

It asserts computed `display`, **not** Playwright's visibility heuristic. My first attempt used `toBeHidden()` and **passed against the broken code** — the heuristic reported the bar as hidden at 1440px while it was measurably painting 1440×58 on screen. I only caught that because I re-ran the test against the pre-fix version to check it could fail.

Verified both directions: fails `Expected "none", Received "flex"` at 1440 and 768 on the old code, passes on the new, and correctly asserts `flex` at 390.


cner-smith:
> No blockers, concerns, or nits — looks good.

---

